PR #153, opened 2019-06-27
Bug 3451381

asachin96:
Dev certs now supports window7!


packages/office-addin-dev-certs/scripts/install.ps1
```diff
@@ -0,0 +1,19 @@
+if($args.Count -ne 2){
+    throw "Usage install.ps1 <LocalMachine/CurrentUser> <CA certificate path>"
```
akrantz: It would be more readable to have `Usage: ` so colon and space separates it from the rest.

akrantz: Using pipe would be clearer than slash for a choice: `<LocalMachine | CurrentUser>`

akrantz: Using dashes instead of spaces would be clearer that it is one thing: `<CA-certificate-path>`

packages/office-addin-dev-certs/scripts/install.ps1
```diff
@@ -0,0 +1,19 @@
...
+else{
+    # Legacy system support
+    $pfx = new-object System.Security.Cryptography.X509Certificates.X509Certificate2
+    $pfx.import($caCertificatePath)
+
+    $store = new-object System.Security.Cryptography.X509Certificates.X509Store("Root", $machine)
+    $store.open("MaxAllowed")
+    $store.add($pfx)
+    $store.close()
+}
```
akrantz: It's usually best to have a newline on the last line. GitHub diff is nice enough to show whether this is the case.

packages/office-addin-dev-certs/scripts/uninstall.ps1
```diff
@@ -0,0 +1,19 @@
+if($args.Count -ne 2){
+    throw "Usage uninstall.ps1 <LocalMachine/CurrentUser> <CA certficate Name>"
```
akrantz: same comments as before apply here and other script...

packages/office-addin-dev-certs/src/install.ts
```diff
@@ -8,19 +8,16 @@ import { deleteCertificateFiles, uninstallCaCertificate } from "./uninstall";
 import { isCaCertificateInstalled, verifyCertificates } from "./verify";
 
 function getInstallCommand(caCertificatePath: string, machine: boolean = false): string {
-   let command: string;
-
    switch (process.platform) {
       case "win32":
-         command = `powershell Import-Certificate -CertStoreLocation cert:\\${machine ? "LocalMachine" : "CurrentUser"}\\Root '${caCertificatePath}'`;
+         return `powershell -ExecutionPolicy Bypass scripts\\install.ps1 ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'`;
```
akrantz: Does this assume that the ExecutionPolicy will allow running the script. What happens if it is not allowed? Perhaps we need to explain how to enable it if that is the case. Yet another thing a user could run into. If this is a problem, perhaps then for the single command case you run it like earlier and otherwise you only use the script if needed.

asachin96: -ExecutionPolicy Bypass allows running scripts when the policy is disabled.

asachin96: For eg in my VM policy didn't allow me to load scripts, by adding this flag I was able to load/run scripts

akrantz:
What happed if policy doesn't allow scripts to be run?

akrantz:
I'm worried that this won't be enabled by default.

akrantz:
I'm worried that this won't be enabled by default.

asachin96:
> What happed if policy doesn't allow scripts to be run?

with  -ExecutionPolicy Bypass it doesn't matter what policy is

PR #154, opened 2019-06-27
create parameters object

donaldgu:
Update the getParameters function to take an object instead of multiple parameters.  

https://office.visualstudio.com/DefaultCollection/OC/_workitems/edit/3363084/

### DIFF
--- a/packages/custom-functions-metadata/src/generate.ts
+++ b/packages/custom-functions-metadata/src/generate.ts
@@ -75,6 +75,15 @@ interface IArrayType {
     type: ts.SyntaxKind;
 }
 
+interface IParameterType {
+    enumList: string[];
+    extra: IFunctionExtras;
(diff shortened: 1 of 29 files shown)


akrantz:
This seems OK. Perhaps the names could be updated, because IParameterType is not for a single parameter. I'm not clear though what this change gets us.

donaldgu:
It’s more for readability of the code.  The getParameters function parameters were getting quite large and this creates an object of the said parameters.

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -75,7 +75,7 @@ interface IArrayType {
     type: ts.SyntaxKind;
 }
 
-interface IParameterType {
+interface IGetParametersObjectType {
```
akrantz: Suggest: IGetParametersArguments

Issue #129, opened 2019-05-16
office-addin-debugging: detached dev-server process not visible on Mac


akrantz:
On Mac, if the dev certs are not installed and the dev-server is not running, the `office-addin-debugging start` command will start the dev-server as a detached process. Since there is no window, the user cannot enter the password for the dev certs. After 30 seconds it times out.

PR #156, opened 2019-06-28
Linux support

GabeMedrash:
This provides a 'linux' case for installing, uninstalling, and verifying a cert in `office-addin-dev-certs`. Before adding changes, tests fail on my machine (Ubuntu 18.04). After adding, tests pass. That said, it's a fairly naive implementation.

### DIFF
--- a/packages/office-addin-dev-certs/src/install.ts
+++ b/packages/office-addin-dev-certs/src/install.ts
@@ -17,7 +17,7 @@ function getInstallCommand(caCertificatePath: string, machine: boolean = false):
       case "darwin": // macOS
          return `sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain '${caCertificatePath}'`;
       case "linux":
-         return `sudo mv ${caCertificatePath} /usr/local/share/ca-certificates && sudo /usr/sbin/update-ca-certificates`;
+         return `sudo cp ${caCertificatePath} /usr/local/share/ca-certificates && sudo /usr/sbin/update-ca-certificates`;
       default:
          throw new Error(`Platform not supported: ${process.platform}`);
    }
--- a/packages/office-addin-dev-certs/src/install.ts
+++ b/packages/office-addin-dev-certs/src/install.ts
@@ -16,6 +16,8 @@ function getInstallCommand(caCertificatePath: string, machine: boolean = false):
          return `powershell -ExecutionPolicy Bypass -File "${script}" ${machine ? "LocalMachine" : "CurrentUser"} "${caCertificatePath}"`;
       case "darwin": // macOS
          return `sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain '${caCertificatePath}'`;
+      case "linux":
+         return `sudo mv ${caCertificatePath} /usr/local/share/ca-certificates && sudo /usr/sbin/update-ca-certificates`;
       default:
          throw new Error(`Platform not supported: ${process.platform}`);
    }
--- a/packages/office-addin-dev-certs/src/uninstall.ts
+++ b/packages/office-addin-dev-certs/src/uninstall.ts
@@ -15,6 +15,8 @@ function getUninstallCommand(machine: boolean = false): string {
          return `powershell -ExecutionPolicy Bypass -File "${script}" ${machine ? "LocalMachine" : "CurrentUser"} "${defaults.certificateName}"`;
       case "darwin": // macOS
          return `sudo security delete-certificate -c '${defaults.certificateName}'`;
+      case "linux":
+         return `sudo rm /usr/local/share/ca-certificates/${defaults.certificateName} && sudo /usr/sbin/update-ca-certificates`
       default:
          throw new Error(`Platform not supported: ${process.platform}`);
    }
--- a/packages/office-addin-dev-certs/src/uninstall.ts
+++ b/packages/office-addin-dev-certs/src/uninstall.ts
@@ -16,7 +16,7 @@ function getUninstallCommand(machine: boolean = false): string {
       case "darwin": // macOS
          return `sudo security delete-certificate -c '${defaults.certificateName}'`;
       case "linux":
-         return `sudo rm /usr/local/share/ca-certificates/${defaults.certificateName} && sudo /usr/sbin/update-ca-certificates`
+         return `sudo rm /usr/local/share/ca-certificates/${defaults.caCertificateFileName} && sudo /usr/sbin/update-ca-certificates --fresh`
       default:
          throw new Error(`Platform not supported: ${process.platform}`);
    }
--- a/packages/office-addin-dev-certs/src/verify.ts
+++ b/packages/office-addin-dev-certs/src/verify.ts
@@ -16,7 +16,7 @@ function getVerifyCommand(): string {
        case "darwin": // macOS
           return `security find-certificate -c '${defaults.certificateName}' -p | openssl x509 -checkend 86400 -noout`;
         case "linux":
-          return `[ -f /usr/local/share/ca-certificates/${defaults.caCertificateFileName} ]`;
+          return `[ -f /usr/local/share/ca-certificates/${defaults.caCertificateFileName} ] && openssl x509 -in /usr/local/share/ca-certificates/${defaults.caCertificateFileName} -checkend 86400 -noout`;
        default:
           throw new Error(`Platform not supported: ${process.platform}`);
     }
--- a/packages/office-addin-dev-certs/src/verify.ts
+++ b/packages/office-addin-dev-certs/src/verify.ts
@@ -16,7 +16,7 @@ function getVerifyCommand(): string {
        case "darwin": // macOS
           return `security find-certificate -c '${defaults.certificateName}' -p | openssl x509 -checkend 86400 -noout`;
         case "linux":
-          return `[ -f /usr/local/share/ca-certificates/${defaults.certificateName} ]`;
+          return `[ -f /usr/local/share/ca-certificates/${defaults.caCertificateFileName} ]`;
        default:
           throw new Error(`Platform not supported: ${process.platform}`);
     }
--- a/packages/office-addin-dev-certs/src/verify.ts
+++ b/packages/office-addin-dev-certs/src/verify.ts
@@ -15,6 +15,8 @@ function getVerifyCommand(): string {
           return `powershell -ExecutionPolicy Bypass -File "${script}" "${defaults.certificateName}"`;
        case "darwin": // macOS
           return `security find-certificate -c '${defaults.certificateName}' -p | openssl x509 -checkend 86400 -noout`;
+        case "linux":
+          return `[ -f /usr/local/share/ca-certificates/${defaults.certificateName} ]`;
        default:
           throw new Error(`Platform not supported: ${process.platform}`);
     }


GabeMedrash:
With these changes, running `office-addin-debugging start ...` within a repo created by the Yeoman generator doesn't manage to actually start the dev server, even though running `npx webpack-dev-server` directly works just fine. `office-addin-debuggin start` gives me:
```
> office-addin-taskpane-react@0.0.1 start /home/.../Documents/fms-uploader
> office-addin-debugging start manifest.xml

Debugging is being started...
App type: desktop
Starting the dev server... (webpack-dev-server --mode development)
The developer certificates have been generated in /home/.../.office-addin-dev-certs
Installing CA certificate "Developer CA for Microsoft Office Add-ins"...
[sudo] password for ...: 
You now have trusted access to https://localhost.
Certificate: /home/.../.office-addin-dev-certs/localhost.crt
Key: /home/.../.office-addin-dev-certs/localhost.key
Unable to start the dev server. Error: The dev server is not running on port 3000.
Sideloading the Office Add-in...
Debugging started.
```

The cert generation works without error, though I do find that I have to separately go to `https://localhost:3000` in a browser and manually accept the self-signed cert before the add-in will work. Is that not the case on Windows or Mac?

akrantz:
It sounds like this issue which should be fixed: https://github.com/OfficeDev/Office-Addin-Scripts/issues/129. It should not call `office-addin-dev-certs install` to prompt for the password, and then start the dev server afterward.

akrantz:
The log you showed has "Starting the dev server..." before "Installing CA certificate". With that fix, the certs should be installed first, prompt for the password, and then afterward it should start the dev server.

GabeMedrash:
Oh cool, glad to hear there's already related work in progress.

akrantz:
No, I'm saying this was already fixed and so I am unclear why you're seeing the behavior.

akrantz:
If the dev server starts and needs to prompt, you won't see the prompt, and after it times out, it shows the error. So the dev cert install needs to happen before running the dev server so that you can enter the password.

You can do "office-addin-dev-certs install" separately first. Once installed, it won't be an issue.

GabeMedrash:
Shouldn't you need to [`await` the result of `ensureCertificatesAreInstalled`](https://github.com/OfficeDev/Office-Addin-Scripts/blame/master/packages/office-addin-debugging/src/start.ts#L84)?

I did that locally and found the logs coming in the correct order, but still getting an `Unable to start the dev server` errror.

akrantz:
Yes, there is an await missing there.

akrantz:
You can do "npm run dev-server" or use Terminal, Run Task in VS Code to run the dev server before you start debugging. Once it's running, office-addin-debugging will check and should find it is already running.

The code that checks whether the port is in use may not be working right on linux.

packages/office-addin-dev-certs/src/uninstall.ts
```diff
@@ -13,6 +13,8 @@ function getUninstallCommand(machine: boolean = false): string {
          return `powershell -ExecutionPolicy Bypass scripts\\uninstall.ps1 ${machine ? "LocalMachine" : "CurrentUser"} '${defaults.certificateName}'`;
       case "darwin": // macOS
          return `sudo security delete-certificate -c '${defaults.certificateName}'`;
+      case "linux":
+         return `sudo rm /usr/local/share/ca-certificates/${defaults.certificateName} && sudo /usr/sbin/update-ca-certificates`
```
asachin96: When you move ca.crt file to /usr/local/share/ca-certificates shouldn't you be deleting ca.crt?. So this should be sudo rm /usr/local/share/ca-certificates/${defaults.caCertificateFileName}

packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -12,6 +12,8 @@ function getVerifyCommand(): string {
           return `powershell -ExecutionPolicy Bypass scripts\\verify.ps1 '${defaults.certificateName}'`;
        case "darwin": // macOS
           return `security find-certificate -c '${defaults.certificateName}' -p | openssl x509 -checkend 86400 -noout`;
+        case "linux":
+          return `[ -f /usr/local/share/ca-certificates/${defaults.certificateName} ]`;
```
asachin96: This should be [ -f /usr/local/share/ca-certificates/${defaults.caCertificateFileName} ]. Still this doesn't tell if certificate is expired or not

GabeMedrash: Thanks, will fix.

Yeah, it's fairly naive. Off the cuff, we could do something like `openssl verify /usr/local/share/ca-certificates/${defaults.caCertificateFileName}`. Will do two seconds of research to make sure that's legit.

packages/office-addin-dev-certs/src/uninstall.ts
GabeMedrash: I'm not sure I understand the question. 

Also I think the `update-ca-certificates` statement here should be called with `--fresh`.

asachin96: defaults.certificateName value is "Developer CA for Microsoft Office Add-ins" not ca.crt. So rm won't work

GabeMedrash: Ah, thanks. Of course. That's what I get for doing anything after 9pm. Will change.

packages/office-addin-dev-certs/src/verify.ts
GabeMedrash: I opted to use the same command used to verify the cert on MacOS, which just checks it won't expire within 86400s.

akrantz:
@GabeMedrash I made a fix for the missing await. Please merge the latest changes from master and let me know whether it fixed that issue or not. Thanks.

GabeMedrash:
Confirmed--I saw that and rebased my changes on top of master. 

Output on first run:
```
$ npm run start:web

> office-addin-taskpane-react@0.0.1 start:web /.../
> office-addin-debugging start manifest.xml web

Debugging is being started...
App type: web
The developer certificates have been generated in /.../.office-addin-dev-certs
Installing CA certificate "Developer CA for Microsoft Office Add-ins"...
You now have trusted access to https://localhost.
Certificate: /.../.office-addin-dev-certs/localhost.crt
Key: /.../.office-addin-dev-certs/localhost.key
Starting the dev server... (webpack-dev-server --mode development)
The dev server is running on port 3000. Process id: 8300
Debugging started.
```

Output on second run, after having already run `npm run stop`:
```
$ npm run start:web

> office-addin-taskpane-react@0.0.1 start:web /.../
> office-addin-debugging start manifest.xml web

Debugging is being started...
App type: web
The dev server is already running on port 3000.
Debugging started.
```

As before though--and maybe this is just a Linux thing--applications can choose whether or not they trust certs registered with the system. Neither Firefox nor Chrome do, so if doing development against the web-version of Excel, a developer must first navigate to `https://localhost:3000` and accept the self-signed cert before the add-in will render. Maybe that's just a thing to document?

akrantz:
The certs that are created are not self-signed. It creates an SSL certificate for https://localhost which is signed by a CA certificate. The install command is to install the CA cert so it is trusted for the system (after a prompt on Windows or password entry on other platforms). The point of this is to automate the trust for the CA certificate so there should be no other steps afterwards. On Windows or Mac today once the certs are generated and the CA cert is trusted (installed), a browser like chrome should not report any issues going to https://localhost.

GabeMedrash:
Hm. Here is the warning displayed by the browser when I navigate to `https://localhost:3000`:
```
Your connection is not private
Attackers might be trying to steal your information from localhost (for example, passwords, messages, or credit cards). Learn more
NET::ERR_CERT_AUTHORITY_INVALID
Subject: 127.0.0.1

Issuer: Developer CA for Microsoft Office Add-ins

Expires on: Aug 2, 2019

Current date: Jul 3, 2019

PEM encoded chain:
-----BEGIN CERTIFICATE-----
MIIDgz...
-----END CERTIFICATE-----
```

I'm not aware of another way to non-interactively trust a certificate on Linux.

akrantz:
OK, it sounds like on Linux the browsers don't automatically trust the certs installed on the system. If you get the warning, is it then easy to tell the browser to trust the cert?

GabeMedrash:
Yep, definitely easy to tell the browser to trust the cert. Similarly to dealing with a self-signed cert, you can click a couple buttons to tell the browser to trust it. 

But you need to navigate separately to `https://localhost:3000` to trust the cert before attempting to side-load your add-in into the web version of Office--to my knowledge, that isn't an option within the iframe. So, fundamentally, I believe this is now a documentation problem.

GabeMedrash:
@akrantz @asachin96 Anything further I can do to help move this across the finish line?

asachin96:
Could you place the certificate in office-addin-dev-certs folder in /usr/share/ca-certificate. So that the user knows what that certificate is for.

packages/office-addin-dev-certs/src/install.ts
```diff
@@ -13,6 +13,8 @@ function getInstallCommand(caCertificatePath: string, machine: boolean = false):
          return `powershell -ExecutionPolicy Bypass scripts\\install.ps1 ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'`;
       case "darwin": // macOS
          return `sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain '${caCertificatePath}'`;
+      case "linux":
+         return `sudo mv ${caCertificatePath} /usr/local/share/ca-certificates && sudo /usr/sbin/update-ca-certificates`;
```
asachin96: This should be copy instead of a move since other packages try to read from the original location

grantspeelman:
Would be great if this could be merged in

amiram:
@GabeMedrash will you be able to finish this?

GabeMedrash:
Clearly I've been derelict in my stewardship of this feature--apologies. I'll take a look at this in the next few days to see if I have the bandwidth to take this across this finish line and I'll report back. It looks like there are only two small changes required from the original review.

packages/office-addin-dev-certs/src/install.ts
GabeMedrash: Done in https://github.com/OfficeDev/Office-Addin-Scripts/pull/156/commits/a7747c876cbcf7a6c65013824cf10556aa5c9079

GabeMedrash:
> Could you place the certificate in office-addin-dev-certs folder in /usr/share/ca-certificate. So that the user knows what that certificate is for.

As in, instead of `/usr/local/share/ca-certificates`? Or in addition?

TCourtneyOwen:
@GabeMedrash thanks for submitting this PR.  Would it be possible for you to add tests for this change to src/test.ts?  Broadly speaking we don't support Linux at this time, but if you can add tests to ensure your changes in fact work, this will boost our confidence so we can move forward with this PR.

Thanks,

Courtney

unhammer:
> > Could you place the certificate in office-addin-dev-certs folder in /usr/share/ca-certificate. So that the user knows what that certificate is for.
> 
> As in, instead of `/usr/local/share/ca-certificates`? Or in addition?

I would prefer if it goes to `/usr/local/share/ca-certificates/office-addin-dev-certs/` (something like `path.join("/usr/local/share/ca-certificates/", defaults.certificateDirectoryName)`. 

It should *not* go to `/usr/share/…` (without the `local`), that prefix is reserved for stuff installed by your system package manager.

unhammer:
> Broadly speaking we don't support Linux at this time, but if you can add tests to ensure your changes in fact work, this will boost our confidence so we can move forward with this PR.

Just to add a datapoint, it works for me on Ubuntu 19.10 (though I changed the path since as I mentioned to make it clearer for my future self what the cert was for).

nrueckmann:
Just created a new PR as it seems like this PR is going nowhere.

igor-ribeiiro:
Closing this PR, as https://github.com/OfficeDev/Office-Addin-Scripts/pull/426 solved this issue

Issue #160, opened 2019-07-02
office-addin-debugging 'stop' doesn't work on MacOS


andrewconnell:
On macOS 10.14.5, starting the debugger with `npm run start` which runs `office-addin-debugging start manifest.xml` spins up a debugging server but in a detached process. Running `npm run stop` or `office-addin-debugging stop manifest.xml` does not stop the process.

The only way I was able to get it to stop was to find the node process that had port 3000 open and kill it with Activity Monitor.

akrantz:
You can avoid this by starting it with `npm run dev-server` in a separate Terminal window tab, or if using VS Code, use `Terminal`, `Run Task...`, `Dev Server` to run it inside VS Code. You can then close the Terminal window tab or close VS Code to stop the task.

We've already done some work to know the process id of the detached process which is started. We need `office-addin-debugging start` to write this to disk so that `office-addin-debugging stop` can know to stop the process.

andrewconnell:
Good to know... would have helped me this week. However as it stands now, the `stop` command doesn't work when it says it successfully completed it's task.

akrantz:
Yes, I agree. I'm not disputing that. We will work to make improvements here. Sorry if this caused problems for you.

akrantz:
This is now fixed. I have published office-addin-debugging using the "next" dist-tag, so you install the updated package using "npm install -D office-addin-debugging@next".

jpshankle:
It still appears to have the problem of getting
`$ office-addin-debugging stop manifest.xml
env: node\r: No such file or directory
error Command failed with exit code 127.`
The way to resolve the issue is to change the line ending in .bin -> office-addin-debugging from CRLF to LF.

akrantz:
@jpshankle I opened a separate issue about that: https://github.com/OfficeDev/Office-Addin-Scripts/issues/168.

akrantz:
@jpshankle Please see comments in https://github.com/OfficeDev/Office-Addin-Scripts/issues/168. I wasn't able to mention your username there. Not sure why, but please reply there and we will follow up. Thanks.

suuunly:
> This is now fixed. I have published office-addin-debugging using the "next" dist-tag, so you install the updated package using "npm install -D office-addin-debugging@next".

I tried doing so, updating from `5.0.12`, to `5.1.0`. However, running 
`office-addin-debugging stop manifest.xml`
[...]

PR #161, opened 2019-07-03
Fix for npm run stop to actually stop the running process

donaldgu:
Running npm run stop will now stop the running dev server process.  This now works from both the command line and the terminal window is VS.  When the process is started the process id is written to disk and then is able to be read when the user wants to stop the process.  

Also added tests to verify the new functions.
#160

### DIFF
--- a/packages/office-addin-debugging/src/debugInfo.ts
+++ b/packages/office-addin-debugging/src/debugInfo.ts
@@ -68,7 +68,6 @@ function readDebuggingInfo(pathToFile: string): any {
  * @param id process id
  */
 export async function saveDevServerProcessId(id: number): Promise<void> {
-    console.log(`Writing process id: ${id} to file: ${processIdFilePath}`);
     process.env.OfficeAddinDevServerProcessId = id.toString();
     const debuggingInfo = createDebuggingInfo();
     setDevServerProcessId(debuggingInfo, id);
@@ -83,12 +82,12 @@ export function readDevServerProcessId(): number | undefined {
     let id;
     if (process.env.OfficeAddinDevServerProcessId) {
         id = parseNumber(process.env.OfficeAddinDevServerProcessId);
-        console.log(`Process id read from env: ${id}`);
     } else if (fs.existsSync(processIdFilePath)) {
         const devServerProperties = readDebuggingInfo(processIdFilePath);
-        const pid = devServerProperties.devServer.processId;
-        id = parseNumber(pid.toString(), `Invalid process id found in ${processIdFilePath}`);
-        console.log(`Process id read from file: ${id}`);
+        if (devServerProperties.devServer && devServerProperties.devServer.processId) {
+            const pid = devServerProperties.devServer.processId;
+            id = parseNumber(pid.toString(), `Invalid process id found in ${processIdFilePath}`);
+        }
     }
     return id;
 }
@@ -98,7 +97,6 @@ export function readDevServerProcessId(): number | undefined {
  * and deletes the debug info json file if it exists
  */
 export function clearDevServerProcessId() {
-    console.log("Clearing dev server process id.");
     if (fs.existsSync(processIdFilePath)) {
         fs.unlinkSync(processIdFilePath);
     }
--- a/packages/office-addin-debugging/src/debugInfo.ts
+++ b/packages/office-addin-debugging/src/debugInfo.ts
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as fs from "fs";
+import { parseNumber } from "office-addin-cli";
+import * as os from "os";
+import * as path from "path";
+
+const processIdFile = "office-addin-debugging.json";
+const processIdFilePath = path.join(os.tmpdir(), processIdFile);
+
+export interface IDebugInfo {
+    devServer: IDevServerInfo;
+}
+
+export interface IDevServerInfo {
+    processId: number;
+}
+
+export function getDebuggingInfoPath(): string {
+    return processIdFilePath;
+}
+
+/**
+ * Creates the DebugInfo object
+ * @param id - process id of the dev server
+ */
+function createDebugInfoObject(id: number): IDebugInfo {
+    const devServer: IDevServerInfo = {
+        processId : id,
+    };
+    const debugInfo: IDebugInfo = {
+        devServer,
+    };
+    return debugInfo;
+}
+
+/**
+ * Read the DebugInfo object
+ * @param pathToFile - Path to the json file containing debug info
+ */
+function readDebugInfoObject(pathToFile: string): any {
+    const json = fs.readFileSync(pathToFile);
+    return JSON.parse(json.toString());
+}
+
+/**
+ * Saves the process id of the dev server
+ * @param id process id
+ */
+export async function saveDevServerProcessId(id: number): Promise<void> {
+    console.log(`Writing process id: ${id} to file: ${processIdFilePath}`);
+    process.env.OfficeAddinDevServerProcessId = id.toString();
+    const json = JSON.stringify(createDebugInfoObject(id), null, 4 );
+    fs.writeFileSync(processIdFilePath, json);
+}
+
+/**
+ * Read the process id from either the environment variable or
+ * from the json file containing the debug info
+ */
+export function readDevServerProcessId(): number | undefined {
+    let id;
+    if (process.env.OfficeAddinDevServerProcessId) {
+        id = parseNumber(process.env.OfficeAddinDevServerProcessId);
+        console.log(`Process id read from env: ${id}`);
+    } else if (fs.existsSync(processIdFilePath)) {
+        const devServerProperties = readDebugInfoObject(processIdFilePath);
+        const pid = devServerProperties.devServer.processId;
+        id = parseNumber(pid.toString(), `Invalid process id found in ${processIdFilePath}`);
+        console.log(`Process id read from file: ${id}`);
+    }
+    return id;
+}
+
+/**
+ * Deletes the environment variable containing process id
+ * and deletes the debug info json file if it exists
+ */
+export function clearDevServerProcessId() {
+    console.log("Clearing dev server process id.");
+    if (fs.existsSync(processIdFilePath)) {
+        fs.unlinkSync(processIdFilePath);
+    }
+    delete process.env.OfficeAddinDevServerProcessId;
+}
--- a/packages/office-addin-debugging/src/debugInfo.ts
+++ b/packages/office-addin-debugging/src/debugInfo.ts
@@ -5,6 +5,7 @@ import * as fs from "fs";
 import { parseNumber } from "office-addin-cli";
 import * as os from "os";
 import * as path from "path";
+import { debug } from "util";
 
 const processIdFile = "office-addin-debugging.json";
 const processIdFilePath = path.join(os.tmpdir(), processIdFile);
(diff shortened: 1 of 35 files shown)


packages/office-addin-debugging/src/start.ts
```diff
@@ -104,6 +108,11 @@ export async function runDevServer(commandLine: string, port?: number): Promise<
     }
 }
 
+export async function writeProcessIdToFile(id: number): Promise<void> {
```
akrantz: Suggest using the name that reflect more on the purpose of the function rather than its implementation. The purpose is to save the process id for the dev server. While it does this by writing to a file, that doesn't neccessarily need to be in the name. I would recommend having functions `saveDevServerProcessId(id: string)`, `readDevServerProcessId()` and `clearDevServerProcessId()`. That provides functions to save, read, and clear regardless of the implementation (as env var and/or in a file).

You can move the call to set the process id env var into the function, passing in `devServer.pid.toString()`.

packages/office-addin-debugging/src/start.ts
```diff
@@ -10,6 +11,8 @@ import * as nodeDebugger from "office-addin-node-debugger";
 import { getProcessIdsForPort } from "./port";
 import { startDetachedProcess, startProcess  } from "./process";
 
+export const processIdFile = "devServerProcessId.txt";
```
akrantz: I don't think we should just create the file in the current directory. It could stick around.

We might want a way to have a central place to put the file and have a mapping from the directory where run or from the add-in id to the devServerProcessId. This could be created in a temp folder.

packages/office-addin-debugging/src/stop.ts
```diff
@@ -41,5 +43,26 @@ export async function stopDebugging(manifestPath: string, unregisterCommandLine?
         delete process.env.OfficeAddinDevServerProcessId;
     }
 
+    const processIdReadFromFile = parseNumber(readProcessIdFromFile());
```
akrantz: As recommended before, use functions to read and clear the dev server process Id. I would move any parsing that is required into the function so that it returns "number | undefined".

packages/office-addin-debugging/src/stop.ts
```diff
@@ -29,7 +31,7 @@ export async function stopDebugging(manifestPath: string, unregisterCommandLine?
             console.log(`Unable to unregister the Office Add-in. ${err}`);
         }
     }
-
+ 
     if (process.env.OfficeAddinDevServerProcessId) {
```
akrantz: You can combine writing and reading the process id whether from env var or file.

packages/office-addin-debugging/test/test.ts
```diff
@@ -76,3 +79,21 @@ describe("port functions", function() {
...
+    it("writing process id file", async function() {
+      await start.writeProcessIdToFile(pid);
+      const processId = fs.readFileSync(start.processIdFile);
+      assert.strictEqual(processId.toString(), pid.toString());
+    });
+    it("reading process id file", async function() {
+      const id = stop.readProcessIdFromFile();
+      assert.strictEqual(id, pid.toString());
+    });
+    it("deleting process id file", async function() {
```
akrantz: need additional test cases:
* read from file but file doesn't exist 
* write to file when file already exists
* read to file that has unexpected contents
* current directory is not writable (error)

packages/office-addin-debugging/src/start.ts
donaldgu: I can make this change, but want to make sure this is needed.  I believe we are expecting the users to run npm run start and npm run stop from the same directory.  Is this correct?  Also with this design even if the file is not deleted after closing the process, the file will be overwritten the next time.  User should not get multiple files left around.  If we go with a design of a unique name for the file and it doesn't get deleted because the user manually closes the process, they can end up with multiple files left around. I could design it to delete all the files in the temp folder when starting the process and write a new file, thus making sure we don't have multiple files left around. This would be a little more complex, but doable. Let me know which design you prefer.

packages/office-addin-debugging/src/start.ts
```diff
@@ -8,10 +8,13 @@ import * as devSettings from "office-addin-dev-settings";
 import { DebuggingMethod } from "office-addin-dev-settings";
 import * as manifest from "office-addin-manifest";
 import * as nodeDebugger from "office-addin-node-debugger";
+import * as os from "os";
+import * as path from "path";
 import { getProcessIdsForPort } from "./port";
 import { startDetachedProcess, startProcess  } from "./process";
 
-export const processIdFile = "devServerProcessId.txt";
+const processIdFile = "devServerProcessId.txt";
```
akrantz: Suggest making the name ".office-addin-debugging" and in the file perhaps write it as a JSON object:

```
{
  "devServer": {
    "processId": 123
   }
}
```
That would be more extensible to having other info in there. 

Also it would better reflect that this is related to Office add-in debugging. 

Do you think this would be an easy change? What do you think?

akrantz: The variable names could be updated to reflect it, e.g. debugInfoFile.

akrantz: If you think it's overkill to do as a JSON object, then we could just do the file contents the same way but rename it to reflect that it is for debugging.

packages/office-addin-debugging/src/start.ts
```diff
@@ -8,10 +8,13 @@ import * as devSettings from "office-addin-dev-settings";
...
+export const processIdFilePath = path.join(os.tmpdir(), processIdFile);
```
akrantz: It would be better to export as a function, e.g. getDebuggingInfoPath(). It would also be good to add a JSDoc comment which would provide intellisense info.

akrantz: It might help to move functions related to reading, writing, deleting the debugging info file into its own file, and providing the functions related to it there. That can then be used by both start.ts and stop.ts

TCourtneyOwen:
It’s been a while since I looked this code. Is the issue that the old implementation didn’t allow for stopping via command line? I know we are able to to stop the dev server from within our UI tests so why is this change necessary?

akrantz:
The environment variable is not sufficient to communicate to the command line invocation of "npm start" and "npm stop". It needs to use a file to persist the info.

akrantz:
@donaldgu please update to the latest changes in master, and ping me when this is ready to be reviewed again.

packages/office-addin-debugging/test/test.ts
```diff
@@ -76,3 +80,50 @@ describe("port functions", function() {
...
+    assert.strictEqual(id, undefined);
+  });
+  it("write process id file that already exists", async function() {
+    const secondPid = 5678;
+    await debugInfo.saveDevServerProcessId(pid);
+    await debugInfo.saveDevServerProcessId(secondPid);
+    const json = fs.readFileSync(debugInfo.getDebuggingInfoPath());
+    const devServerInfo = JSON.parse(json.toString());
+    const processId = devServerInfo.devServer.processId;
+    assert.strictEqual(processId.toString(), secondPid.toString());
```
akrantz: do you need to use toString() here if they are both numbers?

donaldgu: Fixed. toString() is not needed.

packages/office-addin-debugging/src/debugInfo.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+export function getDebuggingInfoPath(): string {
+    return processIdFilePath;
+}
+
+/**
+ * Creates the DebugInfo object
+ */
+function createDebugInfo(): IDebugInfo {
```
akrantz: When I look at this, it seems like a new object can just be empty. I don't see a need for having the devServer there initially. The code when settings the devServer info can handle adding it if it is not there. 

I'm not even sure that you need this createDebugInfo() object. If the function just returns `{ }` then is it that useful?

donaldgu: I like the design of initializing the object with all the fields/properties once and updating the values as it needs to.  That way if we add new fields/properties in the future we do it in just one place.

packages/office-addin-debugging/src/debugInfo.ts
```diff
@@ -0,0 +1,103 @@
...
+
+import * as fs from "fs";
+import { parseNumber } from "office-addin-cli";
+import * as os from "os";
+import * as path from "path";
+
+const processIdFile = "office-addin-debugging.json";
+const processIdFilePath = path.join(os.tmpdir(), processIdFile);
+
+export interface IDebugInfo {
```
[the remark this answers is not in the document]
donaldgu: Done

packages/office-addin-debugging/src/debugInfo.ts
```diff
@@ -0,0 +1,103 @@
...
+export function getDebuggingInfoPath(): string {
+    return processIdFilePath;
+}
+
+/**
+ * Creates the DebuggingInfo object
+ */
+function createDebuggingInfo(): IDebuggingInfo {
+    const devServer: IDevServerInfo = {
+        processId : 0,
```
akrantz: undefined.

akrantz: or just make empty object.

packages/office-addin-debugging/src/debugInfo.ts
```diff
@@ -0,0 +1,103 @@
...
+    return debuggingInfo;
+}
+
+/**
+ * Set the process id of the dev server in the Debugging object
+ * @param debuggingInfo DebuggingInfo object
+ * @param id dev server process id
+ */
+function setDevServerProcessId(debuggingInfo: IDebuggingInfo, id: number) {
+    debuggingInfo.devServer.processId = id;
```
akrantz: what happens if devServer doesn't exist?

packages/office-addin-debugging/src/debugInfo.ts
```diff
@@ -0,0 +1,103 @@
...
+    const json = fs.readFileSync(pathToFile);
+    return JSON.parse(json.toString());
+}
+
+/**
+ * Saves the process id of the dev server
+ * @param id process id
+ */
+export async function saveDevServerProcessId(id: number): Promise<void> {
+    console.log(`Writing process id: ${id} to file: ${processIdFilePath}`);
```
akrantz: you can move writeDebuggingInfo() into its own function so read/write are the operations, and then this can function can modify the object and call write(). The loadDebuggingInfo() function really isn't needed and can be combined into the writeDebuggingInfo().

akrantz:
I ran `npm start` and `npm stop`. 
Let's not write the messages `Writing process id to file`, `Process id read from file` and `Clearing dev server process id`.

```

Debugging is being started...
App type: desktop
Enabled debugging for add-in 05c2e1c9-3e1d-406e-9a91-e9ac64854143. Debug method: 0
Starting the dev server... (webpack-dev-server --mode development)
Writing process id: 7032 to file: C:\Users\adamk\AppData\Local\Temp\office-addin-debugging.json
The dev server is running on port 3000. Process id: 7032
Sideloading the Office Add-in...
Debugging started.

c:\dev\git\Office-Addin-TaskPane>npm stop

> office-addin-taskpane@0.0.1 stop c:\dev\git\Office-Addin-TaskPane
> office-addin-debugging stop manifest.xml

Debugging is being stopped...
Process id read from file: 7032
Stopped dev server. Process id: 7032
Clearing dev server process id.
Debugging has been stopped.
```

akrantz:
If you edit the file so it is:

```
{
    "devServer": {
   
    }
}
```

Then do `npm stop`, it will cause an error:
```
Error: Unable to stop debugging.
TypeError: Cannot read property 'toString' of undefined
```

The code should be more resilient and not be an error if there is no "devServer" object or no "processId" property.

PR #162, opened 2019-07-11
Scripts relative path fix

asachin96:


packages/office-addin-dev-certs/src/install.ts
```diff
@@ -10,7 +11,7 @@ import { isCaCertificateInstalled, verifyCertificates } from "./verify";
 function getInstallCommand(caCertificatePath: string, machine: boolean = false): string {
    switch (process.platform) {
       case "win32":
-         return `powershell -ExecutionPolicy Bypass scripts\\install.ps1 ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'`;
+         return `powershell -ExecutionPolicy Bypass ${path.resolve(__dirname, "..\\scripts\\install.ps1")} ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'`;
```
akrantz: Suggestion, use separate variable to compose the script path to simplify reading the command line:

```
const script = `${path.resolve(__dirname, "..\\scripts\\install.ps1")}`;
return `powershell .....   ${script} ......`;
```

packages/office-addin-dev-certs/test/test.ts
```diff
@@ -143,7 +143,7 @@ describe("office-addin-dev-certs", function() {
                 sandbox.stub(verify, "isCaCertificateInstalled").returns(false);
                 await install.installCaCertificate(testCaCertificatePath, machine);
                 assert.strictEqual(execSync.callCount, 1);
-                assert.strictEqual(execSync.calledWith(`powershell -ExecutionPolicy Bypass scripts\\install.ps1 LocalMachine '${testCaCertificatePath}'`), true);
+                assert.strictEqual(execSync.calledWith(`powershell -ExecutionPolicy Bypass ${path.resolve(__dirname, "..\\scripts\\install.ps1")} LocalMachine '${testCaCertificatePath}'`), true);
```
akrantz: You could move the path to the script file to a `const script = ....` line at a higher scope so the variable can be shared for all of the test cases.

PR #163, opened 2019-07-11
Move test server listener code to separate function

TCourtneyOwen:
- Per suggestion of AKrantz: "There is no try/catch around the await to call reject on the Promise which is returned. Await on a rejected promise throws from that point. I highly suggest not mixing async/await code with callbacks that resolve/reject a Promise in the same function. Its too easy to miss things"
- Validated tests pass
- Also linked changes to Office-Addin-Taskpane and validated UI tests pass

### DIFF
--- a/packages/office-addin-test-server/src/testServer.ts
+++ b/packages/office-addin-test-server/src/testServer.ts
@@ -25,7 +25,7 @@ export class TestServer {
         this.m_testServerStarted = false;
     }
 
-    public async startTestServer(mochaTest: boolean = false): Promise<any> {
+    public async startTestServer(mochaTest: boolean = false): Promise<boolean> {
         try {
             if (mochaTest) {
                 process.env["NODE_TLS_REJECT_UNAUTHORIZED"] = "0";
@@ -59,8 +59,8 @@ export class TestServer {
         }
     }
 
-    private async startListening(): Promise<any> {
-        return new Promise<any>(async (resolve, reject) => {
+    private async startListening(): Promise<boolean> {
+        return new Promise<boolean>(async (resolve, reject) => {
             try {
                 // set server to listen on specified port
                 this.m_server.listen(this.m_port, () => {
--- a/packages/office-addin-test-server/src/testServer.ts
+++ b/packages/office-addin-test-server/src/testServer.ts
@@ -40,7 +40,7 @@ export class TestServer {
             return await this.startListening();
 
         } catch (err) {
-            throw new Error(`Test server failed with error.\n${err}`);
+            throw new Error(`Unable to start test server.\n${err}`);
         }
     }
 
@@ -57,7 +57,6 @@ export class TestServer {
                 const platformName = this.getPlatformName();
                 this.m_app.get("/ping", function (req: any, res: any, next: any) {
                     res.send(platformName);
-                    resolve(true);
                 });
 
                 // listen for posting of test results
@@ -70,7 +69,7 @@ export class TestServer {
                 });
 
             } catch (err) {
-                reject(new Error(`Test server failed with error.\n${err}`));
+                reject(new Error(`Unable to start test server.\n${err}`));
             }
         });
     }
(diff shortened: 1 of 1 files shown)


packages/office-addin-test-server/src/testServer.ts
```diff
@@ -24,46 +24,58 @@ export class TestServer {
...
                 this.m_server.listen(this.m_port, () => {
                     this.m_testServerStarted = true;
                     resolve(true);
                 });
+
+                // listen for 'ping'
+                const platformName = this.getPlatformName();
+                this.m_app.get("/ping", function (req: any, res: any, next: any) {
+                    res.send(platformName);
+                    resolve(true);
```
akrantz: I'm not clear why you call resolve here. Isn't the resolve call once the server is listening enough?

packages/office-addin-test-server/src/testServer.ts
```diff
@@ -24,46 +24,58 @@ export class TestServer {
...
+                });
+
+                // listen for posting of test results
+                this.m_resultsPromise = new Promise<JSON>(async (resolveResults) => {
```
akrantz: I don't think this code needs to go inside here.

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: yeah we probably don't need this - it wasn't there originally and doesn't seem required. I guess I was just doing it for consistency sake

packages/office-addin-test-server/src/testServer.ts
```diff
@@ -24,46 +24,58 @@ export class TestServer {
...
+                    this.m_app.post("/results", async (req: any, res: any) => {
+                        res.send("200");
+                        this.m_jsonData = JSON.parse(req.query.data);
+                        resolveResults(this.m_jsonData);
+                    });
+                });
+
             } catch (err) {
-                reject(new Error(`Unable to start test server.\n${err}`));
+                reject(new Error(`Test server failed with error.\n${err}`));
```
akrantz: I would change this back to "Unable to start test server".

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: but I think it can safely there and really it does fall in the category of a listening event

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: reverted

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: removed resolve here

packages/office-addin-test-server/src/testServer.ts
```diff
@@ -24,19 +24,21 @@ export class TestServer {
         this.m_resultsPromise = undefined;
         this.m_testServerStarted = false;
     }
-    
-    public async startTestServer(mochaTest: boolean = false): Promise<any>{
-        return new Promise<boolean>(async (resolve, reject) => {
+
+    public async startTestServer(mochaTest: boolean = false): Promise<any> {
```
akrantz: One thing I just noticed, this is "Promise<any>". Shouldn't this be "Promise<boolean>"?

packages/office-addin-test-server/src/testServer.ts
```diff
@@ -48,22 +50,31 @@ export class TestServer {
...
+
             // start listening on specified port
+            return await this.startListening();
+
+        } catch (err) {
+            throw new Error(`Unable to start test server.\n${err}`);
+        }
+    }
+
+    private async startListening(): Promise<any> {
```
akrantz: This should be "Promise<boolean>" as well.

TCourtneyOwen: fixed locally

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: resolved locally - will push commit

PR #174, opened 2019-07-19
eslint and prettier integration package

donaldgu:
cli to run eslint and prettier is as follows:

office-addin-lint lint -files <pathToFiles> This only reports back any issues found
office-addin-lint lint:fix -files This fixes any issues it can and reports back ones that need manual fixes
office-addin-lint prettier -files This fixes all prettier issues


packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+import * as commands from "./commands";
+import * as defaults from "./defaults";
+
+commander.name("office-addin-lint");
+commander.version(process.env.npm_package_version || "(version not available)");
+
+commander
+    .command("lint")
+    .option("--files <files>", `Path to files to perform lint check. Default: ${defaults.lintFiles}`)
+    .description(`Perform lint and prettier check`)
```
akrantz: This doesn't run Prettier, just lint. Suggest: `Check source files against lint rules`.

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+    .action(commands.lint);
+
+commander
+    .command("lint:fix")
+    .option("--files <files>", `Path to files to perform lint fix. Default: ${defaults.lintFiles}`)
+    .description(`Fix lint and prettier errors.`)
```
akrantz: Suggest: `Apply fixes to source based on lint rules.`

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+    .action(commands.lintFix);
+
+commander
+    .command("prettier")
+    .option("--files <files>", `Path to files to perform prettier. Default: ${defaults.lintFiles}`)
+    .description(`Fix all prettier issues.`)
```
akrantz: Suggest: `Make the source prettier.`

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+
+commander
+    .command("lint:fix")
+    .option("--files <files>", `Path to files to perform lint fix. Default: ${defaults.lintFiles}`)
+    .description(`Fix lint and prettier errors.`)
+    .action(commands.lintFix);
+
+commander
+    .command("prettier")
+    .option("--files <files>", `Path to files to perform prettier. Default: ${defaults.lintFiles}`)
```
akrantz: Instead of `Path to files to perform prettier.` => `Specifies which files to use.`  (keep the default as well)

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+
+commander
+    .command("lint")
+    .option("--files <files>", `Path to files to perform lint check. Default: ${defaults.lintFiles}`)
+    .description(`Perform lint and prettier check`)
+    .action(commands.lint);
+
+commander
+    .command("lint:fix")
+    .option("--files <files>", `Path to files to perform lint fix. Default: ${defaults.lintFiles}`)
```
akrantz: `Path to files to perform lint fix` => `Specifies the source files to fix.`  (keep the default)

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+import { logErrorMessage } from "office-addin-cli";
+import * as commands from "./commands";
+import * as defaults from "./defaults";
+
+commander.name("office-addin-lint");
+commander.version(process.env.npm_package_version || "(version not available)");
+
+commander
+    .command("lint")
+    .option("--files <files>", `Path to files to perform lint check. Default: ${defaults.lintFiles}`)
```
akrantz: `Path to files to perform lint check.` => `Specifies the source files to check.`  (keep the default)

packages/office-addin-lint/src/commands.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Determines path to files to run lint against. Priority order follows:
+ * 1 --files option passed through cli
+ * 2 lint_files config property in package.json
+ * 3 default location
+ * @param command command options which can contain files
+ */
+function getPathToFiles(command: commander.Command):string {
+    const pathToFiles: any = command.files ? command.files : process.env.npm_package_config_lint_files;
```
akrantz: I'm not sure if it will pick up the config from the correct package.json. It should come from the package where this is being used, not this package.json itself.

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+  return eslintCommand;
+}
+
+export function lintFix(files: string) {
+    console.log("Fixing lint issues for files: " + files);
+    const command = lintFixCommand(files);
+    execCommand(command);
+}
+
+export function prettierCommand(files: string): string {
```
akrantz: rename to `getPrettierCommand`

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import {logErrorMessage} from "office-addin-cli";
+import * as defaults from "./defaults";
+
+function execCommand(command: string) {
+    const execSync = require('child_process').execSync;
+    const child = execSync(command, { stdio: 'inherit' });
+}
+
+export function lintCheckCommand(files: string): string {
```
akrantz: rename to `getLintCheckCommand`

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+  const eslintCommand: string = "eslint " + files;
+  return eslintCommand;
+}
+
+export function lintCheck(files: string) {
```
akrantz: rename to `performLintCheck`.  I think the verb just helps it be more descriptive.

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+  console.log("Lint check for files: " + files);
+  const command = lintCheckCommand(files);
+  execCommand(command);
+}
+
+export function lintFixCommand(files: string): string {
```
akrantz: rename to `getLintFixCommand`

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+  const eslintCommand: string = "eslint --fix " + files;
+  return eslintCommand;
+}
+
+export function lintFix(files: string) {
```
akrantz: rename to `performLintFix`

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+    const command = lintFixCommand(files);
+    execCommand(command);
+}
+
+export function prettierCommand(files: string): string {
+  const prettierFixCommand: string = "prettier --parser typescript --write " + files;
+  return prettierFixCommand;
+}
+
+export function prettier(files: string) {
```
akrantz: rename to `makeFilesPrettier`

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,43 @@
...
+    console.log("Fixing prettier issues for files: " + files);
```
akrantz: We don't want console.log directly in the implementation. We could have it as part of the CLI, but if this is imported from another package, they could decide whether to do any console output. 

Along those lines we might want the command to capture console output and it could return it back to the CLI and the CLI could print it.

donaldgu: removing console.log calls

packages/office-addin-lint/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "office-addin-lint",
+  "version": "1.0.0",
+  "description": "Provide eslint and prettier integration",
+  "main": "./lib/main.js",
+  "scripts": {
+    "cli": "node lib/cli.js",
+    "tsc": "tsc -p tsconfig.json",
+    "build": "rimraf lib && concurrently \"npm run tsc\"",
+    "test2": "echo \"Error: no test specified\" && exit 1",
```
jayrathi: what is the test2 script for?

packages/office-addin-lint/test/src/test.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as prettierLint from "../../src/lint";
+
+describe("test cases", function() {
+    it("eslint and prettier commands", async function() {
+        const inputFile = "./test/cases/basic/functions.ts";
+        const lintExpectedCommand = "eslint ./test/cases/basic/functions.ts";
+        const lintFixExpectedCommand = "eslint --fix ./test/cases/basic/functions.ts";
+        const prettierExpectedCommand = "prettier --parser typescript --write ./test/cases/basic/functions.ts";
+        const lintCheckCommand = prettierLint.lintCheckCommand(inputFile);
+        assert.strictEqual(lintCheckCommand, lintExpectedCommand, "eslint lint command");
```
jayrathi: Can we have a message saying what is wrong instead of "eslint lint command". Suggest - "Lint command does not match expected value" or use the default message which will list out both values with AssertionError.

packages/office-addin-lint/test/src/test.ts
```diff
@@ -0,0 +1,23 @@
...
+        const lintFixCommand = prettierLint.lintFixCommand(inputFile);
+        assert.strictEqual(lintFixCommand, lintFixExpectedCommand, "eslint fix command");
```
jayrathi: Can we have a message saying what is wrong instead of "eslint fix command". Suggest - "Lint fix command does not match expected value" or use the default message which will list out both values with AssertionError.

packages/office-addin-lint/test/src/test.ts
```diff
@@ -0,0 +1,23 @@
...
+        const prettierCommand = prettierLint.prettierCommand(inputFile);
+        assert.strictEqual(prettierCommand, prettierExpectedCommand, "prettier command");
```
jayrathi: Can we have a message saying what is wrong instead of "prettier command". Suggest - "Prettier command does not match expected value" or use the default message which will list out both values with AssertionError.

packages/office-addin-lint/package.json
donaldgu: removed extra test2

packages/office-addin-lint/test/src/test.ts
donaldgu: updated

packages/office-addin-lint/test/src/test.ts
donaldgu: updated

packages/office-addin-lint/test/src/test.ts
donaldgu: updated

TCourtneyOwen:
I personally find it helpful to include a .vscode\launch.json file for debugging tests that would contain a config such as the following:

    {
      "type": "node",
      "request": "launch",
      "name": "Debug Tests",
      "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
      "args": [
        "-u",
        "tdd",
        "--timeout",
        "999999",
        "--colors",
        "${workspaceFolder}/test",
        "-r",
        "ts-node/register",
        "${workspaceFolder}/test/**/*.ts"
      ],
      "internalConsoleOptions": "openOnSessionStart"
    }

akrantz:
@donaldgu When running "npm link" in the office-addin-lint package dir to test this, I get these warnings:
```
npm WARN @typescript-eslint/eslint-plugin@1.12.0 requires a peer of eslint@^5.0.0 but none is installed. You must install peer dependencies yourself.
npm WARN @typescript-eslint/parser@1.12.0 requires a peer of eslint@^5.0.0 but none is installed. You must install peer dependencies yourself.
```

This appears to be because we're using eslint 6.0.0 but those packages declare a peer dependency for 5.x. Can you see if there is an updated package version, or perhaps we could make a PR to update the peer dependency? We want to avoid warnings like this if we can.

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+import * as commander from "commander";
+import { logErrorMessage } from "office-addin-cli";
+import * as commands from "./commands";
+import * as defaults from "./defaults";
+
+commander.name("office-addin-lint");
+commander.version(process.env.npm_package_version || "(version not available)");
+
+commander
+    .command("lint")
```
akrantz: I think we should make this the "check" command so that this is used via the package.json scripts as:
` "lint": "office-addin-lint check"`

packages/office-addin-lint/src/cli.ts
```diff
@@ -0,0 +1,43 @@
...
+    .option("--files <files>", `Specifies the source files to check. Default: ${defaults.lintFiles}`)
+    .description(`Check source files against lint rules.`)
+    .action(commands.lint);
+
+commander
+    .command("lint:fix")
```
akrantz: Along those same lines, this command would be "fix" so it is defined as ` "lint:fix": "office-addin-lint fix"`

packages/office-addin-lint/src/lint.ts
```diff
@@ -5,20 +5,18 @@ import * as path from "path";
 
 const esLintPath = require.resolve("eslint");
 const prettierPath = require.resolve("prettier");
-const config = require.resolve("office-addin-lint");
-const configDir = path.parse(config).dir;
 const esLintDir = path.parse(esLintPath).dir;
 const esLintFilePath = path.resolve(esLintDir, "\..\\bin/eslint.js");
 const prettierFilePath = path.resolve(prettierPath, "\..\\bin-prettier.js");
-const esLintConfigPath = path.resolve(configDir, "\..\\src/.eslintrc.json");
+const esLintConfigPath = path.resolve(__dirname, "..\\src\\.eslintrc.json");
```
akrantz: don't put the .eslintrc.json under src. Perhaps it could be in a config subdir instead.

if "config" and "src" are both dirs from the root, I would think this should use "../config/.eslintrc.json" to find the file. 

Remember the .ts will be compiled into the "lib" folder which is where the code will actually run from.

packages/office-addin-lint/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "lint_files": "src/**/*.{ts,tsx,js,jsx}"
+  },
+  "keywords": [
+    "Office",
+    "Office Add-in"
+  ],
+  "dependencies": {
+    "@typescript-eslint/eslint-plugin": "^1.13.0",
+    "@typescript-eslint/parser": "^1.13.0",
+    "assert": "^1.4.1",
```
akrantz: Does "assert" need to be a dependency?

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,51 @@
...
+const prettierFilePath = path.resolve(prettierPath, "\..\\bin-prettier.js");
+const esLintConfigPath = path.resolve(__dirname, "..\\config\\.eslintrc.json");
+
+function execCommand(command: string) {
+    const execSync = require("child_process").execSync;
+    const child = execSync(command, { stdio: "inherit" });
+}
+
+function getEsLintBaseCommand(): string {
+  const eslintBaseCommand: string = "node " + esLintFilePath + " -c " + esLintConfigPath + " --resolve-plugins-relative-to " + __dirname + " ";
```
akrantz: Please use back-tick string: 
```
`node ${esLintFilePath} -c ${esLintConfigPath} --resolve-plugins-relative-to ${__dirname}`;
```

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,51 @@
...
+  return eslintBaseCommand;
+}
+
+export function getLintCheckCommand(files: string): string {
+  const eslintCommand: string = getEsLintBaseCommand() + files;
```
akrantz: You can add the space before files here: 
```
const eslintCommand: string = `${getEsLintBaseCommand()} ${files}`;
```

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,51 @@
...
+  return eslintCommand;
+}
+
+export function performLintCheck(files: string) {
+  const command = getLintCheckCommand(files);
+  execCommand(command);
+}
+
+export function getLintFixCommand(files: string): string {
+  const eslintCommand: string =  getEsLintBaseCommand() + "--fix " + files;
```
akrantz: back-tick string here too:
```
const eslintCommand: string = `${getEsLintBaseCommand()} --fix ${files}`;
```

packages/office-addin-lint/src/lint.ts
```diff
@@ -0,0 +1,51 @@
...
+  return eslintCommand;
+}
+
+export function performLintFix(files: string) {
+    const command = getLintFixCommand(files);
+    execCommand(command);
+}
+
+export function getPrettierCommand(files: string): string {
+  const prettierFixCommand: string = "node " + prettierFilePath + " --parser typescript --write " + files;
```
akrantz: back-tick string here please...

packages/office-addin-lint/package.json
donaldgu: used in tests

akrantz: It should be a dev dependency then.

PR #175, opened 2019-07-19
support debugging when manifest file is in build output directory

zhngx:
-modify start and debug functions to support sideloading and debugging for sdx.


packages/office-addin-debugging/src/start.ts
```diff
@@ -58,6 +62,14 @@ export function parseAppType(text: string): AppType | undefined {
             return AppType.Desktop;
         case "web":
             return AppType.Web;
+        case "win32":
```
akrantz: This may be OK, or might be better as a platform enum. I was thinking that appType and platform are two separate values.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -7,6 +7,47 @@ import * as devSettings from "office-addin-dev-settings";
 import { AppType, parseAppType, parseDebuggingMethod, startDebugging } from "./start";
 import { stopDebugging } from "./stop";
 
+function getManifestPath(manifestPath: string, appTypeToDebug: string | undefined, prod: boolean): string {
+    let manifestPathToDebug: string = manifestPath;
+    try {
+        if (appTypeToDebug === undefined) {
```
jayrathi: It appTypeToDebug : undefined is not supported, why do we have it as an allowed type to the function?

packages/office-addin-debugging/src/commands.ts
```diff
@@ -7,6 +7,47 @@ import * as devSettings from "office-addin-dev-settings";
 import { AppType, parseAppType, parseDebuggingMethod, startDebugging } from "./start";
 import { stopDebugging } from "./stop";
 
+function getManifestPath(manifestPath: string, appTypeToDebug: string | undefined, prod: boolean): string {
```
jayrathi: It would be really useful to have some comments explaining the logic of this function. Like - under what conditions do we modify the manifestPath and how. Can you add that?

packages/office-addin-debugging/src/commands.ts
```diff
@@ -34,29 +75,34 @@ export async function start(manifestPath: string, appType: string | undefined, c
...
+        await startDebugging(manifestPathToDebug, appTypeToDebug, debuggingMethod, sourceBundleUrlComponents,
+            devServer, devServerPort, packager, packagerHost, packagerPort, sideloadCommand,
             enableDebugging, enableLiveReload);
     } catch (err) {
         logErrorMessage(`Unable to start debugging.\n${err}`);
     }
 }
 
-export async function stop(manifestPath: string, command: commander.Command) {
+export async function stop(manifestPath: string, appType: string | undefined, command: commander.Command) {
```
jayrathi: The appType parameter should not allow undefined. It looks like this function never handles it when it is undefined.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -86,7 +68,7 @@ export async function start(manifestPath: string, appType: string | undefined, c
             throw new Error("Please specify the application type to debug.");
         }
 
-        await startDebugging(manifestPath, appTypeToDebug, app, debuggingMethod, sourceBundleUrlComponents,
+        await startDebugging(manifestPathToDebug, appTypeToDebug, app, debuggingMethod, sourceBundleUrlComponents,
```
akrantz: Please don't rename `manifestPath`. The addition of "ToDebug" doesn't really add any meaning.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -95,7 +77,9 @@ export async function start(manifestPath: string, appType: string | undefined, c
 
 export async function stop(manifestPath: string, appType: string | undefined, command: commander.Command) {
     try {
-        await stopDebugging(manifestPath);
+        const appTypeToDebug: AppType | undefined = parseAppType(appType || process.env.npm_package_config_app_type_to_debug || "desktop");
```
akrantz: Since there is common code in start() and stop(), how about moving it to a determineAppType() function?

Are you planning to extend AppType to cover the specific platforms (e.g. "win32", etc.)?

akrantz:
Replaced by https://github.com/OfficeDev/Office-Addin-Scripts/pull/210.

PR #179, opened 2019-08-06
initial checkin for eslint config

donaldgu:
eslint config that can be used to extend config settings.  This is just the initial checkin to get the package setup.  Additional rules can be added later.


packages/eslint-config-addin/src/main.ts
```diff
@@ -0,0 +1,8 @@
+export = {
+    rules: {
+        "no-eval": "error",
```
akrantz: Are these the rules based on the security recommendations? It would be good to have the readme explain the rationale for our choices, although that's not needed to complete this check-in. 

Another question is whether our config should also extend a standard eslint config.

packages/eslint-config-addin/src/main.ts
```diff
@@ -0,0 +1,8 @@
...
+        "no-delete-var": "error",
+        "no-octal": "error",
+        "no-inner-declarations": "error"
+      }
+};
```
akrantz: Put a newline at the end of this line.

packages/eslint-config-addin/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+            "es6-promise"
+        ]
+    },
+    "include": [
+        "src/**/*",
+    ],
+    "exclude": [
+        "node_modules",
+        "lib",
+        "test/cases/**/*"
```
akrantz: You can remove the "test/cases/**/*" line since this doesn't apply.

packages/eslint-config-addin/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "eslint-config-addin",
+  "version": "1.0.0",
+  "description": "eslint configuration package for office-addin",
+  "main": "./lib/main.js",
+  "scripts": {
+    "tsc": "tsc -p tsconfig.json",
+    "build": "rimraf lib && concurrently \"npm run tsc\"",
+    "test": "echo \"Error: no test specified\" && exit 1"
```
akrantz: You can make this "echo \"No tests.\"" without the exit 1 so it's not an error.

packages/eslint-config-addin/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
```
TCourtneyOwen: Builds failed in CI pipeline with error "error TS2688: Cannot find type definition file for 'node'.  May need to to add @types/node to package.json

packages/eslint-config-addin/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "eslint-config-addin",
```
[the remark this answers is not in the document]
donaldgu: Per the recommended instructions for creating a shared config it needs to start with eslint-config.  https://eslint.org/docs/developer-guide/shareable-configs

packages/eslint-config-addin/src/main.ts
donaldgu: Yes these are the rules given to me.  With this config I was just focusing on the security issues as the standard configs are way more opinotated and are not related to security.

packages/eslint-config-addin/package.json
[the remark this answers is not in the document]
akrantz: OK, then how about `eslint-config-office-addins` as the name?

packages/eslint-config-addin/src/main.ts
akrantz: image

Let's have this extend from "eslint:recommended".  We can then see how it looks and tweak from there.

akrantz: We might need to also need to have more than one config, for differences with React or Angular.

donaldgu: added the extends from eslint:recommended.  Will look into created more configs based on the framework used.

packages/eslint-config-office-addins/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+        ]
+    },
+    "include": [
+        "src/**/*",
+    ],
+    "exclude": [
+        "node_modules",
+        "lib"
+    ]
+}
```
akrantz: nit: add a newline here

PR #180, opened 2019-08-06
use Yarn instead of NPM

akrantz:
@TCourtneyOwen This modifies the Lerna configuration to use Yarn instead of NPM.

This has the effect of hoisting the packages to the root node_modules folder. I'm not sure whether this is desirable or not, but wanted to experiment. Yarn is faster and the Yarn Workspaces provides the hoisting/symlinking of packages in the package manager which others seem to use. I'm not sure what the downsides are -- if any.


packages/custom-functions-metadata/package.json
```diff
@@ -32,7 +32,6 @@
   },
   "devDependencies": {
     "@types/assert": "^1.4.2",
-    "@types/commander": "^2.12.2",
```
TCourtneyOwen: should types/commander in fact be removed here?

packages/office-addin-lint/package.json
```diff
@@ -33,7 +33,6 @@
   },
   "devDependencies": {
     "@types/assert": "^1.4.2",
-    "@types/commander": "^2.12.2",
```
TCourtneyOwen: same question here about whether this is intentionally removed

packages/custom-functions-metadata/package.json
akrantz: Yes, yarn gives a warning that "commander" includes types, so "@types/commander" should not be used.

packages/office-addin-lint/package.json
akrantz: same answer.  :-)

TCourtneyOwen:
Looks like install fails in the CI loop.  Any idea what the problem is?

akrantz:
I'm not planning to merge this right now. That's why it is a draft PR.

akrantz:
Yes, you need to change the loop to use Yarn instead of NPM, e.g "yarn install", "yarn run build", etc.

This is one reason I think it would be good to move to yaml files because you can make PRs that include build pipeline changes as part of the PR.

PR #181, opened 2019-08-08
shared prettier config

donaldgu:
This creates a shared prettier config for Office-Addins. Packages that want to use it can add the following to package.json : "prettier": "office-addin-prettier-config"

### DIFF
--- a/packages/office-addin-prettier-config/prettier.config.js
+++ b/packages/office-addin-prettier-config/prettier.config.js
@@ -1,7 +1,3 @@
 module.exports = {
-    bracketSpacing: true,
-    printWidth: 100,
-    semi: false,
-    singleQuote: true,
-    trailingComma: 'all',
-  }
+  printWidth: 120
+}
--- a/packages/office-addin-prettier-config/prettier.config.js
+++ b/packages/office-addin-prettier-config/prettier.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+    bracketSpacing: true,
+    printWidth: 100,
+    semi: false,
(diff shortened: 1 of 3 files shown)


packages/office-addin-prettier-config/prettier.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    bracketSpacing: true,
+    printWidth: 100,
```
akrantz: Let's use the defaults. We want semi-colons and double-quotes. Trailing commas should use the default of none, otherwise it will put trailing commas in function call parameters which is weird. Bracket spacing defaults should also be good.

printWidth is the one option that I felt we may want to override to be 100 as 80 felt like it caused functions to wrap to having parameters on separate lines, which I personally don't like. I'd like to experiment to know more.

akrantz: I just looked at fabric-internal, and they use 120 for printWidth.

akrantz: https://github.com/OfficeDev/office-ui-fabric-react uses 140 for printWidth.

akrantz: Based on this, I think we should make printWidth to be 120 as a starting value.

akrantz: I went ahead and made these changes.

akrantz: @donaldgu Are you OK with these changes?

PR #184, opened 2019-08-12
Azure DevOps build pipeline

akrantz:
Use yaml files for the Azure DevOps pipeline.


.azure-devops/build-pipeline.yaml
```diff
@@ -0,0 +1,26 @@
+# Build a general Node.js project with npm.
+# Add steps that analyze code, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/javascript
+
+trigger:
+- master
+
+jobs:
+- job: Windows
+  pool:
```
TCourtneyOwen: Should we add "OPX Windows 1903 Dogfood Pool" as well?

akrantz: It's using that now.

PR #185, opened 2019-08-12
Fix issue where dev-server doesn't start for project w/ spaces

TCourtneyOwen:
- Update command returned by getInstallCommand()
- Update tests to reflect new command path
- Add ca to getHttpsServerOptions()
- Update tests to reflect addition of ca being returned from getHttpsServerOptions()


packages/office-addin-dev-certs/src/install.ts
```diff
@@ -12,7 +12,7 @@ function getInstallCommand(caCertificatePath: string, machine: boolean = false):
    switch (process.platform) {
       case "win32":
          const script = path.resolve(__dirname, "..\\scripts\\install.ps1");
-         return `powershell -ExecutionPolicy Bypass ${script} ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'`;
+         return `powershell -ExecutionPolicy Bypass "& '${script}' ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'`;
```
akrantz: What does the `"&` here do? I'm unclear if the double-quote is escaping the ampersand and what this means for the Powershell command line?

akrantz: The single quotes around `{script}` should allow it to work even if the script path contains a space character, which I presume is the key part of the fix here.

packages/office-addin-dev-certs/src/defaults.ts
```diff
@@ -9,6 +9,8 @@ export const certificateDirectoryName = ".office-addin-dev-certs";
 export const certificateDirectory =  path.join(os.homedir(), certificateDirectoryName);
 export const caCertificateFileName = "ca.crt";
 export const caCertificatePath = path.join(certificateDirectory, caCertificateFileName);
+export const localhostCaCertificateFileName = "ca.crt";
```
akrantz: `localhostCaCertificateFileName` and `localhostCaCertificatePath` seems to duplicate the same thing that is already provided by `caCertificateFileName` and `caCertificatePath`.

packages/office-addin-dev-certs/src/httpsServerOptions.ts
```diff
@@ -14,6 +15,12 @@ export async function getHttpsServerOptions(): Promise<IHttpsServerOptions> {
     await ensureCertificatesAreInstalled();
 
     const httpsServerOptions = {} as IHttpsServerOptions;
+    try {
+        httpsServerOptions.ca = fs.readFileSync(defaults.localhostCaCertificatePath);
```
akrantz: Should just be `defaults.caCertificatePath` once the duplicative `defaults.localhostCaCertificatePath` is removed.

packages/office-addin-dev-certs/src/httpsServerOptions.ts
```diff
@@ -14,6 +15,12 @@ export async function getHttpsServerOptions(): Promise<IHttpsServerOptions> {
     await ensureCertificatesAreInstalled();
 
     const httpsServerOptions = {} as IHttpsServerOptions;
+    try {
+        httpsServerOptions.ca = fs.readFileSync(defaults.localhostCaCertificatePath);
+    } catch (err) {
+        throw new Error(`Unable to read the ca certificate file.\n${err}`);
```
akrantz: Make the `ca` uppercase. `Unable to read the CA certificate file.`  CA stands for Certificate Authority.

packages/office-addin-dev-certs/src/httpsServerOptions.ts
TCourtneyOwen: caCertifatePath resulted in test failures. Found I had to add this new default

packages/office-addin-dev-certs/src/install.ts
TCourtneyOwen: Single quotes didn’t work. Tried many variations and this is the only one I found that worked. Found it in stack overflow

packages/office-addin-dev-certs/src/httpsServerOptions.ts
akrantz: Aren't these the same? It would be better to fix the tests that introduce this. Sachin can help with the test failures...

packages/office-addin-dev-certs/src/install.ts
TCourtneyOwen: single quotes around {script} results in the following error:
Error: Unable to install the CA certificate. At line:1 char:91
+ ... dules\office-addin-dev-certs\scripts\install.ps1' CurrentUser 'C:\Use ...
+                                                       ~~~~~~~~~~~
Unexpected token 'CurrentUser' in expression or statement.

Regarding what the & means: https://stackoverflow.com/questions/22074507/what-does-the-symbol-in-powershell-mean

And it was this stack overflow article where I got the idea to use the &:  https://stackoverflow.com/questions/18537098/spaces-cause-split-in-path-with-powershell

akrantz: There is a mix of single and double-quotes. Should there also be the double-quote in front of the ampersand?

If this is normalized to use double-quotes and the ampersand on its own, I would expect:
```
return `powershell -ExecutionPolicy Bypass & "${script}" ${machine ? "LocalMachine" : "CurrentUser"} "${caCertificatePath}"`;
```

Could you let me know if that would be OK or not?

packages/office-addin-dev-certs/src/defaults.ts
TCourtneyOwen: I guess I don't need to include localhostCaCertificateFileName and localhostCaCertificatePath. I could've sworn I was getting a test failure without this change but must have been some other reason.  Removing them

packages/office-addin-dev-certs/src/httpsServerOptions.ts
TCourtneyOwen: I guess I don't need to include localhostCaCertificateFileName and localhostCaCertificatePath. I could've sworn I was getting a test failure without this change but must have been some other reason. Removing them

TCourtneyOwen: Removing them

packages/office-addin-dev-certs/src/install.ts
TCourtneyOwen: yes, this works as well:
powershell -ExecutionPolicy Bypass & "${script}" ${machine ? "LocalMachine" : "CurrentUser"} '${caCertificatePath}'

I can make that change

akrantz: OK, and the last set of quotes should be double-quotes not single-quotes. (You probably just mistyped it here.)

TCourtneyOwen: OK updating and pushing change

akrantz:
@TCourtneyOwen I merged master into this branch. Just letting you know in case you want to fetch the changes locally.

PR #187, opened 2019-08-14
office-addin-dev-settings: fix sideload command

akrantz:
Need better logic for the case where the is an app specified and the manifest contains multiple hosts.


akrantz:
@TCourtneyOwen Please review.

PR #188, opened 2019-08-14
fix office-addin-dev-settings dependencies

akrantz:
fs-extra package should be a runtime dependency.


akrantz:
@TCourtneyOwen Simple fix needed. Please review and approve.

TCourtneyOwen:
Does this fix the side loading issue we discussed

akrantz:
Yes, I did not see this when symlinking locally as my install had the dev dependencies. But when dev-settings package is run via the debugging package, it doesn't have the dev dependencies.

PR #189, opened 2019-08-15
office-addin-dev-settings: register command should write full path

akrantz:
On Windows, if the register command is called with a relative path or just a file name,
the full path should be written into the registry.


akrantz:
@TCourtneyOwen Please review.

TCourtneyOwen:
Please address test failure in CI loop

akrantz:
There was a timeout in one test case. I don't see why it should have taken more than 2 seconds. I'm retrying it.

akrantz:
@TCourtneyOwen Tests passed on retry.

PR #191, opened 2019-08-15
Enable http port for test server

TCourtneyOwen:
Update test server to start http port as well as https

NOTE: I opted not to add an http port as requirement for the test server constructor because this allows us not to have to update all are projects that currently use the test server.  They should continue to work with these changes. I can however add a required http port if we feel we should (honestly it seems like we should)

- remove process.env.NODE_TLS_REJECT_UNAUTHORIZED from test server code.  Tests can now set this.
- Update test helper code to add ability to ping http and https
- Add setTLSRejectUnauthorize method to test helpers


packages/office-addin-test-helpers/test/test.ts
```diff
@@ -43,5 +49,5 @@ async function _sendTestData(): Promise<boolean> {
     testData[valueKey] = testValue;
     testValues.push(testData);
 
-    return testHelper.sendTestResults(testValues, port);
+    return testHelper.sendTestResults(testValues, defaultHttpsPort);
```
akrantz: shoudn't this just have `useHttps: boolean = true` in the sendTestResults method? Then you don't need this extra parameter when calling it here. Is there any reason to sendTestResults() using HTTP and not HTTPS?

packages/office-addin-test-helpers/test/test.ts
```diff
@@ -4,24 +4,30 @@
 import * as assert from "assert";
 import * as mocha from "mocha";
 import * as testHelper from "../src/testHelpers";
-import { TestServer } from "../../office-addin-test-server";
+import { defaultHttpPort, defaultHttpsPort, TestServer } from "../../office-addin-test-server";
```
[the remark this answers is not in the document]
akrantz: OK, these are port numbers. Instead, please make defaults.ts which has "httpPort" and "httpsPort". Then you can import * as defaults from "defaults"; and use "defaults.httpPort" or "defaults.httpsPorts".  main.ts can export * from "defaults".

packages/office-addin-test-helpers/test/test.ts
```diff
@@ -4,24 +4,30 @@
...
         it("Test server should have responded to ping", async function () {
-            const testServerResponse: object = await testHelper.pingTestServer(port);
-            assert.equal(testServerResponse["status"], 200);
-            assert.equal(testServerResponse["platform"], testServer.getPlatformName());
+            // ping test server on https port
+            const httpsTestServerResponse: object = await testHelper.pingTestServer(defaultHttpsPort);
+            assert.equal(httpsTestServerResponse["status"], 200);
+            assert.equal(httpsTestServerResponse["platform"], testServer.getPlatformName());
+            // ping test server on http port
+            const httpTestServerResponse: object = await testHelper.pingTestServer(defaultHttpPort, false /* https */);
```
akrantz: The fact that you have `false ?* https */` here along with defaultHttpPort concerns me. Why have two different parameter to say whether or not to use https? I think it would be better to make pingTestServer just have a useHttps: boolean parameter, and it can know that it needs to use: `const port = useHttps ? defaults.httpsPort : defaults.httpPort;`

packages/office-addin-test-helpers/test/test.ts
akrantz: I see this is a port, but it would be better to have `useHttps: boolean` parameter here instead of the port. You need to change both the procotol and matching port number, right?

packages/office-addin-test-server/src/testServer.ts
```diff
@@ -3,37 +3,40 @@
...
+    private httpPort: number;
+    private httpsPort: number;
     private testServerStarted: boolean;
     private app: express.Express;
     private resultsPromise: Promise<JSON>;
-    private server: https.Server;
+    private httpServer: http.Server;
+    private httpsServer: https.Server;
 
     constructor(port: number) {
```
akrantz: why can you change https port but not http port?

akrantz: This would be better as `constructor(httpsPort: number = defaults.httpsPort, httpPort: number = defaults.httpPort)`

This means that they automatically use the default unless explicitly overridden. If the defaults are OK, then you can just use `new TestServer()`

packages/office-addin-test-server/src/testServer.ts
```diff
@@ -3,37 +3,40 @@
...
         this.jsonData = {};
-        this.port = port;
+        this.httpsPort = port;
+        this.httpPort = defaultHttpPort;
         this.resultsPromise = undefined;
         this.testServerStarted = false;
     }
 
-    public async startTestServer(mochaTest: boolean = false): Promise<boolean> {
+    public async startTestServer(mochaTest = false /* deprecated */): Promise<boolean> {
```
akrantz: Use JSDoc comments to annotate this.
/** 
  * @param mochaTest @deprecated
 * 
*/

It would be nice to have all exported functions have JSDoc comments so you get intellisense.

akrantz: You'll have to check if that syntax actually works. I think it does but I could be wrong.

packages/office-addin-test-helpers/src/testHelpers.ts
```diff
@@ -39,3 +39,7 @@ export async function sendTestResults(data: object, port: number = defaultPort):
     });
 }
 
+export function setTLSRejectUnauthorized(on: boolean) {
```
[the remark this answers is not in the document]
TCourtneyOwen: we can remove the function and just have it as a method in the test file

packages/office-addin-test-helpers/test/test.ts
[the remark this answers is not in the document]
TCourtneyOwen: created defaults.ts in office-addin-test-server with default httpPort and httpsPort consts

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: Good idea - I will add JSDoc comments

packages/office-addin-test-helpers/test/test.ts
TCourtneyOwen: made suggested change locally

packages/office-addin-test-helpers/test/test.ts
TCourtneyOwen: the only problem is that this will be a breaking change that will require udpate to all the tests in all the project repos.  Come to think of it, the suggested change to pingTestServer to take a https:boolean parameter is also a breaking change. I will like to avoid this if possible

packages/office-addin-test-server/src/testServer.ts
TCourtneyOwen: fixed locally

packages/office-addin-test-helpers/test/test.ts
TCourtneyOwen: I am going to hold off on changing this so I don't have to update all the projects to use the boolean useHttps instead of the port

TCourtneyOwen:
Abandoning PR due to merge conflicts

PR #193, opened 2019-08-16
office-addin-dev-certs: fix install

akrantz:
The ampersand should not be in the command. It causes it to execute as two separate commands
 and opens the script file rather than executing it.


akrantz:
@TCourtneyOwen Please review. I went ahead and merged this so I can quickly publish the fix.

PR #195, opened 2019-08-16
run powershell script using -File option

akrantz:


akrantz:
@TCourtneyOwen I found that this appears the right thing to use to actually get it to run.

PR #196, opened 2019-08-16
office-addin-dev-certs: fix uninstall and verify commands too

akrantz:


akrantz:
@TCourtneyOwen This should hopefully do it...

akrantz:
@TCourtneyOwen I validated that this has fixed it when run in a path with spaces.

TCourtneyOwen:
Interesting that all the test pass with the change I made. Maybe are tests need improvement?

TCourtneyOwen:
“Our tests”

akrantz:
The tests only verify that the powershell command is formatted as expected. It does not execute the command. The reason the command is not executed is that it requires user input. You have to test this manually to ensure that it works.

TCourtneyOwen:
I know in the first iteration of my changes I checked to see if the dev-server successfully started.  It could be after revising that PS command I neglected to check and that's where the problem surfaced

PR #198, opened 2019-08-16
Fix for sideloading race condition

TCourtneyOwen:
We have had an issue recently where UI tests fail in the CI loops because sideloading is occurring too soon before the dev build is done and thus the task pane fails to load.  These changes a check to see if the dev build is complete by doing a fetch call to the default SourceLocation and waiting for a successful response before sideloading.  I built and linked these changes on a CI VM and they look to address the failure we were seeing

NOTE: I initially thought the error was related to enabling EdgeWebView on the VMs, but turns out that's not the problem, as we already set this up as part of VM deployment

### DIFF
--- a/packages/office-addin-dev-settings/package-lock.json
+++ b/packages/office-addin-dev-settings/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "office-addin-dev-settings",
-  "version": "1.7.5",
+  "version": "1.7.6",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


TCourtneyOwen:
Come to think of it, it might make more sense to move this to runDevServer in office-addin-debugging

TCourtneyOwen:
closing this PR and moving this code over to office-addin-debugging

akrantz:
Ha! I didn't see your message after clicking the email to review this. Glad you already realized this.

PR #201, opened 2019-08-21
Add new call to check that the dev server build is complete

TCourtneyOwen:
This is an additional safeguard to ensure that sideloading doesn't initiate before the dev build is completed


packages/office-addin-debugging/src/start.ts
```diff
@@ -35,6 +35,15 @@ export async function isDevServerRunning(port: number): Promise<boolean> {
     return isRunning;
 }
 
+export async function isDevBuildComplete(sourceLocation: string): Promise<boolean> {
```
akrantz: You could name this to reflect what it is actually doing, which is just seeing if the specified sourceLocation url can be fetched. You could make the parameter name simply "url", and call it isUrlAvailable().

packages/office-addin-debugging/src/start.ts
```diff
@@ -100,6 +109,16 @@ export async function runDevServer(commandLine: string, port?: number): Promise<
                 } else {
                     throw new Error(`The dev server is not running on port ${port}.`);
                 }
+
+                const manifestInfo: any = await readManifestFile(manifestPath);
+                const sourceLocation: string = manifestInfo.defaultSettings.sourceLocation;
+                const isBuildComplete: boolean = await waitUntilDevBuildIsComplete(sourceLocation);
```
akrantz: Per my previous comment, you can update naming so it is `const isAvailable: boolean = await waitUntilUrlIsAvailable(sourceLocation);`

packages/office-addin-debugging/src/start.ts
```diff
@@ -100,6 +109,16 @@ export async function runDevServer(commandLine: string, port?: number): Promise<
...
+
+                if (isBuildComplete) {
+                    console.log(`The dev build completed successfully and fetch of dev source at ${sourceLocation} was successful`);
```
akrantz:
```
if (isUrlAvailable) {
   console.log(`Url is available: ${sourceLocation}`);
} else {
   throw new Error(`The source location is not available from the dev server: ${sourceLocation}`);
}

File: packages/office-addin-debugging/src/start.ts [hunk 4]
```diff
@@ -214,7 +233,7 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
     }
 
     if (devServerCommandLine) {
-        devServerPromise = runDevServer(devServerCommandLine, devServerPort);
+        devServerPromise = runDevServer(devServerCommandLine, manifestPath, devServerPort);
```
akrantz: A new optional parameter should be added to the end, so it is not a breaking api change.

packages/office-addin-debugging/src/start.ts
```diff
@@ -267,6 +286,10 @@ export async function waitUntil(callback: (() => Promise<boolean>), retryCount:
     return done;
 }
 
+export async function waitUntilDevBuildIsComplete(sourceLocation: string, retryCount: number = 30, retryDelay: number = 1000): Promise<boolean> {
```
akrantz: Along lines of previous comments, change name to `waitUntilUrlIsAvailalbe` and change `sourceLocation` to `url`.

packages/office-addin-debugging/src/start.ts
TCourtneyOwen: made change locally

akrantz:
@TCourtneyOwen What's the status of this PR?

TCourtneyOwen:
@akrantz I think it should be good to go. I will go ahead and merge it

PR #204, opened 2019-08-23
Update package.json to include relevant keywords for better discoverability

asachin96:


packages/office-addin-dev-certs/package.json
```diff
@@ -17,7 +17,14 @@
     "office-addin-dev-certs": "./cli.js"
   },
   "keywords": [
-    "Office"
+    "Office",
+    "SSL",
+    "https",
+    "Certificates",
+    "Self-Signed",
+    "devlopment",
```
TCourtneyOwen: "development"

packages/office-addin-dev-certs/package.json
```diff
@@ -17,7 +17,14 @@
     "office-addin-dev-certs": "./cli.js"
   },
   "keywords": [
-    "Office"
+    "Office",
+    "SSL",
+    "https",
+    "Certificates",
```
TCourtneyOwen: Any particular reason for capitalizing Certificates and Self-Signed?

asachin96: updated

asachin96:
@TCourtneyOwen windows machines are not available in CI?

packages/office-addin-dev-certs/package.json
```diff
@@ -17,7 +17,14 @@
     "office-addin-dev-certs": "./cli.js"
   },
   "keywords": [
```
akrantz: Should "Add-in" be added in? :-)

Some of these like "development", "server", and "secure" are very generic and seem of little value for search purposes.  

"self-signed" seems misleading, since they aren't self-signed.

"localhost" might also be good to add.

asachin96: If people search like "development server certificate" this package should be suggested right. Generally speaking, I've seen people adding so many keywords for better discoverability. (Eg https://www.npmjs.com/package/https-pem). I'll add "add-in" and "localhost" and remove "self-signed"

TCourtneyOwen:
Looking into why Windows agents aren't available. I see that the VMs were successfully created last night but that the agents are listed as offline. I wonder if there could be some broader outage

TCourtneyOwen:
I am noticing that all the custom artifacts we created failed to install last night. I am betting there is some connectivity issue between Azure and the repo where our custom artifacts are stored

TCourtneyOwen:
@asachin96 you job now passes

asachin96:
Yep. Can please close the review

PR #208, opened 2019-09-03
integrate office-addin-lint

donaldgu:
Integrate the office-addin-lint package to the custom functions metadata package.  Add the commands to run the linting and apply the changes from running the linting and prettier.


packages/custom-functions-metadata/src/cli.ts
```diff
@@ -12,7 +12,9 @@ commander.version(process.env.npm_package_version || "(version not available)");
 
 commander
   .command("generate <source-file> <metadata-file>")
-  .description("Generate the metadata for the custom functions from the source code.")
+  .description(
```
akrantz: Is this on separate lines because of the line length?

packages/custom-functions-metadata/package.json
```diff
@@ -40,6 +43,7 @@
     "@types/xregexp": "^3.0.29",
     "concurrently": "^3.6.1",
     "mocha": "^5.2.0",
+    "office-addin-lint": "^1.0.13",
     "rimraf": "^2.6.2",
     "ts-node": "^8.3.0",
     "tslint": "^5.18.0",
```
akrantz: shouldn't you remove tslint?

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -151,43 +151,53 @@ type CustomFunctionsSchemaDimensionality = "invalid" | "scalar" | "matrix";
...
-                if (wantConsoleOutput) {
-                    console.error(err);
-                }
-                throw new Error(`Error writing: ${outputFileName} : ${err}`);
-            }
-        } else if (wantConsoleOutput) {
-            console.error("Errors in file: " + inputFile);
-            errors.forEach((err) => console.error(err));
+export async function generate(
+  inputFile: string,
```
akrantz: I personally dislike parameters on different lines.

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -196,157 +206,208 @@ export async function generate(inputFile: string, outputFileName: string, wantCo
...
+          );
+          const isInvocationFunction = hasInvocationParameter(
+            lastParameter,
+            jsDocParamTypeInfo
+          );
+
+          const parametersToParse =
+            isStreamingFunction || isCancelableFunction || isInvocationFunction
+              ? functionDeclaration.parameters.slice(
+                  0,
```
akrantz: yuck, if the line length is longer, would it prevent this? I wish there was a way to wrap across multiple lines but put as many parameters per line as fits.

akrantz:
Just to comment, instead of setting the prettier line length to 120 locally, the office-addin-prettier-config package has 120 as the line length, so this needs to be using that config by default.

packages/custom-functions-metadata/package.json
akrantz: @donaldgu see above.   :-)

packages/custom-functions-metadata/src/commands.ts
```diff
@@ -8,17 +7,17 @@ import * as generateMetadata from "./generate";
...
-      }
-      if (!outputFile) {
-        throw new Error("You need to provide the path to the output file for the custom functions metadata.");
-      }
-      const results = await generateMetadata.generate(inputFile, outputFile);
-      if (results.errors.length > 0) {
-        console.error("Errors found:" );
-        results.errors.forEach((err) => console.log(err));
-      }
+    if (!inputFile) {
```
akrantz: This looks better now that it is using the prettier config with the 120 line length.

packages/custom-functions-metadata/src/generate.ts
donaldgu: The parameters are on different lines because they exceed the 120 limit.  I don't want to increase the limit and don't know of another change in prettier we could make to allow the parameters to remain on the same line.

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -151,43 +151,53 @@ type CustomFunctionsSchemaDimensionality = "invalid" | "scalar" | "matrix";
...
+  if (fs.existsSync(inputFile)) {
+    const sourceCode = fs.readFileSync(inputFile, "utf-8");
+    const parseTreeResult: IParseTreeResult = parseTree(sourceCode, inputFile);
+    parseTreeResult.extras.forEach(extra =>
+      extra.errors.forEach(err => errors.push(err))
+    );
+    generateResults.associate = [...parseTreeResult.associate];
+
+    if (errors.length === 0) {
+      const json = JSON.stringify(
```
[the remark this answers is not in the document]
donaldgu: Issue is the same as it exceeds the 120 limit.  I'm going to leave it as is for now.  If this becomes an issue when others start implementing the package we will look into increasing the limit.

PR #210, opened 2019-09-03
Implement sideloading for bundled manifest

TCourtneyOwen:
- Replace app-type with platform param
- If manifestPath is an empty string, determine the manifest path via the platform and new --prod option. This allows for the creation the correct manifest for SDX's where the manifest is copied to the dist folder as part of build
- Remove the --unload option for stop, which as far as I can tell is no longer used
- Tested this to ensure backwards compat and to ensure it can construct the manifest path correctly when the manifest is in the dist folder

### DIFF
--- a/packages/office-addin-debugging/README.md
+++ b/packages/office-addin-debugging/README.md
@@ -59,12 +59,11 @@ Host name of the packager. Default: `localhost`.
 
 ` --packager-port <port>`
 
-
 Port number of the packager. Default: `8081`.
 
 ` --prod`
 
-Specifies that debugging session is for production mode. Default is dev mode.
+Specifies that debugging session is for production mode. Default is development mode.
 
 ` --sideload <command>`
 
@@ -105,7 +104,7 @@ Options:
 
 ` --prod`
 
-Specifies that debugging session is for production mode. Default is dev mode.
+Specifies that debugging session is for production mode. Default is development mode.
 
 #
 
--- a/packages/office-addin-debugging/README.md
+++ b/packages/office-addin-debugging/README.md
@@ -59,6 +59,10 @@ Host name of the packager. Default: `localhost`.
 
 ` --packager-port <port>`
 
+Specifies that debugging session is for production mode. Default is dev mode.
+
+` --prod`
+
 Port number of the packager. Default: `8081`.
 
 ` --sideload <command>`
@@ -98,9 +102,9 @@ Syntax:
 
 Options:
 
-`--unload <command>`
+Specifies if in production mode. Default is dev mode.
 
-Unregister the Office Add-in using the specified command. For example: `office-toolbox remove -m <manifest> -a <app>`.
- 
+` --prod`
+
 #
 
--- a/packages/office-addin-debugging/README.md
+++ b/packages/office-addin-debugging/README.md
@@ -59,11 +59,12 @@ Host name of the packager. Default: `localhost`.
 
 ` --packager-port <port>`
 
-Specifies that debugging session is for production mode. Default is dev mode.
+
+Port number of the packager. Default: `8081`.
 
 ` --prod`
 
-Port number of the packager. Default: `8081`.
+Specifies that debugging session is for production mode. Default is dev mode.
 
 ` --sideload <command>`
 
@@ -102,9 +103,9 @@ Syntax:
 
 Options:
 
-Specifies if in production mode. Default is dev mode.
-
 ` --prod`
+
+Specifies that debugging session is for production mode. Default is dev mode.
 
 #
 
--- a/packages/office-addin-debugging/src/commands.ts
+++ b/packages/office-addin-debugging/src/commands.ts
@@ -2,11 +2,26 @@
 // Licensed under the MIT license.
 
 import * as commander from "commander";
+import * as fs from "fs";
 import { logErrorMessage, parseNumber } from "office-addin-cli";
 import * as devSettings from "office-addin-dev-settings";
 import { OfficeApp, parseOfficeApp } from "office-addin-manifest";
-import { AppType, parseAppType, parseDebuggingMethod, startDebugging, parseAppPlatform, AppPlaform } from "./start";
+import { AppPlatform, AppType, parseAppPlatform, parseAppType, parseDebuggingMethod, startDebugging } from "./start";
 import { stopDebugging } from "./stop";
+
+function determineManifestPath(platform: AppPlatform, dev: boolean): string {
+    let manifestPath = "";
+    if (process.env.npm_package_config_manifest_location) {
+        const manifestPathToDebug = process.env.npm_package_config_manifest_location;
+        manifestPath = manifestPathToDebug.replace("${flavor}", dev ? "dev" : "prod");
+        manifestPath = manifestPath.replace("${platform}", platform);
+    }
+
+    if (manifestPath === "" || !fs.existsSync(manifestPath)) {
+        throw new Error(`manifest path ${manifestPath} is not valid.`);
+    }
+    return manifestPath;
+}
 
 function parseDevServerPort(optionValue: any): number | undefined {
     const devServerPort = parseNumber(optionValue, "--dev-server-port should specify a number.");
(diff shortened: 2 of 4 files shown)


packages/office-addin-debugging/README.md
```diff
@@ -59,8 +59,13 @@ Host name of the packager. Default: `localhost`.
 
 ` --packager-port <port>`
 
+
 Port number of the packager. Default: `8081`.
 
+` --prod`
+
+Specifies that debugging session is for production mode. Default is dev mode.
```
akrantz: "Default is **development** mode."

packages/office-addin-debugging/README.md
```diff
@@ -98,9 +103,9 @@ Syntax:
 
 Options:
 
-`--unload <command>`
+` --prod`
+
+Specifies that debugging session is for production mode. Default is dev mode.
```
akrantz: "Default is **development** mode."  Although does this actually get used?

packages/office-addin-debugging/src/commands.ts
```diff
@@ -2,12 +2,27 @@
...
 import * as commander from "commander";
+import * as fs from "fs";
 import { logErrorMessage, parseNumber } from "office-addin-cli";
 import * as devSettings from "office-addin-dev-settings";
 import { OfficeApp, parseOfficeApp } from "office-addin-manifest";
-import { AppType, parseAppType, parseDebuggingMethod, startDebugging } from "./start";
+import { AppPlatform, AppType, parseAppPlatform, parseAppType, parseDebuggingMethod, startDebugging } from "./start";
 import { stopDebugging } from "./stop";
 
+function determineManifestPath(platform: AppPlatform, dev: boolean): string {
```
akrantz: Please rename AppPlatform to just Platform.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -2,12 +2,27 @@
...
+    let manifestPath = "";
+    if (process.env.npm_package_config_manifest_location) {
```
akrantz: please use a blank line between variables and code. 
please use a bank line before the return line.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -2,12 +2,27 @@
...
+        const manifestPathToDebug = process.env.npm_package_config_manifest_location;
+        manifestPath = manifestPathToDebug.replace("${flavor}", dev ? "dev" : "prod");
```
akrantz: manifestPathToDebug could be undefined and would cause an exception on .replace(). 

Rather than use manifestPathToDebug as const variable and manifestPath as let variable, you could just use the manifestPath variable.

Better error messages could be provided.

```
let manifestPath = process.env.npm_package_config_manifest_location || "";

manifestPath.replace(...);

if (!manifestPath) {
  throw new Error(`The manifest path was not provided.`);
}

if (!fs.existsSync(manifestPath) {
  throw new Error(`The manifest path does not exist: ${manifestPath}.`);
}

return manifestPath;
```

packages/office-addin-debugging/src/commands.ts
```diff
@@ -23,14 +38,13 @@ function parseDevServerPort(optionValue: any): number | undefined {
     return devServerPort;
 }
 
-export async function start(manifestPath: string, appType: string | undefined, command: commander.Command) {
+export async function start(manifestPath: string, platform: string | undefined, command: commander.Command) {
     try {
-        const appTypeToDebug: AppType | undefined = parseAppType(appType || process.env.npm_package_config_app_type_to_debug || "desktop");
+        const appPlatformToDebug: AppPlatform | undefined = parseAppPlatform(platform || process.env.npm_package_config_app_platform_to_debug || AppPlatform.Win32);
```
akrantz: Please rename to `platformToDebug`. 
Please use `npm_package_config_platform_to_debug`,

packages/office-addin-debugging/src/commands.ts
```diff
@@ -23,14 +38,13 @@ function parseDevServerPort(optionValue: any): number | undefined {
...
+        const appTypeToDebug: AppType | undefined = parseAppType(appPlatformToDebug || process.env.npm_package_config_app_type_to_debug || AppType.Desktop);
         const appToDebug: string | undefined = command.app || process.env.npm_package_config_app_to_debug;
         const app: OfficeApp | undefined = appToDebug ? parseOfficeApp(appToDebug) : undefined;
-        const sourceBundleUrlComponents = new devSettings.SourceBundleUrlComponents(
-            command.sourceBundleUrlHost, command.sourceBundleUrlPort,
-            command.sourceBundleUrlPath, command.sourceBundleUrlExtension);
+        const devBuild: boolean = command.prod ? false : true;
```
akrantz: Can this just be `const dev: boolean`? It would be consistent with other places.

I keep wondering about using something more like isDevMode or using an enum such as BuildFlavor or Mode = { Development, Production } but it does seem simpler to use a boolean. I think "dev" is OK if used consistently.

The replacement parameter is ${BuildFlavor} but I wonder if ${Mode} is better. Of course we could use "--mode production" which is more like webpack, but that is probably overkill. Anyway these various thoughts keep going through my head, but the simplest "dev: boolean" and "--prod" seems OK. I'm curious if you have any opinions.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -39,20 +53,41 @@ export async function start(manifestPath: string, appType: string | undefined, c
         const packager: string | undefined = command.packager || process.env.npm_package_scripts_packager;
         const packagerHost: string | undefined = command.PackagerHost || process.env.npm_package_config_packager_host;
         const packagerPort: string | undefined = command.PackagerPort || process.env.npm_package_config_packager_port;
+        const sourceBundleUrlComponents = new devSettings.SourceBundleUrlComponents(
+            command.sourceBundleUrlHost, command.sourceBundleUrlPort,
+            command.sourceBundleUrlPath, command.sourceBundleUrlExtension);
+
+        if (appPlatformToDebug === undefined) {
+            throw new Error("Please specify the application platform to debug.");
```
akrantz: Please omit "application". `Please specify the platform to debug.` is sufficient.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -39,20 +53,41 @@ export async function start(manifestPath: string, appType: string | undefined, c
...
+        }
 
         if (appTypeToDebug === undefined) {
             throw new Error("Please specify the application type to debug.");
         }
 
+        if (manifestPath === "") {
```
akrantz: if (!manifestPath)

packages/office-addin-debugging/src/commands.ts
```diff
@@ -39,20 +53,41 @@ export async function start(manifestPath: string, appType: string | undefined, c
...
             devServer, devServerPort, packager, packagerHost, packagerPort, enableDebugging, enableLiveReload);
     } catch (err) {
         logErrorMessage(`Unable to start debugging.\n${err}`);
     }
 }
 
-export async function stop(manifestPath: string, command: commander.Command) {
+export async function stop(manifestPath: string, platform: string | undefined, command: commander.Command) {
     try {
+        const appPlatformToDebug: AppPlatform | undefined = parseAppPlatform(platform || process.env.npm_package_config_app_plaform_to_debug || AppPlatform.Win32);
```
akrantz: same comments about not needed App/application with platform; parsePlatform(), "Please specify the platform to debug.".

packages/office-addin-debugging/src/commands.ts
```diff
@@ -39,20 +53,41 @@ export async function start(manifestPath: string, appType: string | undefined, c
...
+        const devBuild: boolean = command.prod ? false : true;
+
+        if (appPlatformToDebug === undefined) {
+            throw new Error("Please specify the application platform to debug.");
+        }
+
+        if (manifestPath === "") {
+            manifestPath = determineManifestPath(appPlatformToDebug, devBuild);
+        }
         await stopDebugging(manifestPath);
```
akrantz: nit: blank line after if statement (before await line)

packages/office-addin-debugging/src/start.ts
```diff
@@ -11,6 +11,15 @@ import * as debugInfo from "./debugInfo";
 import { getProcessIdsForPort } from "./port";
 import { startDetachedProcess  } from "./process";
 
+export enum AppPlatform {
```
akrantz: rename to Platform and move to keep in alphabetic order.

packages/office-addin-debugging/src/start.ts
```diff
@@ -52,9 +61,32 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
...
     switch (text) {
         case "desktop":
+        case "macos":
+        case "win32":
             return AppType.Desktop;
         case "web":
             return AppType.Web;
+        case "ios":
+        case "android":
+            throw new Error(`Platform type ${text} not currently supported for debugging`);
```
akrantz: I think should parse AppType even for ios and android (as Desktop). Instead of having unsupported platforms fail the parse, you should check the enum value where appropriate.

packages/office-addin-debugging/src/start.ts
```diff
@@ -52,9 +61,32 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
...
+        default:
+            return undefined;
+    }
+}
+
+export function parseAppPlatform(text: string): AppPlatform | undefined {
+    switch (text) {
+        case "desktop":
+            return process.platform === "win32" ? AppPlatform.Win32 : AppPlatform.MacOS;
```
akrantz: Use a switch. For "win32" and "darwin" return Platform.Win32 / MacOS. For the default case, throw new Error(`The current platform is not supported: ${process.platform}`);

packages/office-addin-debugging/src/start.ts
```diff
@@ -52,9 +61,32 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
...
+        case "macos":
+            return AppPlatform.MacOS;
+        case "web":
```
akrantz: It would be better to put the case statements in alphabetic order.

packages/office-addin-debugging/src/start.ts
```diff
@@ -52,9 +61,32 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
...
+            return AppPlatform.Web;
+        case "win32":
+            return AppPlatform.Win32;
+        case "ios":
+        case "android":
+            throw new Error(`Platform type ${text} not currently supported for debugging`);
```
akrantz: As I said before, implement this for ios and android. Move the error check elsewhere.

packages/office-addin-debugging/README.md
TCourtneyOwen: we need to be able to contruct the manifest for SDX's so we would need to know the mode (prod or dev)

TCourtneyOwen: I need to rethink how stop is going to work

akrantz: @TCourtneyOwen Can you change "dev mode" to "development mode" in the help text?

PR #211, opened 2019-09-04
Fix office addin dev settings tests

TCourtneyOwen:
Fix unit tests so they don't hang on enabling EdgeWV loopback
- Tests were hanging in CI loop due to prompt surfacing for enabling Edge Webview loopback
- Set command.yes to true to bypass prompt

### DIFF
--- a/packages/office-addin-dev-settings/test/integration-tests/test.ts
+++ b/packages/office-addin-dev-settings/test/integration-tests/test.ts
@@ -12,20 +12,20 @@ import * as commands from "../../src/commands";
 const addinId = "9982ab78-55fb-472d-b969-b52ed294e173";
 
 if (process.platform === "win32") { // only windows is supported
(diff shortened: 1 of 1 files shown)


packages/office-addin-dev-settings/test/integration-tests/test.ts
```diff
@@ -25,7 +23,9 @@ if (process.platform === "win32") { // only windows is supported
       sandbox.restore();
     });
     it("loopback already enabled", async function() {
+      const command: commander.Command = new commander.Command();
```
akrantz: The issue here is really that loopback isn't already enabled on the test machine. The test never enables it originally, and because it is not enabled it is showing the prompt. This fixes the test to enable it without the prompt, but doesn't really configure setup or cleanup to save the current state, reset the state so it is not enabled, then have a test case to enable it.

PR #212, opened 2019-09-06
Update event and error reporting to capture whether test data

TCourtneyOwen:
also add instrumentation key to default.ts

### DIFF
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -111,7 +111,7 @@ export class OfficeAddinUsageData {
       usageDataEvent.name = `${eventName}${this.options.isForTesting ? "-test" : ""}`;
       try {
         for (const [key, [value, elapsedTime]] of Object.entries(data)) {
-          usageDataEvent.properties[key] = value;
+          usageDataEvent.properties[key] = `${value}${this.options.isForTesting ? "-test" : ""}`;
           usageDataEvent.measurements[key + " durationElapsed"] = elapsedTime;
         }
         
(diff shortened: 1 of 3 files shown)


TCourtneyOwen:
looking into failing test

TCourtneyOwen:
updated event properties to include isForTesting

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -108,15 +108,14 @@ export class OfficeAddinUsageData {
   public async reportEventApplicationInsights(eventName: string, data: object): Promise<void> {
     if (this.getUsageDataLevel() === UsageDataLevel.on) {
       const usageDataEvent = new appInsights.Contracts.EventData();
-      usageDataEvent.name = eventName;
+      usageDataEvent.name = `${eventName}${this.options.isForTesting ? "-test" : ""}`;
```
akrantz: This is minor but it would be cleaner to write the code this way:
```
usageDataEvent.name = this.options.isForTesting ? "${eventName}-test" : eventName;
```

TCourtneyOwen: OK I will make that change but every incremental change of course requires verification

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -108,15 +108,14 @@ export class OfficeAddinUsageData {
...
       try {
         for (const [key, [value, elapsedTime]] of Object.entries(data)) {
-          usageDataEvent.properties[key] = value;
+          usageDataEvent.properties[key] = `${value}${this.options.isForTesting ? "-test" : ""}`;
```
akrantz: I'm confused about this. Shouldn't this be adding ` "isForTesting" = isForTesting ` after the passed in values in data?

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -108,15 +108,14 @@ export class OfficeAddinUsageData {
...
           usageDataEvent.measurements[key + " durationElapsed"] = elapsedTime;
```
akrantz: Similarly, shouldn't this set the key "durationElapsed" = elapsedTime after the for loop?

PR #214, opened 2019-09-10
Update isUrlAvailable to include agent and to use https.request

TCourtneyOwen:
- Adding an agent addresses the problem where we were getting a failure with "reason: self signed certificate in certificate chain"

Testing Performed:
Tried out these changes linked to Office-Addin-Taskpane after running npm update --dev

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import * as https from "https";
 import * as fetch from "node-fetch";
 import * as devCerts from "office-addin-dev-certs";
 import * as devSettings from "office-addin-dev-settings";
@@ -60,7 +61,17 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
 
 export async function isUrlAvailable(url: string): Promise<boolean> {
     try {
-        await fetch.default(url);
+        const serverOptions = await devCerts.getHttpsServerOptions();
+        const options: any = {
+            cert: serverOptions.cert,
+            key: serverOptions.key,
+            method: "GET",
+        };
+        options.agent = new https.Agent(options);
+        const request = https.request(url, options);
+        request.on("error", () => {
+            return false;
+        });
         return true;
     } catch {
         return false;
@@ -97,7 +108,7 @@ export function parsePlatform(text: string): Platform | undefined {
     if (text === AppType.Desktop) {
         text = process.platform;
     }
-    
+
     switch (text) {
         case "android":
             return Platform.Android;


packages/office-addin-debugging/src/start.ts
```diff
@@ -60,7 +61,17 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
...
-        await fetch.default(url);
+        const serverOptions = await devCerts.getHttpsServerOptions();
+        const options: any = {
+            cert: serverOptions.cert,
+            key: serverOptions.key,
+            method: "GET",
+        };
+        options.agent = new https.Agent(options);
+        const request = https.request(url, options);
+        request.on("error", () => {
```
akrantz: This doesn't look right to me. It will immediately return true before the async "error" event would have a chance to return false. It would seem that for both the success and error case, there needs to be an async wait to know if the request was successful or not.

packages/office-addin-debugging/src/start.ts
```diff
@@ -60,7 +61,17 @@ export async function isPackagerRunning(statusUrl: string): Promise<boolean> {
 
 export async function isUrlAvailable(url: string): Promise<boolean> {
     try {
-        await fetch.default(url);
+        const serverOptions = await devCerts.getHttpsServerOptions();
+        const options: any = {
```
akrantz: Please don't use "any". Do you even need to specify a type? Isn't it inferred from the assigned value?

packages/office-addin-debugging/src/start.ts
TCourtneyOwen: I actually tested to ensure it was hitting the request.on(error), which it does the first couple times around because the url isn't available yet, so it return false the first couple times

packages/office-addin-debugging/src/start.ts
TCourtneyOwen: I couldn't get this to compile unlesss I use any

packages/office-addin-debugging/src/start.ts
TCourtneyOwen: But I will look into it some more tomorrow

TCourtneyOwen: On second thought I think you may be right. Will investigate more tomorrow. I’d like to get a fix in soon so we can update the templates with all the office-addin-scripts changes. Honestly I’m not even sure we need isUrlAvailable. This was put in originally to address test failures which have since been obviated by providing a dev port when calling startDebugging. That said it’d be nice to get it working

TCourtneyOwen: @akrantz I think this is now working as expected.  I tested the latest implementation with Office-Addin-Taskpane

TCourtneyOwen:
Might be tricky to implement tests for this given that we have to start a dev server and then have a valid url to ping.

akrantz:
I think you can have unit tests which mock fetch, and then make sure that the mocking just reflects the actual behavior of fetch, e.g. when an exception is thrown.

TCourtneyOwen:
This is far more work than I have time for. I simply won’t be able to allot time for creating mocks etc. I suggest we remove the function entirely if you want all this stuff, particularly given that we really don’t need this function for our tests now.

akrantz:
If you cannot adequately test it, then you should remove it.

TCourtneyOwen:
I will remove it then

PR #225, opened 2019-09-29
eslint plugin for office-addins

donaldgu:
This creates an eslint plugin for office-addins to use.  It has 3 configs which can be used.

 "extended": [
        "plugin:office-addins/recommended",
        "plugin:office-addins/react",
        "plugin:office-addins/reactnative"
    ]

### DIFF
--- a/packages/eslint-plugin-office-addins/README.md
+++ b/packages/eslint-plugin-office-addins/README.md
@@ -4,10 +4,12 @@ eslint plugin for office-addins
 
 ## Installation
 
-You'll first need to install [ESLint](http://eslint.org):
+This plugin is designed to work with the office-addin-lint package. 
+
+Install `office-addin-lint`
 
 ```
-$ npm i eslint --save-dev
+$ npm i office-addin-lint --save-dev
 ```
 
 Next, install `eslint-plugin-office-addins`:
@@ -15,8 +17,6 @@ Next, install `eslint-plugin-office-addins`:
 ```
 $ npm install eslint-plugin-office-addins --save-dev
 ```
-
-**Note:** If you installed ESLint globally (using the `-g` flag) then you must also install `eslint-plugin-office-addins` globally.
 
 ## Usage
 
--- a/packages/eslint-plugin-office-addins/README.md
+++ b/packages/eslint-plugin-office-addins/README.md
@@ -30,23 +30,15 @@ Add `office-addins` to the plugins section of your `.eslintrc` configuration fil
 }
 ```
 
-Then configure the extends property.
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin-office-addins/README.md
```diff
@@ -0,0 +1,52 @@
...
+    "plugins": [
+        "office-addins"
+    ]
+}
+```
+
+Then configure the extends property.
+```json
+{
+    "extended": [
```
akrantz: This doesn't look right. You say "extends" property but it says "extended" in the JSON.  Shouldn't there just also only be one plugin? I think it could be more clear about just choosing one (recommended), and choosing a different one if React or React Native is being used.

packages/eslint-plugin-office-addins/README.md
```diff
@@ -0,0 +1,52 @@
...
+        "plugin:office-addins/recommended",
+        "plugin:office-addins/react",
+        "plugin:office-addins/reactnative"
+    ]
+}
+```
+
+Then configure the rules you want to use under the rules section.
```
akrantz: This makes it seem like rules need to be specified. But this is only if you want to override rules. I don't think this section is needed, as if someone wants to modify eslint rules, they would do so based on the eslint documentation. Maybe instead you should just say to see the eslint docs for more info about how to modify rules if desired and provide a link.

packages/eslint-plugin-office-addins/src/main.ts
```diff
@@ -0,0 +1,34 @@
...
+            parserOptions: {
+                ecmaFeatures: {
+                    jsx: true
+                  }
+            },
+            extends: ['plugin:react-native/all'],
+            rules: {}
+        }
+    }
+  };
```
akrantz: It is better to have a newline on the last line. GitHub diff shows the fact that one is missing.

akrantz:
Is this intended to replace **eslint-config-office-addins**? Do we need both "config" and "plugin" packages?

packages/eslint-plugin-office-addins/README.md
```diff
@@ -0,0 +1,44 @@
+# eslint-plugin-office-addins
+
+eslint plugin for office-addins
+
+## Installation
+
+You'll first need to install [ESLint](http://eslint.org):
```
akrantz: This seems wrong. Shouldn't users install office-addin-lint rather than use eslint directly?

TCourtneyOwen:
Does the build for this pass locally? It's failing in the CI loop?

PR #238, opened 2019-12-02
Add new package for Office-Addin-Taskpane-SSO scenario

TCourtneyOwen:
- This package configures a taskpane addin to use SSO and also starts the SSO service to make graph API calls on behalf of the taskpane application


packages/office-addin-sso/src/error-handler.ts
```diff
@@ -0,0 +1,45 @@
+import { showMessage } from "./message-helper";
+
+export function handleClientSideErrors(error: any): boolean {
+  let invokeFallBackDialog: boolean = false;
+  switch (error.code) {
+    case 13001:
+      // No one is signed into Office. If the add-in cannot be effectively used when no one
+      // is logged into Office, then the first call of getAccessToken should pass the
+      // `allowSignInPrompt: true` option. Since this sample does that, you should not see this error
```
Rick-Kirkham: Need to remove the last sentence of this comment: "Since this sample does that, you should not see this error"

packages/office-addin-sso/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+    "compilerOptions": {
+        "target": "es6",
```
Rick-Kirkham: If the tsconfig **complilerOptions.target** is "es6", will the add-in work on IE 11?

packages/office-addin-sso/src/error-handler.ts
TCourtneyOwen: fixed

packages/office-addin-sso/tsconfig.json
TCourtneyOwen: I assume this should be OK, as this is the same setting we have in tsconfig for all other packages in Office-Addin-Scripts

TCourtneyOwen:
@Rick-Kirkham , regarding this comment : Should we say somewhere (the readme?) that the dev cannot work with a tenancy that's configured to require MFA?

I think we should add that to the documentation @kbrandl  will write up for the project template and not put it in the package readme

PR #242, opened 2020-01-09
Add usage data for SSO configuration and server processes

TCourtneyOwen:
- Note, I am running into problems when trying to add usage data for files that the SSO taskpane app imports.  Webpack fails to build with errors related to the applicationInsights package that office-addin-usage-data consumes

Sample data available at https://ms.portal.azure.com#@72f988bf-86f1-41af-91ab-2d7cd011db47/blade/Microsoft_Azure_Monitoring_Logs/LogsBlade/resourceId/%2Fsubscriptions%2Ffb31a92f-824a-4de5-bf31-722222f81009%2FresourceGroups%2FOffice-Addin-Usage-Data%2Fproviders%2Fmicrosoft.insights%2Fcomponents%2FOffice-Addin-Usage-Data/source/LogsBlade.AnalyticsShareLinkToQuery/q/H4sIAAAAAAAAA0suLS7Jz3UtS80rKeaqUcjJzM0sUTA1MACyyzNSi1IV8hJzUxVsbRWU8tPSMpNTdRNTUjLzdIuL85UUuAAtgoKQPQAAAA%253D%253D/timespan/P1D

### DIFF
--- a/packages/office-addin-sso/src/authRoute.ts
+++ b/packages/office-addin-sso/src/authRoute.ts
@@ -7,6 +7,7 @@
 import * as express from 'express';
 import * as fetch from 'node-fetch';
 import * as form from 'form-urlencoded';
+import * as usageDataHelper from './usagedata-helper';
 
 export class AuthRouter {
     public router: express.Router;
@@ -45,11 +46,20 @@ export class AuthRouter {
                         }
                     });
                     const json = await tokenResponse.json();
-
                     res.send(json);
+
+                    // Send usage data
+                    const usageDataInfo: Object = {
+                        Method: ['authRouter'],
+                        scope: [scopeName],
+                        Platform: [process.platform],
+                        Succeeded: [true]
+                    }
+                    usageDataHelper.sendUsageDataCustomEvent(usageDataInfo);
                 }
                 catch (error) {
                     res.status(500).send(error);
+                    usageDataHelper.sendUsageDataException('authRouter', error);
                 }
             }
         });
--- a/packages/office-addin-sso/src/configure.ts
+++ b/packages/office-addin-sso/src/configure.ts
@@ -71,8 +71,6 @@ async function createNewApplication(ssoAppName: string, port: string, userJson: 
             if (await isUserTenantAdmin(userJson)){
                 await grantAdminContent(applicationJson);
                 await setTenantReplyUrls(applicationJson);
-            } else {
-                console.log(chalk.yellow("You are not a tenant admin so you cannot grant admin consent for your application.  Contact your tenant admin to grant consent"));
             }
 
             // Create an application secret and add to the credential store
--- a/packages/office-addin-sso/src/configure.ts
+++ b/packages/office-addin-sso/src/configure.ts
@@ -14,6 +14,9 @@ let usageDataInfo: Object = {};
 require('dotenv').config();
 
 export async function configureSSOApplication(manifestPath: string, port: string) {
+    // Log start time for configuration process
+    const startSsoCongiguration = (new Date()).getTime();
+
     // Check to see if Azure CLI is installed.  If it isn't installed, then install it
     const cliInstalled = await isAzureCliInstalled();
 
@@ -41,7 +44,19 @@ export async function configureSSOApplication(manifestPath: string, port: string
         await logoutAzure();
 
         console.log(chalk.green(`Application with id ${applicationJson['appId']} successfully registered in Azure.  Go to https://ms.portal.azure.com/#home and search for 'App Registrations' to see your application`));
-        usageDataHelper.sendUsageDataSuccessEvent('configureSSOApplication');
+
+        // Log end time for configuration process and compute in seconds
+        const endSsoCongiguration = (new Date()).getTime();
+        const durationSsoConfiguration = (endSsoCongiguration - startSsoCongiguration) /1000
+        
+        // Send usage data
+        usageDataInfo = {
+            Method: ['configureSSOApplication'],
+            configDuration: [durationSsoConfiguration],
+            Platform: [process.platform],
+            Succeeded: [true]
+        }
+        usageDataHelper.sendUsageDataCustomEvent(usageDataInfo);
     }
     else {
         usageDataHelper.sendUsageDataException('configureSSOApplication', 'Login to Azure did not succeed');
@@ -139,7 +154,8 @@ export async function isAzureCliInstalled(): Promise<boolean> {
                 usageDataInfo = {
                     Method: ['isAzureCliInstalled'],
                     cliInstalled: [cliInstalled],
-                    Platform: [process.platform]
+                    Platform: [process.platform],
+                    Succeeded: [true]
                 }
                 usageDataHelper.sendUsageDataCustomEvent(usageDataInfo);
                 return cliInstalled;
--- a/packages/office-addin-sso/src/configure.ts
+++ b/packages/office-addin-sso/src/configure.ts
@@ -9,11 +9,14 @@ import * as defaults from './defaults';
 import * as fs from 'fs';
 import { addSecretToCredentialStore, writeApplicationData } from './ssoDataSettings';
 import { ManifestInfo, readManifestFile } from 'office-addin-manifest';
+import * as usageDataHelper from './usagedata-helper';
+let usageDataInfo: Object = {};
 require('dotenv').config();
 
 export async function configureSSOApplication(manifestPath: string, port: string) {
     // Check to see if Azure CLI is installed.  If it isn't installed, then install it
     const cliInstalled = await isAzureCliInstalled();
+
     if(!cliInstalled) {
         console.log(chalk.yellow("Azure CLI is not installed.  Installing now before proceeding"));
         await installAzureCli();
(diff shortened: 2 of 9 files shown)


akrantz:
It sounds like there is an issue with office-addin-usage-data. Are you going to investigate that?

TCourtneyOwen:
@akrantz I checked and office-addin-usage-data is using the latest version of the appInsights package. I will continue investigating but don't want this to hold up the PR because we are planning to go live Monday.  @nokafor, do you agree?

TCourtneyOwen:
I think the tests are failing because they are hanging on the dev-certs dialog. Investigating

nokafor:
I think it's ok to limit the data to just the configuration and server processes, and when we do the telemetry pass on all our tools, we can fill in the gaps and make sure any outlying issues are handled. At minimum, when we ship the initial version, I'd like us to know that: someone ran this tool, how long it took, and was it successful. @TCourtneyOwen is appinsights failing on any of those?

TCourtneyOwen:
@LouMM and I were talking and there's an @microsoft/applicationinsights-web" package that might solve the webpack issues I was seeing.  I am going to locally try this out, and see if this fixes things. If it does, then we would need to make changes to the office-addin-usage-data package.  But I'd rather not hold up the train for SSO while we make this change.  Of course, the @microsoft/applicationinsights-web may not fix the problem at all, but I will see

packages/office-addin-sso/src/authRoute.ts
```diff
@@ -7,6 +7,7 @@
 import * as express from 'express';
 import * as fetch from 'node-fetch';
 import * as form from 'form-urlencoded';
+import * as usageDataHelper from './usagedata-helper';
```
akrantz: What is usagedata-helper? Is this a dependency in package.json? Why not just `import * as usageData from `office-addin-usage-data`? Also, you don't need `./` here.

packages/office-addin-sso/src/configure.ts
```diff
@@ -9,11 +9,17 @@ import * as defaults from './defaults';
 import * as fs from 'fs';
 import { addSecretToCredentialStore, writeApplicationData } from './ssoDataSettings';
 import { ManifestInfo, readManifestFile } from 'office-addin-manifest';
+import * as usageDataHelper from './usagedata-helper';
+let usageDataInfo: Object = {};
 require('dotenv').config();
 
 export async function configureSSOApplication(manifestPath: string, port: string) {
+    // Log start time for configuration process
+    const startSsoCongiguration = (new Date()).getTime();
```
akrantz: misspelling. I recommmend using `startTime` or `ssoConfigStartTime`. Phrasing as `startSsoConfig` makes it a verb that implies it is going to start something, rather than a noun which indicates a time.

packages/office-addin-sso/src/configure.ts
```diff
@@ -38,8 +44,22 @@ export async function configureSSOApplication(manifestPath: string, port: string
...
+        usageDataInfo = {
+            Method: ['configureSSOApplication'],
+            configDuration: [durationSsoConfiguration],
+            Platform: [process.platform],
+            Succeeded: [true]
+        }
+        usageDataHelper.sendUsageDataCustomEvent(usageDataInfo);
     }
     else {
+        usageDataHelper.sendUsageDataException('configureSSOApplication', 'Login to Azure did not succeed');
```
akrantz: I see many examples here were it sends the usage data exception and then throws it. To ensure that the message matches between the two, it would be better to first assign the message to a variable and use the same variable for both.  This applies to all of the places where you report the usage data exception.

```
const errorMessage = "Unable to log in to Azure.";
usageDataHelper.sendUsageDataException('configureSSOApplication`, errorMessage);
throw new Error(errorMessage);
```

packages/office-addin-sso/src/configure.ts
```diff
@@ -66,22 +86,22 @@ async function createNewApplication(ssoAppName: string, port: string, userJson:
...
+            usageDataHelper.sendUsageDataSuccessEvent('createNewApplication');
             return applicationJson;
         } else {
+            usageDataHelper.sendUsageDataException('createNewApplication', 'application data returned undefined');
             console.log(chalk.red("Failed to register application"));
             return undefined;
         }
-
     } catch (err) {
+        usageDataHelper.sendUsageDataException("createNewApplication", err);
```
akrantz: Is it OK to send arbitrary err value back? Does this get sanitized?

packages/office-addin-sso/src/authRoute.ts
```diff
@@ -45,11 +46,20 @@ export class AuthRouter {
...
                     const json = await tokenResponse.json();
-
                     res.send(json);
+
+                    // Send usage data
+                    const usageDataInfo: Object = {
+                        Method: ['authRouter'],
+                        scope: [scopeName],
+                        Platform: [process.platform],
+                        Succeeded: [true]
```
akrantz: Are all of these values really supposed to be in arrays?

packages/office-addin-sso/src/authRoute.ts
TCourtneyOwen: usagedataHelper is intended as a wrapper for office-addin-usage data methods. I implemented it because I felt it de-cluttered the code a bit and made things neater. Otherwise I have to fashion an object each time I was to send a new success event

packages/office-addin-sso/src/configure.ts
TCourtneyOwen: yes it does get sanitized by the office-addin-usage-data package

packages/office-addin-sso/src/authRoute.ts
TCourtneyOwen: this format is the expectation of the office-addin-usage-data package as it's currently implemented. I tried not doing it in an array initially and git an error when looping through the object in reportEventApplicationInsights

packages/office-addin-sso/src/authRoute.ts
akrantz: I think it points out that the underlying methods exposed weren't designed to be used as easily as they could. But even with the helper, I don't think you should be importing the same file across the other packages. It breaks the modular design. It would be better to just make a local helper method for now, but we should really discuss a better pattern for this in the future. Perhaps we can chat over Teams as this may not fully capture my point.

packages/office-addin-sso/src/authRoute.ts
akrantz: OK. It looks like there are some good learnings here about how the methods exposed from the office-addin-usage-data could have been better designed.

packages/office-addin-sso/src/ssoDataSettings.ts
```diff
@@ -49,11 +53,12 @@ export function getSecretFromCredentialStore(ssoAppName: string, isTest: boolean
                 const getSecretFromMacStoreCommand = `${isTest ? "" : "sudo"} security find-generic-password -a ${os.userInfo().username} -s ${ssoAppName} -w`;
                 return execSync(getSecretFromMacStoreCommand, { stdio: "pipe" }).toString();;
             default:
+                usageDataHelper.sendUsageDataException('getSecretFromCredentialStore', `unsupported platform ${process.platform}`);
                 throw new Error(`Platform not supported: ${process.platform}`);
         }
-
     } catch (err) {
-        return '';
```
TCourtneyOwen: this code shouldn't have been changed - reverting back locally and will recommit

packages/office-addin-sso/src/configure.ts
TCourtneyOwen: Good idea. Fixed throughout

packages/office-addin-sso/src/configure.ts
```diff
@@ -9,11 +9,17 @@ import * as defaults from './defaults';
...
 
 export async function configureSSOApplication(manifestPath: string, port: string) {
+    // Log start time for configuration process
+    const startSsoCongiguration = (new Date()).getTime();
+
     // Check to see if Azure CLI is installed.  If it isn't installed, then install it
     const cliInstalled = await isAzureCliInstalled();
+
     if(!cliInstalled) {
         console.log(chalk.yellow("Azure CLI is not installed.  Installing now before proceeding"));
```
akrantz: there really shouldn't be console.log() calls and chalk used from the package. I would recommend throwing an exception and having the command top-level exception handler print the message and then have the user re-run the command after azure cli is installed.

packages/office-addin-sso/src/configure.ts
```diff
@@ -66,23 +87,25 @@ async function createNewApplication(ssoAppName: string, port: string, userJson:
...
             console.log(chalk.green(`App secret is ${secret}`));
             addSecretToCredentialStore(ssoAppName, secret);
 
+            usageDataHelper.sendUsageDataSuccessEvent('createNewApplication');
             return applicationJson;
         } else {
-            console.log(chalk.red("Failed to register application"));
+            const errorMessage = 'Failed to register application';
+            usageDataHelper.sendUsageDataException('createNewApplication', errorMessage);
+            console.log(chalk.red(errorMessage));
```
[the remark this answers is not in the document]
TCourtneyOwen: @akrantz are you saying we shouldn't be writing to console log anywhere or just for error cases?

packages/office-addin-sso/src/configure.ts
akrantz: We shouldn't be using console.log anywhere in office-addin-scripts packages. The code should instead return exceptions and the consumer of the package can decide what UI to use for any errors. For command-line, the exception would bubble up to the top-level catch in the command which would print the exception info to the console.

TCourtneyOwen: @nokafor are we really fine with removing all status logging from the package?  If so, I can remove the console logging, but I feel like this gives users a sense of progress being made

packages/office-addin-sso/src/configure.ts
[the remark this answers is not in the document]
TCourtneyOwen: Moved all logging to command.ts so there is no logging in the various methods.  This will isolate logging to CLI commands

TCourtneyOwen:
Awesome, thanks!

PR #245, opened 2020-01-17
Fix dev certs for mac

TCourtneyOwen:
With these changes, on Mac we will check to see if the dev-cert exists first.  If it does, we will subsequently see if it's expired and if that's the case uninstall the certs.

I did local testing and this seems to work well, but it might be good to get some other eyes on it


packages/office-addin-dev-certs/src/uninstall.ts
```diff
@@ -32,13 +32,24 @@ export function deleteCertificateFiles(certificateDirectory: string = defaults.c
    }
 }
 
-export async function uninstallCaCertificate(machine: boolean = false, verbose: boolean = true) {
-   if (isCaCertificateInstalled()) {
+export async function uninstallCaCertificate(machine: boolean = false, verbose: boolean = true, expiredCert = false) {
```
akrantz: Let's not have this expiredCert parameter. isCaCertificateInstalled() should be true for an installed cert whether expired or not. There can be another isCaCertificateExpired() which checks expiration. The logic can be refactored to use those.

packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -25,7 +26,18 @@ export function isCaCertificateInstalled(): boolean {
     try {
         const output = execSync(command, {stdio : "pipe" }).toString();
         if (process.platform === "darwin") {
-            return true;
+            // Ceritificate exists. Now check and see if it's expired and remove it if is.
```
asachin96: We should fix this for Windows as well. Even in windows, isCaCertificateInstalled returns false if cert is expired, skipping uninstall phase.

TCourtneyOwen: I validated that this is not an issue on Windows, but yes we should remove expired certs for both platforms

TCourtneyOwen: @asachin96 I spoke with Louis and we decided to do the Windows expired cert check as a separate work item and confine this PR to just Mac so we can quickly get a fix out.  See VSO.3894960: Update office-addin-dev-certs to remove expired certs on Windows

asachin96: Alright

packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -1,72 +1,83 @@
...
     const command = getVerifyCommand();
-
+ 
     try {
         const output = execSync(command, {stdio : "pipe" }).toString();
         if (process.platform === "darwin") {
-            return true;
+            // Ceritificate exists. Now check and see if it's expired and remove it if is.
+            try {
+                const command = `security find-certificate -c '${defaults.certificateName}' -p | openssl x509 -checkend 86400 -noout`;
```
akrantz: can you move this into a function isCaCertificateExpired() and then have the check be:

// on Mac, remove certificate if expired
if (process.platform === darwin) {
  if (isCaCertificateExpired()) {
     uninstallExpiredCertificate();
   }
}

packages/office-addin-dev-certs/src/uninstall.ts
```diff
@@ -18,34 +18,49 @@ function getUninstallCommand(machine: boolean = false): string {
...
    } else {
       if (verbose) {
-         console.log(`The CA certificate is not installed.`);
+         console.log(`The CA certificate is not installed.`);
       }
    }
 }
+ 
+// Currently this method is only intended to be used for Mac due to problems on Mac that occur when CA certificates expire
+export async function uninstallExpiredCaCertificate() {
```
akrantz: don't export this

akrantz: actually I don't think you need this function at all. The uninstall command is the same and the message doesn't need to say whether it is expired. can't you just use the regular uninstallCaCertificate() function?

packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -1,72 +1,83 @@
...
+                uninstallExpiredCaCertificate()
+                return false;
+            }
+ 
         } else if (output.length !== 0) {
             return true; // powershell command return empty string if the certificate not-found/expired
         }
     } catch (error) {
-        // Mac security command throws error if the certifcate is not-found/expired
+        // Mac security command throws error if the certifcate is not found
```
akrantz: certifcate is misspelled.

TCourtneyOwen: fixed

packages/office-addin-dev-certs/src/uninstall.ts
TCourtneyOwen: I can't use the regular uninstallCaCertificate because it calls isCaCertificateInstalled, which will then get us in an endless loop.

TCourtneyOwen: That's why I initially had uninstallCaCertificate take a Boolean expiredCert param so we could bypass the call to isCaCertificateInstalled for the removing of an expired cert

akrantz: Would it help to add an optional param "expiredOnly = false" to uninstallCaCertificate() which could be called with expiredOnly as true for Mac?

packages/office-addin-dev-certs/src/verify.ts
akrantz: My biggest concern with this is that the isCaCertificateInstalled() function is adding a side-effect which removes a cert. That really shouldn't be done inside this function as a side-effect. Is there a way to move this out of this function and call it for mac from another place?

hyacinthus:
What's new about this pr?

This issue is annoyed in Mac.

millerds:
Did a different fix in more up to date repo.

PR #246, opened 2020-01-22
App insights web addition

arttarawork:
This PR adds the ability to send data up with ApplicationInsights-JS. We haven't been able to use AppInsights in the taskpane, since our current AppInsights methods only work with the node version of AppInsights; these changes add the ability to send events up in the taskpane without breaking existing code that implements office-addin-usage-data.

Additionally, I modified the function that turns on usage data to allow a percentage of events being sent up, instead of all or none.

Testing still needs to be done, but I wanted to send it out earlier to get reviews in.


packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import * as appInsights from "applicationinsights";
+import * as appInsightsWeb from '@microsoft/applicationinsights-web';
```
TCourtneyOwen: Did you test the new web methods in taskpane code (e.g. https://github.com/OfficeDev/Office-Addin-TaskPane/blob/master/src/taskpane/excel.ts).  We need to ensure that if the taskpane code calls the web methods that the project still builds.  Build fails when calling the non-web methods

TCourtneyOwen: We should add tests to test/test.ts for the new web functionality

TCourtneyOwen: I just tried importing this into Office-Addin-Taskpane and because usageData.ts also imports applicationinsights(the non-web version) build fails. I think this suggests that maybe we should have an entirely different package for web

TCourtneyOwen: I just did some local simplifying of the new web package and am able to build. Basically, in order to be able to import to the taskpane and build, all references to fs and applicationinsights need to be removed.  We should discuss because I think the new package doesn't need to support all the cli commands that office-addin-usage-data supports. I confirmed with my local changes that the taskpane project successfully builds and reports usage data when the taskpane app is used

packages/office-addin-usage-data-web/src/usageData.ts
```diff
@@ -0,0 +1,236 @@
...
+
+      if (this.options.instrumentationKey === undefined) {
+        throw new Error("Instrumentation Key not defined - cannot create usage data object");
+      }
+
+      if (this.options.groupName === undefined) {
+        throw new Error("Group Name not defined - cannot create usage data object");
+      }
+
+      if (this.options.promptQuestion === undefined) {
```
TCourtneyOwen: you can remove this part

packages/office-addin-usage-data-web/src/usageDataSettings.ts
```diff
@@ -0,0 +1,75 @@
...
+import * as fs from "fs";
+import * as defaults from "./defaults";
+import { UsageDataLevel } from "./usageData";
+
+/**
+ * Allows developer to check if the program has already prompted before
+ * @param groupName Group name of the usage data object
+ * @returns Boolean of whether the program should prompt
+ */
+export function needToPromptForUsageData(groupName: string): boolean {
```
TCourtneyOwen: not needed

packages/office-addin-usage-data-web/src/usageDataSettings.ts
```diff
@@ -0,0 +1,75 @@
...
+    return !groupNameExists(groupName);
+}
+
+/**
+ * Reads data from the usage data json config file
+ * @returns Parsed object from json file if it exists
+ */
+export function readUsageDataJsonData(): any {
```
TCourtneyOwen: seems like there might be some redundancy in all these read methods -we should figure out which ones we absolutely need and remove an redundancy

packages/office-addin-usage-data-web/test/test.ts
```diff
@@ -0,0 +1,252 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
```
TCourtneyOwen: do these tests all pass? Seems like they aren't all relevant to this package

arttarawork: They all pass on my machine, but there probably are redundancies, since it was just a copy/paste from the non-web version and a change to use the web methods. I'll see what makes sense to remove.

arttarawork:
> This loads both web and non-web versions and uses a bool on the call to distinguish.  This is inefficient.
> Consider instead having a separate package office-addin-usage-data-web which targets the web.
> Instead of a bool on the method, if only one of appInsights versus appInsightsWeb is loaded, then it can know to use the one that if available. There wouldn't need to be a parameter added to the methods.

I've managed to split it into two separate packages. I should add more testing to cover scenarios that we would find in the web version, but let me know if the package structure is to your liking @akrantz

akrantz:
As we've discussed, we're not going to do this.

PR #250, opened 2020-01-28
Usage data implementation

arttarawork:


packages/office-addin-sso/src/configure.ts
```diff
@@ -17,11 +15,9 @@ export async function createNewApplication(ssoAppName: string, port: string, use
         const rePort = new RegExp('{PORT}', 'g');
         azRestCommand = azRestCommand.replace(reName, ssoAppName).replace(rePort, port);
         const applicationJson: Object = await promiseExecuteCommand(azRestCommand, true /* returnJson */);
-        usageDataHelper.sendUsageDataSuccessEvent('createNewApplication');
         return applicationJson;
     } catch (err) {
         const errorMessage: string = `Unable to register new application: \n${err}`
-        usageDataHelper.sendUsageDataException("createNewApplication", errorMessage);
```
TCourtneyOwen: I think we may still want the exception telemetry here so we get a granular sense of where the error occurs

packages/office-addin-lint/src/lint.ts
```diff
@@ -11,8 +12,12 @@ const prettierFilePath = path.resolve(prettierPath, "../bin-prettier.js");
 const esLintConfigPath = path.resolve(__dirname, "../config/.eslintrc.json");
 
 function execCommand(command: string) {
-  const execSync = require("child_process").execSync;
-  const child = execSync(command, { stdio: "inherit" });
+  try {
+    const execSync = require("child_process").execSync;
+    const child = execSync(command, { stdio: "inherit" });
+  } catch (err) {
+    throw err;
```
akrantz: It doesn't make sense to `catch err` just to `throw err`. It is the same as no catch at all. Perhaps you removed telemetry reporting here but missed removing the catch as well.

packages/office-addin-lint/src/lint.ts
```diff
@@ -36,8 +47,14 @@ export function getLintFixCommand(files: string): string {
 }
 
 export function performLintFix(files: string) {
-  const command = getLintFixCommand(files);
-  execCommand(command);
+  try{
+    const command = getLintFixCommand(files);
+    execCommand(command);
+    usageDataHelper.sendUsageDataSuccessEvent("fix");
```
akrantz: `performLintFix` is a better name. It might be good pattern to use the function name, unless there is a good reason not to do so. I like consistency like that.

packages/office-addin-lint/src/lint.ts
```diff
@@ -26,8 +31,14 @@ export function getLintCheckCommand(files: string): string {
 }
 
 export function performLintCheck(files: string) {
-  const command = getLintCheckCommand(files);
-  execCommand(command);
+  try {
+    const command = getLintCheckCommand(files);
+    execCommand(command);
+    usageDataHelper.sendUsageDataSuccessEvent("check");
```
akrantz: `check` is very generic. `performLintCheck` -- same as the function -- seems more appropriate.

Also, put the name into a const variable which is used for both success and error reporting. That pattern ensures that the same name is used.

`const usageDataName = "performLintCheck";`

usageDataHelper.sendUsageDataSuccessEvent(usageDataName);
or
usageDataHelper.sendUsageDataException(usageDataName, err);

packages/office-addin-lint/src/lint.ts
```diff
@@ -26,8 +31,14 @@ export function getLintCheckCommand(files: string): string {
...
+  } catch (err) {
+    usageDataHelper.sendUsageDataException("check", `${err}`);
```
akrantz: I'm not so sure about the conversion of `err` to a string here. `err.message` would be the message whereas this would stringify the whole object. Also, you cannot be sure of the type for `err` which could be an Error object, a string, or more rarely something else. There is a function in the packages to do this check. I need to look it up, but using a helper function to convert `err` to a string would be recommended.

TCourtneyOwen:
@arttarawork I think we shouldn't add anymore packages to this PR, as it's getting to be quite large and difficult to review.  Let's leave adding telemetry to other packages for separate PRs. It really helps from a review standpoint to break these changes into manageable chunks

packages/office-addin-debugging/src/defaults.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as usageData from "office-addin-usage-data";
+
+// Usage data defaults
+export const usageDataProjectName: string = "office-addin-debugging";
+export const sendUsageData: boolean = usageData.groupNameExists(usageData.groupName) && usageData.readUsageDataLevel(usageData.groupName) === usageData.UsageDataLevel.on;
```
akrantz: Why does sendUsageData and usageDataOptions need to be defined here. Can't this stuff come directly from office-addin-usage-data? Are there any plans to improve/fix that API rather than put a lot of redundant stuff in each package?

packages/office-addin-debugging/src/stop.ts
```diff
@@ -5,35 +5,42 @@ import { clearDevSettings, unregisterAddIn } from "office-addin-dev-settings";
 import { readManifestFile } from "office-addin-manifest";
 import * as debugInfo from "./debugInfo";
 import { startProcess, stopProcess } from "./process";
+import * as usageDataHelper from "./usagedata-helper";
```
akrantz: we shouldn't add a file usagedata-helper in each project. This stuf can be exported from office-addin-usage-data itself. It would be better to fix the package to export things properly. That requires a major version update, e.g. a breaking change, but would be better than doing a bunch of redundant coding in each package.

packages/office-addin-debugging/src/usagedata-helper.ts
```diff
@@ -0,0 +1,21 @@
...
+        Succeeded: [true]
+    }
+    usageDataObject.reportEvent(defaults.usageDataProjectName, usageDataInfo);
+}
+
+export function sendUsageDataCustomEvent(usageDataInfo: Object) {
+    usageDataObject.reportEvent(defaults.usageDataProjectName, usageDataInfo);
+}
+
+export function sendUsageDataException(method: string, error: string) {
```
akrantz: This function should take `error: string | Error` and have logic in its implementation to convert the type to a string. 

if error is an Error object, use error.message.
if error is a string, use error.
if another type, could happen anyway, I would turn this into a string like `Unexpected type: ${JSON.stringify(error)}`

packages/office-addin-debugging/src/usagedata-helper.ts
```diff
@@ -0,0 +1,21 @@
...
+    const usageDataInfo = {
+        Method: [method],
+        Platform: [process.platform],
+        NodeVersion: [process.version],
+        Succeeded: [true]
+    }
+    usageDataObject.reportEvent(defaults.usageDataProjectName, usageDataInfo);
+}
+
+export function sendUsageDataCustomEvent(usageDataInfo: Object) {
```
akrantz: This should change to what the helper method is doing. Take some parameters along with an object that can supply the additional parameters and then have the implementation build up the object needed to pass to appinsights.

TCourtneyOwen:
I see the CI pipeline jobs are failing during the build stage. If you go to the root of the office-addin-scripts package and run "npm run build" locally, does this succeed?  It's not clear from looking at the CI build logs what the build problem is

TCourtneyOwen:
I just tried locally building the branch and here's the error that occurs when building office-addin-manifest:

[0] src/defaults.ts(4,28): error TS2307: Cannot find module 'office-addin-usage-data'.
[0] src/usagedata-helper.ts(2,28): error TS2307: Cannot find module 'office-addin-usage-data'.

packages/office-addin-debugging/src/defaults.ts
TCourtneyOwen: Agreed we shouldn't need to define another sendUsageData method in each of the defaults.  We can just create a new exported method (if one doesn't already exist) in office-addin-usage data that wraps groupNameExists and readUsageDataLevel and returns true or false

packages/office-addin-debugging/src/defaults.ts
```diff
@@ -15,4 +15,14 @@ export const usageDataOptions: usageData.IUsageDataOptions = {
     usageDataLevel: sendUsageData ? usageData.UsageDataLevel.on : usageData.UsageDataLevel.off,
     method: usageData.UsageDataReportingMethod.applicationInsights,
     isForTesting: false
-}
+};
+const usageDataObject: usageData.OfficeAddinUsageData = new usageData.OfficeAddinUsageData(usageDataOptions);
+export function sendUsageDataSuccessEvent(method: string, ...data: object[]) {
+    usageDataObject.sendUsageDataSuccessEvent(usageDataProjectName, data.concat({Method: method}));
+};
+export function sendUsageDataException(method: string, error: Error | string, ...data: Object[]) {
```
akrantz: I think using a properties object is preferrable to a variable number of arguments. The caller would create the object with properties rather than just pass additional args. This is a better javascript way to do it here. Glad to explain more in person.

You should also use "object" and not Object for types.

packages/office-addin-manifest/src/manifestInfo.ts
```diff
@@ -79,9 +80,17 @@ export async function modifyManifestFile(manifestPath: string, guid?: string, di
...
+      try {
+        manifestData = await modifyManifestXml(manifestPath, guid, displayName);
+        await writeManifestData(manifestPath, manifestData);
+        let output = await readManifestFile(manifestPath);
+        sendUsageDataSuccessEvent("modifyManifestFile");
+        return output;
+        
+      } catch(err) {
+        sendUsageDataException("modifyManifestFile", err);
+        throw err;
```
TCourtneyOwen: I think this block of code may be the source for the CI loop failure we are seeing on Mac, although I am not sure why it's failing on Mac in particular.  The problem looks to be that when we throw an error we include the full filepath in the error message, and the test is not expecting the full file path.  We should update the test to reflect this code change

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -57,20 +53,31 @@ export class OfficeAddinUsageData {
...
+  private defaultData = {
+    Platform: process.platform,
+    NodeVersion: process.version
+  };
 
   constructor(usageDataOptions: IUsageDataOptions) {
     try {
-      this.overwriteProps(this.options, usageDataOptions);
+      this.options = {
+        groupName: defaults.groupName,
```
TCourtneyOwen: does this override the group name if specified in the UsageDataOptions?  What if a different group name is specified than the defaults groupName

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -57,20 +53,31 @@ export class OfficeAddinUsageData {
...
+        method: UsageDataReportingMethod.applicationInsights,
+        isForTesting: false,
+        ...usageDataOptions
+      };
+      
+      if (this.options.instrumentationKey === undefined) {
+        throw new Error("Instrumentation Key not defined - cannot create usage data object");
+      }
+
+      if (this.options.groupName === undefined) {
```
TCourtneyOwen: effectively, this condition would never occur if we instantiate this.options.groupName to defaults.groupName above

packages/office-addin-usage-data/src/usageData.ts
arttarawork: Yup! Anything in usageDataOptions overwrites the things above, so if someone passed in a usageDataOptions object with a different groupName, theirs would overwrite our defaults.

packages/office-addin-usage-data/src/usageData.ts
arttarawork: It's a safeguard in case a developer consuming our API passes in a `usageDataOptions` object with `groupName: undefined` to the constructor

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -52,11 +54,23 @@ export class OfficeAddinUsageData {
...
-      this.options = usageDataOptions;
-
+      this.options = {
+        groupName: defaults.groupName,
+        promptQuestion: "",
+        raisePrompt: true,
+        usageDataLevel: UsageDataLevel.off,
+        method: UsageDataReportingMethod.applicationInsights,
+        isForTesting: false,
+        ...usageDataOptions
```
TCourtneyOwen: I am confused - where does this input usageDataOptions get used?

arttarawork: The properties of `usageDataOptions` gets written onto `this.options` on line 71.

TCourtneyOwen: OK I just stepped through the code in VS - this makes sense to me now

packages/office-addin-manifest/src/manifestInfo.ts
arttarawork: The error in our pipeline is because the tests expect the full filepath but instead just get `manifest.xml`. I tried running the manifest tests on a separate Mac and got all passing, for some reason

arttarawork: Screen Shot 2020-02-06 at 4 21 36 PM

PR #256, opened 2020-02-19
package & package-lock updates

arttarawork:
Updating the package.json and package-lock.json files of packages implementing usageData

### DIFF
--- a/packages/office-addin-debugging/package-lock.json
+++ b/packages/office-addin-debugging/package-lock.json
@@ -151,9 +151,9 @@
       }
     },
     "applicationinsights": {
-      "version": "1.7.0",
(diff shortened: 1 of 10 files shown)


akrantz:
You shouldn't be updating the package.json here directly. Lerna handles this. As for the package-lock.json files, it shouldn't neccessarily be required. I'd like to understand more about why you're doing this.

akrantz:
OK, after we spoke I found out that the PR title was just misleading. This is work to have other packages use the updated methods in the office-addin-usage-data package. As we also discussed, the minor version of usage-data should be bumped. Then you can make another PR to have the relevant packages use it.

akrantz:
@arttarawork One other suggestion, git branch names are easier to read if you use only lowercase and hyphens as a separator, e.g.  "add-usage-data-to-packages".

PR #262, opened 2020-03-02
office-addin-dev-settings: add webview command

arttarawork:
The webview command allows the developer to specify which web view type to use when debugging, or see the current setting. This command works on Windows only.

webview <manifest-path> [web-view-type]

If web-view-type is specified, the value will be written to the registry.
If "default" is specified, the value is cleared.
If the option is not provided, the current setting is shown.


packages/office-addin-debugging/src/commands.ts
```diff
@@ -56,6 +56,7 @@ export async function start(manifestPath: string, platform: string | undefined,
         const sourceBundleUrlComponents = new devSettings.SourceBundleUrlComponents(
             command.sourceBundleUrlHost, command.sourceBundleUrlPort,
             command.sourceBundleUrlPath, command.sourceBundleUrlExtension);
+        const webview = parseWebview(command.webview) | devSettings.WebViewType.Edge;
```
akrantz: add the type onto the variable: `const webview: WebViewType | undefined = parseWebView()`

packages/office-addin-debugging/src/commands.ts
```diff
@@ -74,7 +75,7 @@ export async function start(manifestPath: string, platform: string | undefined,
         }
 
         await startDebugging(manifestPath, appTypeToDebug, app, debuggingMethod, sourceBundleUrlComponents,
-            devServer, devServerPort, packager, packagerHost, packagerPort, enableDebugging, enableLiveReload);
+            devServer, devServerPort, packager, packagerHost, packagerPort, webview, enableDebugging, enableLiveReload);
```
akrantz: adding a param in the middle is a breaking change. Add it to the end.

packages/office-addin-debugging/src/start.ts
```diff
@@ -108,6 +108,17 @@ export function parsePlatform(text: string): Platform | undefined {
     }
 }
 
+export function parseWebview(text?: string): devSettings.WebViewType {
```
akrantz: This should come from office-addin-dev-settings package.

packages/office-addin-debugging/src/start.ts
```diff
@@ -190,6 +201,7 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
     sourceBundleUrlComponents?: devSettings.SourceBundleUrlComponents,
     devServerCommandLine?: string, devServerPort?: number,
     packagerCommandLine?: string, packagerHost?: string, packagerPort?: string,
+    webview: devSettings.WebViewType = devSettings.WebViewType.Edge,
```
akrantz: default here should be `undefined` not Edge.

packages/office-addin-debugging/src/start.ts
```diff
@@ -224,6 +236,11 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
         //     const name = isDesktopAppType ? "EdgeWebView" : "EdgeWebBrowser";
         //     await devSettings.ensureLoopbackIsEnabled(name);
         // }
+
+        // switch webview
+        if(isWindowsPlatform && webview) {
```
akrantz: nit: space needed for `if (`

akrantz: I would reverse it as `if (webview && isWindowsPlatform)`

packages/office-addin-debugging/src/commands.ts
```diff
@@ -6,7 +6,7 @@ import * as fs from "fs";
 import { logErrorMessage, parseNumber } from "office-addin-cli";
 import * as devSettings from "office-addin-dev-settings";
 import { OfficeApp, parseOfficeApp } from "office-addin-manifest";
-import { AppType, parseAppType, parseDebuggingMethod, parsePlatform, Platform, startDebugging } from "./start";
+import { AppType, parseAppType, parseDebuggingMethod, parsePlatform, Platform, startDebugging} from "./start";
```
TCourtneyOwen, line 9: nit -  reinsert space before '}'

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -165,6 +165,24 @@ export async function enableRuntimeLogging(path?: string) {
   }
 }
 
+export async function getWebView() {
+  try {
+    const webview = await devSettings.getWebView();
+    webview ? console.log("The webview has been set to " + webview + ".") :
```
TCourtneyOwen, line 171: Nit - "The webview is set ..."

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -165,6 +165,24 @@ export async function enableRuntimeLogging(path?: string) {
...
+  try {
+    const webview = await devSettings.getWebView();
+    webview ? console.log("The webview has been set to " + webview + ".") : 
+    console.log("A specific webview has not been selected. Edge is used as a default.");
+  } catch (err) {
+    logErrorMessage(err);
+  }
+}
+
+export async function setWebView(webview?: string) {
```
TCourtneyOwen, line 178: Can't we just check if webview is set in start.ts and then only call setWebView if it is, and then **not** make webview and optional param?

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -346,6 +364,28 @@ function toDebuggingMethod(text?: string): devSettings.DebuggingMethod {
   }
 }
 
+export function parseWebView(text?: string): devSettings.WebViewType | undefined {
+  const newText = text ? text.toLowerCase() : undefined;
```
TCourtneyOwen, line 368: suggestion - call this "parsedWebView" inseated of "newText"

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -173,6 +179,24 @@ export async function setSourceBundleUrl(addinId: string, components: SourceBund
   }
 }
 
+export async function setWebView(webview: WebViewType | undefined): Promise<void> {
```
TCourtneyOwen, line 182: nit - call param "webView" here and throughout

TCourtneyOwen:
Btw, the CI loop failure on the Windows 1903 agent is a known issue.  @akrantz  was working on this but not sure if he ever figured this out.  You may want to merge master is you haven;t already done so

packages/office-addin-dev-settings/src/dev-settings.ts
```diff
@@ -18,9 +18,11 @@ export enum DebuggingMethod {
 }
 
 export enum WebViewType {
+  Default,
   IE,
   Edge,
-  Clear
+  EdgeChromium,
+  Clear,
```
akrantz, line 25: You don't need "Clear". Setting the "Default" should clear the setting.

TCourtneyOwen:
@arttarawork any ideas for the CI build failures?

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -263,6 +251,34 @@ function parseStringCommandOption(optionValue: any): string | undefined {
...
+    case "clear":
+      return devSettings.WebViewType.Clear;
+    case "default":
+    case "":
+      return devSettings.WebViewType.Default;
+    case null:
+    case undefined:
+      return undefined;
+    default:
+      throw new Error(`Please select a valid webview instead of '${parsedWebView}'. Options include (ie, edge, edge chromium, clear, default)`);
```
akrantz, line 278: Don't list the valid options here. Just make the message simpler. It should be phrased like other similar error messages.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -263,6 +251,34 @@ function parseStringCommandOption(optionValue: any): string | undefined {
   return (typeof(optionValue) === "string") ? optionValue : undefined;
 }
 
+export function parseWebView(webViewText?: string): devSettings.WebViewType | undefined {
```
akrantz, line 254: `parseWebViewType` would be a better name.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -263,6 +251,34 @@ function parseStringCommandOption(optionValue: any): string | undefined {
...
+  const parsedWebView = webViewText ? webViewText.toLowerCase() : undefined;
+  switch (parsedWebView) {
```
akrantz, line 256: you don't really need the variable name here. I think it would be better to move the conditional expression directly into the switch.

`switch (webViewType ? webViewText.toLowercase() : undefined)`

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -179,27 +179,25 @@ export async function setSourceBundleUrl(addinId: string, components: SourceBund
   }
 }
 
-export async function setWebView(addinId: string, webview: WebViewType | undefined): Promise<void> {
+export async function setWebView(addinId: string, webView: WebViewType | undefined): Promise<void> {
```
akrantz, line 182: I would use `webViewType` as the variable name. It starts lowercase to distinguish from the enum WebViewType.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -179,27 +179,25 @@ export async function setSourceBundleUrl(addinId: string, components: SourceBund
...
   const key = getDeveloperSettingsRegistryKey(addinId);
-  switch (webview){
+  switch (webView){
     case undefined:
     default:
       break;
     case WebViewType.Default:
-      await registry.addStringValue(key, WebViewSelection, "default");
+    case WebViewType.Clear:
```
akrantz, line 189: You don't need Clear. Please remove that enum type. Default covers this.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -179,27 +179,25 @@ export async function setSourceBundleUrl(addinId: string, components: SourceBund
...
+      await registry.deleteValue(key, WebViewSelection);
       break;
     case WebViewType.IE:
       await registry.addStringValue(key, WebViewSelection, "ie");
```
akrantz, line 193: I would move the string values to constants for the names.

packages/office-addin-dev-settings/test/unit-tests/test.ts
```diff
@@ -115,6 +118,26 @@ describe("DevSettingsForAddIn", function() {
...
+      });
+      it("webView can be set to Edge", async function() {
+        await devSettings.setWebView(addinId, devSettings.WebViewType.Edge);
+        assert.strictEqual(await devSettingsWindows.getWebView(addinId), "edge");
+      });
+      it("webView can be set to Edge Chromium", async function() {
+        await devSettings.setWebView(addinId, devSettings.WebViewType.EdgeChromium);
+        assert.strictEqual(await devSettingsWindows.getWebView(addinId), "edge chromium");
+      });
+      it("webView can be cleared", async function() {
```
akrantz, line 137: You don't need Clear. Setting default is clearing the value. Have I mentioned this enough times?  :-)

TCourtneyOwen:
Kicking off manual steps to install all packages on VM where CI loop fails and retriggering failed run

TCourtneyOwen:
CI job passed

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -75,6 +75,18 @@ commander
...
+commander
+    .command("webview <manifest-path> [web-view-type]")
+    .description("Allows specifying a preference for the type of web view to use when debugging. Windows only")
+    .action(commands.webView)
+    .on("--help", () => {
+        console.log("\nFor [web-view-type], choose one of the following values:\n");
+        console.log("\t'edge' for Microsoft Edge");
+        console.log("\t'ie' for Internet Explorer 11");
+        console.log("\t'default' to remove any preference");
+        console.log("...or leave [web-view-type] blank to return the current setting");
```
akrantz, line 87: Suggest changing the last line to: `\nOmit [web-vew-type] to see the current setting.`

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -237,6 +237,31 @@ function parseStringCommandOption(optionValue: any): string | undefined {
...
+    case "anaheim":
+      return devSettings.WebViewType.EdgeChromium;
+    case "default":
+    case "":
+      return devSettings.WebViewType.Default;
+    case null:
+    case undefined:
+      return undefined;
+    default:
+      throw new Error(`Please select a valid webViewType instead of '${webViewString!.toLowerCase()}'.`);
```
akrantz, line 261: You don't need toLowercase() here as you are just showing the value they put.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -363,3 +388,22 @@ function validateManifestId(manifest: ManifestInfo) {
     throw new Error(`The manifest file doesn't contain the id of the Office Add-in.`);
   }
 }
+
+export async function webView(manifestPath: string, webViewString?: string) {
```
akrantz, line 392: While the command name is `webView`, we could choose a more descriptive name for the exported function. Also, we shouldn't have the console.log() statements be in the exported function. I think what would be better to is to have separate functions setWebViewType() and getWebViewType() and take the code which checks if webViewString is truthy or not and does set or get and move that to the command implementation.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -16,6 +16,11 @@ const SourceBundlePort: string = "SourceBundlePort";
 const UseDirectDebugger: string = "UseDirectDebugger";
 const UseLiveReload: string = "UseLiveReload";
 const UseProxyDebugger: string = "UseWebDebugger";
+const WebViewSelection: string = "WebViewSelection";
+const IE: string = "ie";
```
akrantz, line 20: It would help to use a common prefix for these strings, e.g. `const WebViewTypeString_IE = "ie";  const WebViewTypeString_Edge = "edge"; etc.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -179,6 +179,21 @@ export async function getSourceBundleUrl(manifestPath: string) {
...
 }
 
+async function getWebViewType(manifestPath: string) {
+  try {
+    const manifest = await readManifestFile(manifestPath);
+
+    validateManifestId(manifest);
+
+    const currentWebViewString = await devSettings.getWebView(manifest.id!);
+    currentWebViewString ? console.log("The webViewType is set to " + currentWebViewString + ".") :
```
akrantz, line 189: You could simplify the name to just `webViewString`. 

Why not have return the WebViewType enum. You could write a function in devSettings toWebViewTypeString(webViewType: WebViewType): string   which uses the switch to return the string constant for the enum value. For Default, you can return an empty string.

Then here in the command you could use the function to convert to the display text and show it in the message.

Use `${variable} in a string` form rather than using + to concatenate the string.

Make the string contants mixed case, e.g. "Edge" so they display nicer. The case shouldn't matter for the command input.

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -78,13 +78,13 @@ commander
 commander
     .command("webview <manifest-path> [web-view-type]")
     .description("Allows specifying a preference for the type of web view to use when debugging. Windows only")
-    .action(commands.webView)
+    .action(commands.webViewOverride)
```
akrantz, line 81: Maybe I gave poor guidance earlier but I would be OK with the function webView() being for the "webview" command. It was the exported functions that I wanted to have a better name. Hope I didn't cause confusion.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -389,21 +416,6 @@ function validateManifestId(manifest: ManifestInfo) {
...
-      const currentWebViewString = await devSettings.getWebView(manifest.id!);
-      currentWebViewString ? console.log("The webViewType is set to " + currentWebViewString + ".") : 
-      console.log("A specific webViewType override has not been selected.");
-    }
-
-  } catch (err) {
-    logErrorMessage(err);
-  }
+export async function webViewOverride(manifestPath: string, webViewString?: string) {
+  webViewString ? setWebViewType(manifestPath, webViewString) : getWebViewType(manifestPath);
```
akrantz, line 420: This looks good. You could name the parameter `webViewTypeString` for consistency.  As I mentioned before, this function could just be named webView().

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -17,9 +17,9 @@ const UseDirectDebugger: string = "UseDirectDebugger";
 const UseLiveReload: string = "UseLiveReload";
 const UseProxyDebugger: string = "UseWebDebugger";
 const WebViewSelection: string = "WebViewSelection";
-const IE: string = "ie";
-const Edge: string = "edge";
-const EdgeChromium: string = "edge chromium";
+const WebViewTypeString_IE: string = "ie";
```
akrantz, line 20: This is where it would be good to make the display strings mixed case, e.g. "Edge", "IE"

packages/office-addin-dev-settings/src/commands.ts
arttarawork: When running getWebView(), we read what's in the registry WebView selection for the addin and return that string back. There isn't an enum that gets passed back in the first place, so I don't see the benefit of adding a string-to-enum conversion when we first get this result, only to later have an enum-to-string conversion in the command. 

That being said, the `${variable}` is a good point, and making the strings mixed case should be done. I've modified the constants we're writing to the registry to be mixed case, so it gets returned as mixed case to the user.

packages/office-addin-dev-settings/src/cli.ts
arttarawork: Funny enough, I actually now prefer webViewOverride to my previous name choice! Thank you though!

packages/office-addin-dev-settings/src/commands.ts
akrantz: The point of converting the string to the enum value is so that the function works on enum values. The conversion from the enum value to a string is so the command can show the string. 

Having the exported function return the enum makes for a better "api" for using the package from code.

packages/office-addin-dev-settings/src/commands.ts
arttarawork: I've specifically set webViewType to be for the enums and webViewString to be for strings. By that logic, though, I could rename webViewType to webViewEnum 🤔

akrantz: Except that you typically shouldn't use "Enum" as part of the name for the enum. You would use what the enum describes which in this case is types of webviews.

packages/office-addin-dev-settings/src/commands.ts
arttarawork: Would the recent way I've modified getWebViewType() be an appropriate way to make sure an enum is returned with the exported webViewOverride()?

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -179,6 +179,23 @@ export async function getSourceBundleUrl(manifestPath: string) {
   }
 }
 
+async function getWebViewType(manifestPath: string) {
+  try {
+    const manifest = await readManifestFile(manifestPath);
+
+    validateManifestId(manifest);
+
+    const webViewString = await devSettings.getWebView(manifest.id!);
```
akrantz, line 188:
```
const webViewType: WebViewType = await devSettings.getWebView(manifest.id!);
const webViewTypeString = toWebViewTypeString(webViewType);

console.log(webViewTypeString 
    ? `The web view type is set to ${webViewTypeString}.` 
     : `The web view type has not been set.`);
```

akrantz: Yes, this means that the commands converts the string to an enum and back to the string, but that's OK.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -17,9 +17,6 @@ const UseDirectDebugger: string = "UseDirectDebugger";
 const UseLiveReload: string = "UseLiveReload";
 const UseProxyDebugger: string = "UseWebDebugger";
 const WebViewSelection: string = "WebViewSelection";
-const WebViewTypeString_IE: string = "IE";
-const WebViewTypeString_Edge: string = "Edge";
-const WebViewTypeString_EdgeChromium: string = "Edge Chromium";
```
arttarawork: @akrantz I removed these consts because I have the same strings baked into the enums themselves now.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -173,6 +181,37 @@ export async function setSourceBundleUrl(addinId: string, components: SourceBund
...
+  const key = getDeveloperSettingsRegistryKey(addinId);
+  switch (webViewType){
+    case undefined:
+    case WebViewType.Default:
+      await registry.deleteValue(key, WebViewSelection);
+      break;
+    case WebViewType.IE:
+    case WebViewType.Edge:
+    case WebViewType.EdgeChromium:
+      const webViewString: string = <string> webViewType;
```
akrantz, line 194: I still think it would be better to define string constants for the registry values used and to use them here and in the toWebViewType function.

PR #266, opened 2020-03-19
Addresses most Component Governance issues

TCourtneyOwen:
- But not all due to the fact that there are vulnerabilities in lerna, webpack and cpx, for which we have the latest versions already

### DIFF
--- a/packages/office-addin-cli/package.json
+++ b/packages/office-addin-cli/package.json
@@ -33,7 +33,7 @@
     "rimraf": "^2.7.1",
     "sinon": "^7.5.0",
     "ts-node": "^8.5.2",
-    "tslint": "^5.20.1",
+    "tslint": "^6.1.0",
     "typescript": "^3.7.2"
   },
   "repository": {
--- a/packages/office-addin-debugging/package.json
+++ b/packages/office-addin-debugging/package.json
@@ -40,10 +40,10 @@
     "@types/ws": "^6.0.4",
     "concurrently": "^3.5.0",
     "express": "^4.16.4",
-    "mocha": "^5.2.0",
+    "mocha": "^7.1.1",
     "rimraf": "^2.7.1",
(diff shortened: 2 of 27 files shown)


TCourtneyOwen:
need to look into why the build fails in the CI loop but not locally

packages/office-addin-cli/package.json
```diff
@@ -33,7 +33,7 @@
     "rimraf": "^2.7.1",
     "sinon": "^7.5.0",
     "ts-node": "^8.5.2",
-    "tslint": "^5.20.1",
+    "tslint": "^6.1.0",
     "typescript": "^3.7.2"
```
akrantz: As an aside, you may want to see about a separate PR to upgrade the typescript version.

packages/office-addin-debugging/package.json
```diff
@@ -40,10 +40,10 @@
     "@types/ws": "^6.0.4",
     "concurrently": "^3.5.0",
     "express": "^4.16.4",
-    "mocha": "^5.2.0",
+    "mocha": "^7.1.1",
     "rimraf": "^2.7.1",
     "ts-node": "^8.5.2",
-    "tslint": "^5.20.1",
+    "tslint": "^6.1.0",
```
akrantz: Another aside, having a work item to move from tslint to eslint. If we can use office-addins-lint for these other packages, which itself is updated to use eslint, that might be useful. Not sure whether this will be difficult. I would hope it would be easy.

packages/office-addin-cli/package.json
TCourtneyOwen: I will do this in a separate PR

packages/office-addin-debugging/package.json
TCourtneyOwen: For now, that's going to have to go on the back-burner.

PR #267, opened 2020-03-31
Update packages

akrantz:


packages/office-addin-dev-settings/src/prompt.ts
```diff
@@ -6,7 +6,7 @@ import { getOfficeAppName, OfficeApp } from "office-addin-manifest";
 
 export async function chooseOfficeApp(apps: OfficeApp[]): Promise<OfficeApp> {
   const questionName = "app";
-  const question: inquirer.Question = {
+  const question: inquirer.ListQuestionOptions<inquirer.Answers> = {
```
akrantz: These inquirer changes were required to get it to build, but I need to manually verify that the prompts still work OK.

TCourtneyOwen:
Any idea whats causing this build failure on Windows:

packages/office-addin-test-server/test/test.ts
```diff
@@ -14,6 +14,13 @@ const testValue: string = "Office-Addin-Test-Infrastructure";
 const testValues: any = [];
```
TCourtneyOwen: Are these the only tests that need to be disabled for Linux?  I saw these tests fail in the CI loop but assumed there would be more

akrantz: I've now updated so that the tests that are not supported on Linux are skipped. There is a build error for Windows that I still need to resolve.

akrantz:
@TCourtneyOwen There is an issue with the Windows 1903 build agent. This is now passing on Mac and Linux and passes for me locally on Windows. There seems to be an issue with the node_modules where it is missing things that should be there. It would help to do "npm ci" for each of the packages to clean the node_modules and then "npm install" at the root to ensure they are initiailized properly.

packages/office-addin-dev-settings/src/prompt.ts
akrantz: I validated the prompt using the following command for office-addin-dev-settings:
 `npm run cli -- sideload <manifest>`

Since `--app` is not specified, it prompts for a choice of application.

TCourtneyOwen:
I am at a loss for why this is happening. I did a run of your changes on two different VMs, both which we redeployed last night, and the build keeps failing on the same error

akrantz:
@TCourtneyOwen I got the builds to pass. Could you please approve?

packages/custom-functions-metadata-plugin/package.json
```diff
@@ -24,17 +24,17 @@
     "test": "echo \"No tests.\""
   },
   "dependencies": {
-    "custom-functions-metadata": "^1.0.25",
+    "custom-functions-metadata": "^1.0.24",
```
akrantz, line 27: I think this should stay as 1.0.25

PR #269, opened 2020-04-02
from a clean repo, update package-lock.json files

akrantz:


akrantz:
@TCourtneyOwen I cloned into a new folder and running npm install in each of the package dirs there generated a different package-lock.json file This passed when it ran.

PR #271, opened 2020-04-03
Update the package acceptance service used in office-addin-manifest

aszczepanski:


aszczepanski:
> There are updates needed to the tests. Please run them using "npm run test". Currently there are test compile errors because of renamed properties.
> 
> When I try this out using:
> `npm run cli validate test\manifests\ExcelCustomFunctions.manifest.xml`
> it reports the manifest an not valid. This should be a valid manifest.

I fixed the errors in unit tests, they’re all passing now when I run `npm run test`.

I ran the tool using the version from master to validate that manifest (_test\manifests\ExcelCustomFunctions.manifest.xml_) and it also failed with the same error: _The element 'Host' (…) has invalid child element 'AllFormFactors'_.

I agree that we should fix the test manifest file to adhere to the schema, but not necessarily in this PR.

akrantz:
I believe the failing manifest is a copy from https://github.com/OfficeDev/Excel-Custom-Functions, which is the Yo Office template for Excel Custom Functions. Therefore I would like to understand _why_ this is failing since it should be succeeding with the original manifest validation service.

PR #273, opened 2020-04-06
remove package-lock.json files from lerna packages

akrantz:


akrantz:
@arttarawork @TCourtneyOwen I've been making pull requests to experiment with the fact that lerna bootstrap on Windows when it is a clean git repo will fail to build because the local packages are not properly linked together.

PR #277, opened 2020-04-24
Enable code to ensure Edge loopback is enabled

TCourtneyOwen:
- Only executes if OS is Windows 10

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -4,6 +4,7 @@
 import * as fetch from "node-fetch";
 import * as devCerts from "office-addin-dev-certs";
 import * as devSettings from "office-addin-dev-settings";
+import * as os from "os";
 import { DebuggingMethod, sideloadAddIn } from "office-addin-dev-settings";
 import { OfficeApp, readManifestFile } from "office-addin-manifest";
 import * as nodeDebugger from "office-addin-node-debugger";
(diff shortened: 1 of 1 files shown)


packages/office-addin-debugging/src/start.ts
```diff
@@ -218,12 +219,11 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
             throw new Error("Manifest does not contain the id for the Office Add-in.");
         }
     
-        // DISABLE FOR NOW
         // enable loopback for Edge
-        // if (isWindowsPlatform) {
-        //     const name = isDesktopAppType ? "EdgeWebView" : "EdgeWebBrowser";
-        //     await devSettings.ensureLoopbackIsEnabled(name);
-        // }
+        if (isWindowsPlatform && parseInt(os.release()) === 10) {
```
akrantz, line 223: It would be a bit more readable if add `const isWindows10: boolean = isWindowsPlatform && parseInt(...);` on line 199 just after the declaration of isWindowsPlatform, so that here you can just use `if (isWindows10)`

akrantz: @TCourtneyOwen did you make a change here?

akrantz:
@TCourtneyOwen Please don't merge if the CI is failing. The failures on Windows can be mitigated manually so that we can get the CI passing.

TCourtneyOwen:
@akrantz I propose we remove the 1903 stage entirely if manual steps are required as this effectively undermines having this be part of an automated CI loop.  Thoughts?

PR #278, opened 2020-04-27
Create install script that install dependencies for each package

TCourtneyOwen:
There is an apparent bug in lerna that causes the sub-packages to not have their dependencies fully installed when running "npm install" at root of Office-Addin-Scripts and is causing our CI loop jobs to fail. The proposed fix is to create a new npm intall script that will independently do an npm install for each sub-package

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "scripts": {
     "build": "lerna run build",
     "dist-tags": "lerna exec --concurrency 1 --no-sort --stream npm dist-tag ls",
+    "install": "lerna exec --concurrency 1 --stream -- npm install --no-fund --no-optional",
     "postinstall": "lerna bootstrap",
     "publish:next": "lerna publish --dist-tag next",
     "test": "lerna run test",


package.json
```diff
@@ -4,6 +4,7 @@
   "scripts": {
     "build": "lerna run build",
     "dist-tags": "lerna exec --concurrency 1 --no-sort --stream npm dist-tag ls",
+    "install": "lerna exec --concurrency 1 --stream -- npm install --no-fund --no-optional",
```
akrantz, line 7: I thought about adding this here as well. `install-packages` to avoid overwriting `npm install` behavior generally. 

We don't want to run this all the time. Just the one time as needed. So `npm run install-packages` when neccessary, and then `npm install` after which will do the post-install `lerna bootstrap`.

TCourtneyOwen: Yeah that's what I originally had and then I read your IM message again and took you literally:

[10:12 AM] Adam Krantz
    lerna exec --concurrency 1 --stream -- npm install --no-fund --no-optional
npm install

TCourtneyOwen: I will revert to my original change and push a new commit. I don't see how we can conditionally run npm install packages however

akrantz:
@TCourtneyOwen I recently added "install-packages" script which does this.

PR #281, opened 2020-05-11
Address recent drop in usage data because usagelevel set to off

TCourtneyOwen:
- The fix is to ensure office-addin-scripts packages don't write out the office-addin-usage-data file before Yo Office does
- Created an array of office-addin-scripts packages currently consuming office-addin-usage-data package and check if the project name is contained in that array during object construction.  If it is, then don't write out the office-addin-usage-data file.
- Also fixed some formatting issues

Yo office imports other packages such as office-addin-manifest, which when loaded as part of an import, invoke the following to create a new usageData object:
export const usageDataObject: usageData.OfficeAddinUsageData = new usageData.OfficeAddinUsageData({
    projectName: "office-addin-manifest",
    instrumentationKey:  usageData.instrumentationKeyForOfficeAddinCLITools,
    raisePrompt: false
});

In the constructor for the office-addin-usage-data, we have the following bit of code:

      if (!this.options.isForTesting && this.options.raisePrompt && jsonData.needToPromptForUsageData(this.options.groupName)) {
        this.usageDataOptIn();
      } else {
        jsonData.writeUsageDataJsonData(this.options.groupName, this.options.usageDataLevel);
      }
Because raisePrompt is set to false, we end up calling jsonData.writeUsageDataJsonData with the usageDataLevel turned to off, and write out the office-addin-usage-data file with the setting. Once this file is written, yo office will see it's written and use these settings moving forward.

### DIFF
--- a/packages/office-addin-usage-data/src/defaults.ts
+++ b/packages/office-addin-usage-data/src/defaults.ts
@@ -7,3 +7,12 @@ import * as path from "path";
 export const usageDataJsonFilePath: string = path.join(os.homedir(), "/office-addin-usage-data.json");
 export const groupName = "office-addin-usage-data";
 export const instrumentationKeyForOfficeAddinCLITools = "de0d9e7c-1f46-4552-bc21-4e43e489a015";
+// Office-Addin-Scripts packages currently sending usage data
+export const officeAddinScriptsPackages: string[] = [
+    "office-addin-debugging",
+    "office-addin-dev-certs",
+    "office-addin-lint",
+    "office-addin-manifest",
+    "office-addin-sso",
+    "office-addin-usage-data"
+]
--- a/packages/office-addin-usage-data/src/defaults.ts
+++ b/packages/office-addin-usage-data/src/defaults.ts
@@ -5,14 +5,6 @@ import * as os from "os";
 import * as path from "path";
 
 export const usageDataJsonFilePath: string = path.join(os.homedir(), "/office-addin-usage-data.json");
-export const groupName = "office-addin-usage-data";
-export const instrumentationKeyForOfficeAddinCLITools = "de0d9e7c-1f46-4552-bc21-4e43e489a015";
-// Office-Addin-Scripts packages currently sending usage data
-export const officeAddinScriptsPackages: string[] = [
-    "office-addin-debugging",
-    "office-addin-dev-certs",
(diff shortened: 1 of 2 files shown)


akrantz:
@TCourtneyOwen Wouldn't it be better and simpler to have the default be undefined and to not write anything when undefined so that only the true or false value about whether to prompt would cause the value to be written?

packages/office-addin-usage-data/src/defaults.ts
```diff
@@ -6,4 +6,13 @@ import * as path from "path";
 
 export const usageDataJsonFilePath: string = path.join(os.homedir(), "/office-addin-usage-data.json");
 export const groupName = "office-addin-usage-data";
-export const instrumentationKeyForOfficeAddinCLITools = "de0d9e7c-1f46-4552-bc21-4e43e489a015";
\ No newline at end of file
+export const instrumentationKeyForOfficeAddinCLITools = "de0d9e7c-1f46-4552-bc21-4e43e489a015";
+// Office-Addin-Scripts packages currently sending usage data
+export const officeAddinScriptsPackages: string[] = [
```
akrantz, line 11: I'm concerned about this pattern where these package names are specified. Isn't there a better way?

TCourtneyOwen:
@akrantz can you please clarify - the default of which variable?

packages/office-addin-usage-data/src/defaults.ts
TCourtneyOwen: Yeah I am not crazy about this design either. The other option is to update all the consuming office-addin-script packages to not export an usage data object and to just create it at runtime. But this opens up the clunky pattern of creating the usage data object multiple times in each file that sends usage data.  I am really striving for as minimal changes as possible, but if you thinking creating the object at runtime is the best approach, we can do that

akrantz: I think a chat might help us discuss and come up with a better proposal to deal with this.

TCourtneyOwen:
Looks like VM deployment failed the past couple days so the CI loop is not running.  Investigating now

TCourtneyOwen:
It appears they renamed the base images in Azure that we use as the foundation for creating our VMs.  I need to update the deployment scripts to use the new name.  I wish they wouldn't change these names willy-nilly

PR #283, opened 2020-05-14
update lerna to 3.21.0

akrantz:


TCourtneyOwen:
Does this by chance fix the issue we have been seeing related to package install on our windows VMs?

akrantz:
I'm hoping that it will but I don't know yet.

PR #287, opened 2020-06-04
optimist package generates high vulnerabilities - removing

TCourtneyOwen:
- This package isn't even being used so there is no need to have it in the project


TCourtneyOwen:
Looks like the VMs failed to deploy last night for the 1903 agent pool after trying to update them to use Windows 2004 base template; hence the failure.  I am investigating to see what happened with deployment

TCourtneyOwen:
Figured out the VM deployment issue.  Regenerating VMs and will run CI job again when that's complete

Issue #290, opened 2020-06-22
Application Insights errors


tgoyer:
# Prerequisites

Node: v12.16.3

# Current behavior
When developing an Excel add-in, office-addin-debugging continuously throws the following exceptions:

```
ApplicationInsights:Sender [
  'Ingestion endpoint could not be reached 20 consecutive times. There may be resulting telemetry loss. Most recent error:',
  Error: socket hang up
      at connResetException (internal/errors.js:608:14)
      at Socket.socketOnEnd (_http_client.js:453:23)
      at Socket.emit (events.js:322:22)
      at endReadableNT (_stream_readable.js:1187:12)
      at processTicksAndRejections (internal/process/task_queues.js:84:21) {
    code: 'ECONNRESET'
  }
]

ApplicationInsights:Sender [
  'Not saving data due to max size limit being met. Directory size in bytes is: 50000175'
]
```

Is there any way to either disable Application Insights or squelch this error?  In addition, what directory is the error message referring to?  I have not been able to find any documentation on these issues.

ChrisRomp:
Supposedly running `npx office-addin-usage-data off` suppresses telemetry but I'm still getting this error even after doing that and restarting my shell.

TCourtneyOwen:
Sorry to hear you are having problems.  Which version of the office-addin-debugging package are you running in your project?  If you want to try suppressing the AppInsights calls you can go to node_modules\office-addin-debugging\lib and search for instances of "**sendUsageDataSuccessEvent**" - there's one in start.js and one in stop.js - and comment them out.  Obviously that's not an ideal solution but this may help you get up and running and help us isolate if it's the calls to sendUsageDataSuccessEvent that are problematic.

Please let me know if this helps.

Thanks,

Courtney

tgoyer:
@TCourtneyOwen I'm running the version that comes by default in the `yo generator-office` command.  From the package.json:  `"office-addin-debugging": "^3.0.25"`

I commented these lines out and I still get the same error.

#### start.js
```
            ...
            console.log(enableDebugging
                ? "Debugging started."
                : "Started.");
-->          //defaults_1.usageDataObject.sendUsageDataSuccessEvent("startDebugging");
        }
        catch (err) {
-->         //defaults_1.usageDataObject.sendUsageDataException("startDebugging", err);
            throw err;
        }
```

#### stop.js
```
            ...
            console.log("Debugging has been stopped.");
[...]

PR #304, opened 2020-07-21
Fixes issue #290 - Application Insights errors

TCourtneyOwen:
- Customers with port 443 turned off for outgoing traffic, are hitting an AppInsights.Sender error even when usage data is set to "off".  The reason for this is that we still call the setup method method for appInsights, which makes and https call, even if office-addin-usage-data is set to off.
- The fix is to only call the setup method when office-addin-usage-data usageDataLevel is set to true
- I verifed locally that with port 443 blocked for outgoing traffic and usage data set to off, we no longer hit the AppInsignts.Sender error

### DIFF
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -94,11 +94,13 @@ export class OfficeAddinUsageData {
         this.usageDataOptIn();
       }
 
-      appInsights.setup(this.options.instrumentationKey)
-        .setAutoCollectExceptions(false)
-        .start();
-      this.usageDataClient = appInsights.defaultClient;
-      this.removeApplicationInsightsSensitiveInformation();
+      if (this.options.usageDataLevel === "on") {
+        appInsights.setup(this.options.instrumentationKey)
+          .setAutoCollectExceptions(false)
+          .start();
+        this.usageDataClient = appInsights.defaultClient;
+        this.removeApplicationInsightsSensitiveInformation();
+      }
     } catch (err) {
       throw new Error(err);
     }
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -94,7 +94,7 @@ export class OfficeAddinUsageData {
         this.usageDataOptIn();
       }
 
-      if (this.options.usageDataLevel === "on") {
+      if (this.options.usageDataLevel === UsageDataLevel.on) {
         appInsights.setup(this.options.instrumentationKey)
           .setAutoCollectExceptions(false)
           .start();


packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -94,11 +94,13 @@ export class OfficeAddinUsageData {
...
-      appInsights.setup(this.options.instrumentationKey)
-        .setAutoCollectExceptions(false)
-        .start();
-      this.usageDataClient = appInsights.defaultClient;
-      this.removeApplicationInsightsSensitiveInformation();
+      if (this.options.usageDataLevel === "on") {
+        appInsights.setup(this.options.instrumentationKey)
+          .setAutoCollectExceptions(false)
+          .start();
+        this.usageDataClient = appInsights.defaultClient;
```
akrantz, line 101: Please validate that when this is not turned on and this.usageDataClient is undefined, this is handled OK and doesn't lead to an exception.

Can we add test cases which provide validation for when the option is both on and off?

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -94,11 +94,13 @@ export class OfficeAddinUsageData {
         this.usageDataOptIn();
       }
 
-      appInsights.setup(this.options.instrumentationKey)
-        .setAutoCollectExceptions(false)
-        .start();
-      this.usageDataClient = appInsights.defaultClient;
-      this.removeApplicationInsightsSensitiveInformation();
+      if (this.options.usageDataLevel === "on") {
```
akrantz, line 97: It would be better if there is an enum for the usageDataLevel values. Is this not a boolean because there is more than just "on" and "off"? 

Please ensure the value is normalized to be lowercase or convert it to lowercase for this comparison.

TCourtneyOwen: Yeah we actually can use the UsageDataLevel enum. Will fix this

packages/office-addin-usage-data/src/usageData.ts
TCourtneyOwen: We should add tests for when this is turned off. I think most if not all tests only have usageDataLevel set to on

TCourtneyOwen: We have a couple tests that set usageDataLevel to off and then call reportEvent and reportError and ensure that the event and exception static count hasn't changed. These tests pass with with this change:

  describe("Test getEventsSent method", () => {
    it("should return amount of events successfully sent", () => {
      addInUsageData.setUsageDataOff();
      const testEvent = {
        Test1: [true, 100],
        ScriptType: ["Java", 1],
      };
      addInUsageData.reportEvent("office-addin-usage-data", testEvent);
      assert.equal(addInUsageData.getEventsSent(), 1);
    });
  });
  describe("Test getExceptionsSent method", () => {
    it("should return amount of exceptions successfully sent ", () => {
      addInUsageData.setUsageDataOff();
      addInUsageData.reportError("TestData", err);
      assert.equal(addInUsageData.getExceptionsSent(), 1);
    });
  });

We have tests for when usageDataLevel is set to on:

  describe("Test reportEvent method", () => {
    it("should track event of object passed in with a project name", () => {
      const testEvent = {
        Test1: [true, 100],
        ScriptType: ["JavaScript", 1],
      };
      addInUsageData.reportEvent("office-addin-usage-data", testEvent);
      assert.equal(addInUsageData.getEventsSent(), 1);
    });
  });
  describe("Test reportError method", () => {
    it("should send usage data exception", () => {
      addInUsageData.reportError("ReportErrorCheck", err);
      assert.equal(addInUsageData.getExceptionsSent(), 1);
    });
  });

akrantz: It might be nice to have more validation of things not happening when it is off. The tests seems sufficient in terms of making this fix.

PR #308, opened 2020-08-06
Add new AppId to sso app configuration script

TCourtneyOwen:
- This adds a new pre-authorization is used for OMEX add-ins in SPO.  The AppId is 08e18876-6177-487e-b8b5-cf950c1e598c


TCourtneyOwen:
I tested locally that the new appId is in fact successfully created when during the sso configuration process.  However, it's not really clear to me how to test out the scenario that uses this appId

PR #311, opened 2020-09-02
Add support to parameter addresses feature

jiju-MS:
Support parameter address feature. Please contact jiju for more details.


akrantz:
Before making these changes, the specification should be updated to make it clear about this feature. There is no description here about this feature and what capability it is intended to provide. Let's meet to discuss in more detail.

akrantz:
The specification is here: https://github.com/OfficeDev/Office-Addin-Scripts/blob/master/packages/custom-functions-metadata/Specification.md

When `@requiresParameterAddress` is specified, where is the address for the parameter passed into the function?
Can you use this with multiple parameters?
Can this co-exist with `@requiresAddress` (which provides the address where the function is being invoked in CustomFunctions.Invocation.address)?

Also, the CustomFunctionsRuntime.d.ts file would need to be updated to reflect the IntelliSense that the user would see when editing. I want to understand how they would write code that uses the parameter address, so the IntelliSense shows the appropriate information.

akrantz:
I see that there are changes for the CustomFunctionsRuntime.d.ts file here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/47122

This provides me more context that this is similar to the @requiresAddress tag in that @requiresParameterAddresses allows the invocation context to have another property with the parameter addresses. 

I now see that these code changes are OK.

Please add test cases for the following scenarios:
1. Having both `@requiresAddress` and `@requiresParameterAddress`.
2. A function with more than one parameter which uses @requiresParameterAddresses.

Please also update the specification to add documentation for @requiresParametersAddresses.

akrantz (dismissed):
OK, this looks good.  Please resolve the merge conflict.

jiju-MS:
Sorry for missing the necessary for resolving the conflict. Just commit the resolved PR.

akrantz:
Thanks for doing the merge. The CI pipeline doesn't run on the weekends so we'll work to get this completed on Monday.

akrantz:
I pushed a commit which updates the package-lock.json files.

akrantz (approved):

akrantz:
@TCourtneyOwen The CI pipeline needs to be run again for this PR...

TCourtneyOwen:
Looking into CI pipeline now

TCourtneyOwen:
CI job passed. We have an issue with the CI VM deployments where Edge webview isn't getting enabled and as a result the tests hang on the prompt asking the user if they want to enable Edge webview.  The deployment process is simply running "npx office-addin-dev-settings appcontainer EdgeWebView --loopback --yes" so it's not clear to me why this is not getting set

akrantz:
I just merged the master branch in. That will make it run the CI loop again.

akrantz:
@TCourtneyOwen There's something broken with the pipeline, as it is not running for this PR.

akrantz:
@TCourtneyOwen The command that I added to run the pipeline seems to have kicked it off, even though it said right after than no pipelines are associated.

akrantz:
@TCourtneyOwen The CI pipeline was run and has passed.

PR #321, opened 2020-09-12
add option to open dev tools when debugging

akrantz:
This is being added to support the scenario where a developer wants to use the browser dev tools for debugging, but the add-in is running in the background, so there are no controls available to open the dev tools from the webview.

The office-addin-dev-settings "debugging" command has a new option "--open-dev-tools" which will write a registry value ("OpenDevTools" REG_DWORD 1) when present.

This is implemented by adding an optional parameter to the enableDebugging() function, openDevTools: boolean = false, which will write the registry value when true, or remove it (if present) when false.

The office-addin-debugging "start" command has a new option "--dev-tools" which will specify true for the openDevTools parameter when calling devSettings.enableDebugging().

This will only have an effect when supported.

As Office adds support for the Edge Chromium webview (WebView2) on Windows, this will allow the dev tools to be opened (if desired) when using that webview.


akrantz:
I should update tests to provide coverage of these changes.
I'm not sure if the package-lock.json changes should be included. I just noticed that was added.

TCourtneyOwen (commented):
Looks good - just need to update failing tests to accommodate this change

akrantz:
I updated most of the dependencies in office-addin-dev-settings package to the latest versions. There were a few minor changes required because of breaking changes in dependencies.

akrantz:
I need to run the tests on Mac. Looks like there is a failure there.

packages/office-addin-dev-settings/src/registry.ts
```diff
@@ -121,11 +121,13 @@ export async function deleteValue(key: RegistryKey, value: string): Promise<void
     };
 
     try {
-      key.winreg.valueExists(value, (valueExistsError, exists) => {
+      key.winreg.valueExists(value, (_, exists) => {
         if (exists) {
           key.winreg.remove(value, onError);
```
TCourtneyOwen, line 126: Don't we still need to resolve after calling key.winreg?

akrantz: It actually does. winreg.remove() will call onError() in all cases (even if successful), which then calls reject() for an error or resolve() for success.  I suppose there might be a better name than onError() to convey this. Any suggestions?

packages/office-addin-dev-settings/src/registry.ts
```diff
@@ -121,11 +121,13 @@ export async function deleteValue(key: RegistryKey, value: string): Promise<void
...
-        } else {
-          onError(valueExistsError);
+        }
```
akrantz, line 127: I should fix the formatting. I've been switching between different codebases that use different brace formatting styles.

akrantz: fixed

akrantz:
OK, the tests pass now. On non-Windows platforms, the commands will fail with a Platform Not Supported error message. The tests now only run on Windows.

TCourtneyOwen (approved):

PR #324, opened 2020-09-28
Sideload to web

TCourtneyOwen:
-Enables sideloading to web for Excel, OneNote, PPT and Word
-Implements sideload-web command that takes a documentUrl, dev-server port and manifest file path
- Added tests


arttarawork (dismissed):

packages/office-addin-debugging/src/commands.ts
```diff
@@ -75,7 +76,7 @@ export async function start(manifestPath: string, platform: string | undefined,
         }
 
         await startDebugging(manifestPath, appTypeToDebug, app, debuggingMethod, sourceBundleUrlComponents,
-            devServer, devServerPort, packager, packagerHost, packagerPort, enableDebugging, enableLiveReload, openDevTools);
+            devServer, devServerPort, documentUrl, packager, packagerHost, packagerPort, enableDebugging, enableLiveReload, openDevTools);
```
akrantz, line 79: You must add new optional parameters at the end to avoid breaking changes.

packages/office-addin-debugging/src/start.ts
```diff
@@ -189,7 +190,7 @@ export async function runPackager(commandLine: string, host: string = "localhost
 export async function startDebugging(manifestPath: string, appType: AppType, app: OfficeApp | undefined,
     debuggingMethod: DebuggingMethod = defaultDebuggingMethod(),
     sourceBundleUrlComponents?: devSettings.SourceBundleUrlComponents,
-    devServerCommandLine?: string, devServerPort?: number,
+    devServerCommandLine?: string, devServerPort?: number, documentUrl?: string,
```
akrantz, line 193: again, the new parameter goes at the end.

packages/office-addin-debugging/src/start.ts
```diff
@@ -289,6 +290,15 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
             }
         }
 
+        if (isWebAppType && documentUrl !== undefined && devServerPort !== undefined) {
+            try {
+                console.log(`Sideloading the Office Add-in to web...`);
+                await sideloadWebAddin(documentUrl, manifestPath, devServerPort)
```
akrantz, line 296: I would really like to see sideloading a bit more unified between web and desktop. I haven't looked closely at the code here in a while though. Sideload is part of the manifest or dev-settings package and I think this should be true whether for desktop or web sideloading.

We should also discuss being able to specify the document when doing desktop sideload as well.

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -62,6 +62,12 @@ commander
     .option("-a,--app <app>", `The Office app to launch. ("Excel", "PowerPoint", or "Word")`)
     .action(commands.sideload);
 
+commander
+    .command("sideload-web <document-url> <manifest-path> <port>")
```
akrantz, line 66: Why a separate command? Let's try to make desktop and web sideload more unified.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -322,6 +322,18 @@ export async function sideload(manifestPath: string, command: commander.Command)
   }
 }
 
+export async function sideloadWeb(documentUrl: string, manifestPath: string, port: number, command: commander.Command) {
+  try {
+    const isTest: boolean = (command.test !== undefined);
+    const app: OfficeApp | undefined = command.app ? parseOfficeApp(command.app) : undefined;
+    await sideloadWebAddin(documentUrl, manifestPath, port, isTest);
+
+    // await sideloadAddIn(manifestPath, app, canPrompt);
```
akrantz, line 331: This feels incomplete.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -92,6 +92,40 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo): Pro
...
+/**
+ * Create an Office document url with query params which can be opened
+ * to register an Office add-in in Office Online.
+ * @param documentUrl Office Online document url
+ * @param manifestPath Path to the manifest file for the Office Add-in.
+ * @param port Port number for the dev-server.
+ * @param isTest Indicates whether to append test query param to suppress Office Online dialogs.
+ * @returns Document url with query params appended.
+ */
+export async function generateSideloadUrl(documentUrl: string, manifestPath: string, port: number, isTest:boolean = false): Promise<string> {
```
akrantz, line 104: nit: `isTest:` should have a space after it.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -92,6 +92,40 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo): Pro
...
+  const regEx: RegExp = /^.*[\\\/]/;
+  const devServerPortQueryParam = "&wdaddindevserverport=";
+  const manifestFileQueryParam = "&wdaddinmanifestfile=";
+  const manifestGuidQueryParam = "&wdaddinmanifestguid=";
+  const testQueryParam = "&wdaddintest=true";
+  const manifest = await readManifestFile(manifestPath);
+  const appsInManifest: OfficeApp[] = getOfficeAppsForManifestHosts(manifest.hosts);
+  appsInManifest.forEach(element => {
+    if (element === OfficeApp.Outlook || element === OfficeApp.Project) {
```
akrantz, line 113: Shouldn't this be done before calling this function? It would be clearer to have this check be separate and not doing when adding the query parameters for sideloading to the web.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -92,6 +92,40 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo): Pro
...
+      throw new Error(`${element} is not supported for sideloading to web.`);
+    }
+  });
+
+  if (!manifest.id) {
+    throw new Error("The manifest does not contain the id for the add-in.");
+  }
+  const manifestFile: string = manifestPath.replace(regEx, '');
```
akrantz, line 121: Why are you doing this?

akrantz (changes requested):
Courtney, you've made great progress here but I would like to have a call and see if we can improve the design here a bit. I've made some comments but also it's a bit hard to get the full picture from a quick review of the diffs here.

packages/office-addin-dev-settings/src/cli.ts
akrantz: image

Seems like we can have only one sideload command, but add parameters to allow for specifying whether desktop or web, and then have optional parameters that can be added which apply for web.

akrantz: The startDebugging() command already has AppType enum which will be defined as AppType.Desktop or AppType.Web. If the sideloadCommand has an optional AppType parameter added which can be used to specify whether to sideload to desktop or web. It can default to AppType.Desktop. 

For the implementation, the registration of the add-in can be skipped when appType is not Desktop.

There is just a small amount of logic from that point to generate the sideload file/url and then open it.

akrantz: It would be nice to augment desktop sideload to have an optional parameter to specify an existing file. It can validate that the file already has the add-in webextension id/version in it or error out if not.

Thus, the --document (or --doc if we want to shorten it) can then work for both desktop and web.

Sideload to web would need to have the --document parameter (at least for now). It still leaves open implementing a way to create a new doc or looking for one so we could relax the need to specify the existing document.

packages/office-addin-dev-settings/src/commands.ts
akrantz: From my previous comments, I think we unify the sideload() function using the AppType so we don't need this.

packages/office-addin-dev-settings/src/cli.ts
TCourtneyOwen: I thought of having a unified command as well initially but couldn't come up with a clean solution. I will revisit however

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: We need to extract just the file name from the manifestPath, so that we can add this as a query param.  This brings to mind that I should verify that this approach will work when the manifest is in a subfolder such as c:\office-addin-takpane\manifests\manifest.xml and that it also works when the path folders have spaces in them

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: sure, we can move this up in the stack

packages/office-addin-dev-settings/src/sideload.ts
akrantz: Do you really need the file name of the manifest? Is this just a function of what gets shown in the dialog? Would it be better to use the add-in name which you can get from the the manifest info?

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -92,6 +92,40 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo): Pro
...
+  let queryParms: string = `${devServerPortQueryParam}${port}${manifestFileQueryParam}${manifestFile}${manifestGuidQueryParam}${manifest.id}`;
```
akrantz, line 122: You can put the query param names with the & separator directly into this string so you just have the replacement for the value part after the =.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: We need to name of the manifest so the WAC code knows which file to pull from to get the manifest content. I am not prepared to make changes to the WAC code at this time if that's what you are suggesting?

TCourtneyOwen: e.g. the WAC code needs the file name so it can call https://127.0. 0.1<dev-server-port>/<manifest-file-name>

akrantz: OK, this is not just the file name but it really the full path of the url to the manifest without specifying that this is from localhost, and the port number is passed as a separate argument. This is a function of the fact that the dist folder contains the manifest file and webpack-dev-server serves it up. That will be just the file name part of manifest file path. It would probably be more correct to get the file path to the root of the dev server and remove that part from the path to the manifest file, however, i'm not sure how to do that.

If you just want to get the file path from the url, there are easier ways to do that. You don't need to use your own regex. There should be an easy to get the path component, or just the file name part of the path, from the url.

TCourtneyOwen: looks like path.basename will work to get the file name

packages/office-addin-dev-settings/src/cli.ts
TCourtneyOwen: I think the problem with having a doc param that can apply for desktop as well as web is that the doc must be of a specific type - it has to have a webextension xml part already present in it for our current sideloading code work or else we fail here:

  // replace the placeholder id and version
  const zipFile = zip.file(webExtensionPath);
  if (!zipFile) {
    throw new Error("webextension was not found.")
  }

I think having a doc param that applies for both web and desktop will confuse users and they will specify just any document that doesn't have a webextenion part in it already and the operation will fail.  

We could of course try rewriting our desktop sideloading code to try and accomodate this scenario, but I am not sure how difficult that would be at this point

akrantz: Yeah, it would be beneficial to be able to have the desktop sideload support that. There is an issue asking for it. Obviously the logic becomes different if we are creating the document versus checking that it already exists and using it. This should not be difficult to achieve.

akrantz:
```

  if (app) {
    const sideloadFile = await generateSideloadFile(app, manifest);

    await open(sideloadFile, { wait: false });
  }
```

In this code, if the sideloadFile is provided, then use it rather than generate the file.

TCourtneyOwen: I suppose we can just create a new webExtensionXml part if it doesn't already exist in the document specified via the doc paramter. Probably not too difficult

TCourtneyOwen:
I need to revert that last merge commit.  Not sure why it picked up all the files from origin/master. I explicitly tried to avoid that

akrantz:
You may want to rebase off master. You don't want to merge in a revert of a merge, so it won't be a big deal if you squash this PR for check-in, but don't merge it.

TCourtneyOwen:
I was thinking of just starting with a clean branch and kicking off a new PR.  Does that seem acceptable?  This new branch will already be up to date with master so no need for a merge

akrantz:
That's what rebasing effectively does except using the existing branch name. The commits on the branch would then be off of the latest commit on master. If you close this PR and start a new one, the only down side is you then lose the comments in the PR.

packages/office-addin-debugging/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "office-addin-debugging",
-  "version": "3.0.35",
+  "version": "3.0.34",
```
akrantz, line 3: I think you can revert these package.json changes.

packages/office-addin-debugging/src/start.ts
```diff
@@ -280,13 +281,11 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
...
         }
 
-        if (isDesktopAppType) {
-            try {
-                console.log(`Sideloading the Office Add-in...`);
-                await sideloadAddIn(manifestPath, app, true);
-            } catch (err) {
-                throw new Error(`Unable to sideload the Office Add-in. \n${err}`);
-            }
+        try {
```
akrantz, line 284: This is great. It just works for both desktop and web now.  :-)

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -101,29 +101,32 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo, docu
  * to register an Office add-in in Office Online.
  * @param manifestPath Path to the manifest file for the Office Add-in.
  * @param documentUrl Office Online document url
- * @param devServerPort Port number for the dev-server.
  * @param isTest Indicates whether to append test query param to suppress Office Online dialogs.
  * @returns Document url with query params appended.
  */
-export async function generateSideloadUrl(manifestPath: string, documentUrl: string, devServerPort: number, isTest: boolean = false): Promise<string> {
+export async function generateSideloadUrl(manifestPath: string, documentUrl: string,  isTest: boolean = false): Promise<string> {
```
akrantz, line 107: `generateSideloadFile()` takes the `manifest: ManifestInfo` parameter. It would be better to do the same here, since it could already be read and just pass in the info.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -101,29 +101,32 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo, docu
...
-    throw new Error("Missing --dev-server-port paramater.");
+  if (manifest.defaultSettings === undefined || manifest.defaultSettings.sourceLocation === undefined) {
+    throw new Error("The manifest does not contain the SourceLocation for the add-in")
   }
 
-  if (!manifest.id) {
-    throw new Error("The manifest does not contain the id for the add-in.");
+  const sourceLocationUrl: URL = new URL(manifest.defaultSettings.sourceLocation);
+  if (sourceLocationUrl.protocol.indexOf("https") === -1) {
+    throw new Error("Manifest SourceLocation protocol is not using HTTPS.");
```
akrantz, line 121: "The SourceLocation in the manifest does not use the HTTPS protocol."

TCourtneyOwen:
All our taskpane projects use HTTPS for the SourceLocation.  For example, Office-Addin-Taskpane:

  <DefaultSettings>
    <SourceLocation DefaultValue="https://localhost:3000/taskpane.html"/>
  </DefaultSettings>

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -101,29 +101,32 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo, docu
...
+  }
+
+  if (sourceLocationUrl.host.indexOf("localhost") === -1 && sourceLocationUrl.host.indexOf("127.0.0.1") === -1) {
+    throw new Error("Manifest SourceLocation is not using localhost.");
```
akrantz, line 125: `The hostname specified by the SourceLocation in the manifest is not supported for sideload. The hostname should be "localhost" or 127.0.0.1`

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -224,11 +286,19 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
-    const sideloadFile = await generateSideloadFile(app, manifest);
-
-    await open(sideloadFile, { wait: false });
+  if (isSideloadingSupportedForHost(app, isDesktop)) {
+    if (isDesktop) {
+      await registerAddIn(manifestPath);
+      sideloadFile = await generateSideloadFile(app, manifest, document);
+      await open(sideloadFile, { wait: false });
+    } else {
+      if (document === undefined) {
```
akrantz, line 295: Rather than check for document being defined here, you can pass it to generateSideloadUrl() and it can do the validation.

A better error message would be "The document url was not provided."  

Remember that you can get errors back when importing the function rather than using the command line so it is better to not reference the command line switch.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -224,11 +286,19 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
+        throw new Error("Missing --document paramater.");
+      }    
+      sideloadFile = await generateSideloadUrl(manifestPath, document, isTest);
+      await open(sideloadFile, { wait: false });
```
akrantz, line 299: The open() function is used for both desktop and web. You could move it down below the if statement,

packages/office-addin-debugging/src/commands.ts
```diff
@@ -75,7 +76,7 @@ export async function start(manifestPath: string, platform: string | undefined,
         }
 
         await startDebugging(manifestPath, appTypeToDebug, app, debuggingMethod, sourceBundleUrlComponents,
-            devServer, devServerPort, packager, packagerHost, packagerPort, enableDebugging, enableLiveReload, openDevTools);
+            devServer, devServerPort, packager, packagerHost, packagerPort, enableDebugging, enableLiveReload, openDevTools, document);
```
akrantz, line 79: We should consider a breaking change which uses an object with properties rather than an ever-growing list of parameters on this function. It could provide better flexibility in the future. This isn't required per-se for this change, but something I am thinking about and felt might be good to mention.

The downside is that it would be a breaking change. We could consider adding this as another method, e.g. startDebuggingWithOptions(options: StartDebugginOptions) and then this implementation could call that. It would keep the current function limited at the current functionality, or we could continue to extend it, but add the other method for better code design. We might want to consider separating this because for example, the packager stuff is not really as desired since ReactNative debugging is moving to being direct debugging only.

This is just food for thought.

TCourtneyOwen:
I would like to wrap up this PR so we can get the sideloading to web feature out the door soon.  Once this PR is done, our Writers are going to document the feature and then we consider it done.

Thanks

akrantz:
I am taking a look. Are there any breaking changes with the CLI commands or the exported package functions?

packages/office-addin-debugging/src/start.ts
```diff
@@ -191,13 +192,13 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
...
     packagerCommandLine?: string, packagerHost?: string, packagerPort?: string,
-    enableDebugging: boolean = true, enableLiveReload: boolean = true, openDevTools: boolean = false) {
+    enableDebugging: boolean = true, enableLiveReload: boolean = true,
+    openDevTools: boolean = false, document?: string) {
 
     try {
 
         const isWindowsPlatform = (process.platform === "win32");
         const isDesktopAppType = (appType === AppType.Desktop);
-        const isWebAppType = (appType === AppType.Web);
```
akrantz, line 200: Why is this removed? We may not want to assume that if not desktop that it implies web as we could add another platform type in the future.

TCourtneyOwen:
@akrantz yes it should now be unified between desktop and web

packages/office-addin-debugging/src/start.ts
akrantz: I think it is because there is now not a specific check and the sideload is done for all appTypes, which is good.

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -59,7 +59,10 @@ commander
 commander
     .command("sideload <manifest-path>")
     .description("Launch Office with the Office Add-in loaded.")
-    .option("-a,--app <app>", `The Office app to launch. ("Excel", "PowerPoint", or "Word")`)
+    .option("-a,--app <app>", `The Office app to launch. ("Excel", "OneNote", "PowerPoint", or "Word")`)
+    .option("-p,--platform <platform>", `Plaform to launch sideloading for ("desktop" or "web")`)
+    .option("-d,--document <document>", `The location of the document to be sideloaded - this can be an absolute file path or url`)
+    .option("-t,--test <test>", `Used for testing to ignore Office Online registration dialogs (Internal only)`)
```
akrantz, line 65: Should this option be hidden? It can be abused if it is known and documented.

TCourtneyOwen:
@akrantz I didn't introduce any breaking changes here (e.g having an object with properties rather than an ever-growing list of parameters for startDebugging) - at least none that I know of

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -92,13 +96,54 @@ async function generateSideloadFile(app: OfficeApp, manifest: ManifestInfo): Pro
...
+    throw new Error("The hostname specified by the SourceLocation in the manifest is not supported for sideload. The hostname should be 'localhost' or 127.0.0.1.");
+  }
+
+  let queryParms: string = `&wdaddindevserverport=${sourceLocationUrl.port}&wdaddinmanifestfile=${manifestFileName}&wdaddinmanifestguid=${manifest.id}`;
+  
+  if (isTest) {
+    queryParms = `${queryParms}${testQueryParam}`;
+  }
+
+  return`${documentUrl}${queryParms}`;
```
akrantz, line 137: nit: need a space before the ` char.

TCourtneyOwen:
Looks like my last change introduced a test failure. Investigating

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -267,39 +261,46 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
-        app = appsInManifest[0];
-        break;
-      default:
-        if (canPrompt) {
-          app = await chooseOfficeApp(appsInManifest);
-        } else {
-          throw new Error("Please specify the Office app.");
-        }
-        break;
+  if (platform && platform == AppType.Web && document === undefined) {
```
akrantz, line 264: I think this would be better in the else clause below just before the generateSideloadUrl() call.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -267,39 +261,46 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
+    throw new Error(`Sideload to web requires the document option parameter.`);
```
akrantz, line 265: Suggestion for error working: "For sideload to web, you need to specify a document url."

It might also be good if there is a documentation page related to this, to provide this in the error, e.g. "...specify a document url. For more info, see aka.ms/sideload-to-web."

akrantz (approved):

PR #327, opened 2020-10-07
[office-addin-debugging] add option to skip prompt for enabling edge loopback

zhngx:


akrantz (changes requested):
I don't think we really need to make this change. The problem on the servers is that the appcontainer for the edge webview is not registered properly.

TCourtneyOwen:
@zhngx did you verify that this change in fact addresses the problem on VMs?

akrantz:
It just skips the prompt, but it doesn't address the problem. If the Edge webview cannot access localhost, the add-in won't run.

TCourtneyOwen:
Right, that's exactly what I was thinking, so I am not sure of the real value of this change

zhngx:
It seem to work ok if I just skip the prompt and run through the tests. But sure it will be better to address the root issue instead of this temporary fix. I can always downgrade the package version to make sure our automation are running fine before we fix the root cause.

PR #335, opened 2020-11-10
Update Typescript version and fix a couple vulnerabilties

TCourtneyOwen:


TCourtneyOwen:
Need to investigate why CI job is failing on install. This passes locally on my machine

akrantz:
With the major version update of the TypeScript package, there are likely eslint packages which should be updated to match.

TCourtneyOwen:
I am going to abandon this PR and create a new one that just updates the vulnerable rimraf and concurrently packages. Typescript updates appear to be causing issues

PR #345, opened 2020-11-30
Dependabot addition

arttarawork:


.vscode/dependabot.yml
```diff
@@ -0,0 +1,86 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
```
TCourtneyOwen, line 8: I am not familiar with dependabot but seems like daily may be too often. Can we do weekly?

arttarawork: Changed to weekly with the latest check-in

arttarawork:
Our pipeline tests are failing, but they're failing for reasons unrelated to this PR. The same test failure is occurring in the master branch of this repo.

.vscode/dependabot.yml
akrantz: Why is this in the .vscode folder? I don't think this is related to VS Code.

akrantz:
If the tests on the master branch are not reliable, they need to be made reliable before merging more PRs.

.vscode/dependabot.yml
TCourtneyOwen: Yes, I agree with Adam, this seems strange

arttarawork: Tests have moved to the proper .github folder, per Dependabot instructions: https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/enabling-and-disabling-version-updates

TCourtneyOwen:
All the CI agents are down I believe due to the Building 36 power outage over the weekend

millerds:
This isn't needed.

PR #346, opened 2020-11-30
Add ability to read application secret from project .ENV file

TCourtneyOwen:
- We have been getting complaints about the portability of the add-in projects to other environments because we currenrtly rely on the application secret being stored locally on the same machine the add-in is running on.
- By being able to add the secret to the .ENV users can port the project over to other environments


TCourtneyOwen:
Closing this PR but keeping this branch alive until we get a better idea of what our approach should be here for handing the secret

PR #361, opened 2021-01-06
start debugging with options

wandyezj:


TCourtneyOwen (commented):
Don't we need to update the call to startDebugging in packages/office-addin-debugging/src/commands.ts as well?

wandyezj:
> Don't we need to update the call to startDebugging in packages/office-addin-debugging/src/commands.ts as well?

This review is currently set up to only add a new function `startDebuggingWithOptions` and not make any breaking changes.

I'm happy to completely replace the existing `startDebugging` with the new signature if that's preferable.

TCourtneyOwen:
I think let's just update the existing method and not add a new one for the sake of simplicity

packages/office-addin-debugging/src/start.ts
```diff
@@ -172,6 +172,119 @@ export async function runPackager(commandLine: string, host: string = "localhost
...
+     * If provided, the document to open for sideloading to web.
+     */
+    document?: string,
+};
+
+/**
+ * Start Debugging
+ * @param options The debugging options.
+ */
+export async function startDebuggingWithOptions(options:StartDebuggingOptions) {
```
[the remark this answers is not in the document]
akrantz: Also, need a space after `options:` before the type.

akrantz (changes requested):

packages/office-addin-debugging/src/start.ts
```diff
@@ -172,6 +172,119 @@ export async function runPackager(commandLine: string, host: string = "localhost
     }
 }
 
+
+export interface StartDebuggingOptions {
```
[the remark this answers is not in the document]
akrantz: I cannot remember the differences off of the top of my head. I'm just wondering the best way to do this.

akrantz:
Rather than use the `master` branch in your fork, you should make the a named topic branch and use that.

packages/office-addin-debugging/src/start.ts
```diff
@@ -203,11 +270,13 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
...
         const canEnableLiveReload: boolean = isDesktopAppType && isProxyDebuggingMethod && !!packagerCommandLine;
+        // only enable live reload if it can be enabled
+        const liveReload = enableLiveReload && canEnableLiveReload;
+
         let packagerPromise: Promise<void> | undefined;
         let devServerPromise: Promise<void> | undefined;
 
-        // only enable live reload if it can be enabled
-        enableLiveReload = enableLiveReload && canEnableLiveReload;
+
```
akrantz, line 279: I think there is an extra blank line or two here that you can remove. You should just have one blank line between variable declarations and the code after.

packages/office-addin-debugging/src/start.ts
```diff
@@ -203,11 +270,13 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
         // live reload can only be enabled for the desktop app type
         // when using proxy debugging and the packager
         const canEnableLiveReload: boolean = isDesktopAppType && isProxyDebuggingMethod && !!packagerCommandLine;
+        // only enable live reload if it can be enabled
+        const liveReload = enableLiveReload && canEnableLiveReload;
```
akrantz, line 274: Is there a reason to rename this from `enableLiveReload` to just `liveReload`? I'd prefer to keep the original name unless some reason to rename it.

akrantz:
I think there might be other callers of the startDebugging() function, perhaps in test code. Can you please just double-check?

akrantz (changes requested):
Looking good, but there are some minor fixes needed. Thanks.

packages/office-addin-debugging/src/start.ts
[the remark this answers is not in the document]
wandyezj: Would it be ok to enable prettier for this repository?

I've found prettier quite convenient to maintaining code style.

wandyezj:
> Rather than use the `master` branch in your fork, you should make the a named topic branch and use that.

I was looking at the README in this repository root, and I couldn't find the document explaining the preferred way of making contributions. 

Would you be able to link me to the appropriate document?

akrantz:
That's the standard way in Github to work so we didn't document this specifically in the repository, but I can see how explaining better how others can contribute would be good. Let's do this in a separate PR. Basically, we you just create a branch in your fork, make the PR from the branch, and then we merge.

packages/office-addin-debugging/src/start.ts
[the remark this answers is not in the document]
akrantz: We could but it should be done as a separate pull request. We could even merge that pull request first, and then after that is merged to master, then merge it in here.

The only difference here with the monorepo is that we need to enable prettier for each package. 

There is one prettier config that we should override with is making printWidth = 120.

packages/office-addin-debugging/src/start.ts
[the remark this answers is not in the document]
wandyezj: From a docs point of view an interface may be preferable. 

Automatic documentation generators such as [API Extractor](https://api-extractor.com/) allow individual interface members to be documented.

For Node's @types, an interface is used to describe options objects.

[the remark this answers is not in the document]
akrantz: Looking at the TypeScript docs, I think using `interface` is good.

packages/office-addin-debugging/src/start.ts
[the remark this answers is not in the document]
akrantz: Actual, for prettier we should also integrate with eslint so both lint and prettier are working.

packages/office-addin-debugging/src/start.ts
wandyezj: Yes, the original code overwrote the input variable called `enableLiveReload` generally I try to avoid overwriting input variables because it can make it confusing to debug later.

packages/office-addin-debugging/src/start.ts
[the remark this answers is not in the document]
wandyezj: I'm happy to enable prettier for all packages pointing at a single config in another review.

TCourtneyOwen:
@wandyezj I don't think we document that the preferred method is to create a branch off of a fork.  I suppose we could add that to the "Contributing" section in the ReadMe

packages/office-addin-debugging/src/start.ts
```diff
@@ -172,28 +172,95 @@ export async function runPackager(commandLine: string, host: string = "localhost
...
- * @param devServerCommandLine If provided, starts the dev server.
- * @param devServerPort If provided, port to verify that the dev server is running.
- * @param documentUrl If provided, the document to open for sideloading to web.
- * @param packagerCommandLine If provided, starts the packager.
- * @param packagerHost Specifies the host name of the packager.
- * @param packagerPort Specifies the port of the packager.
- * @param enableDebugging If false, start without debugging.
+ * @param options start debugging options.
  */
-export async function startDebugging(manifestPath: string, appType: AppType, app: OfficeApp | undefined,
```
wandyezj, line 190: @akrantz How come the `app` variable is allowed to be undefined? Should `app` be a required parameter?

packages/office-addin-debugging/src/start.ts
akrantz: Thanks for clarifying that. I agree with your rationale.

packages/office-addin-debugging/src/start.ts
akrantz: No, we will prompt the user if not specified.

akrantz: When you use the CLI, you can do "office-addin-debugging start <manifest>" and it if there is more than one app in the manifest, it will prompt for which one, or the user can add `--app excel` to specify. If there's only one app in the manifest, it doesn't bother prompting. The VS Code configs specify the app, since you need a different launch target for each app.

akrantz: I don't think you should try to change whether or not undefined is allowed for particular parameters here or the defaults. If you see some that you really think we should change, we can discuss it.

packages/office-addin-debugging/src/start.ts
```diff
@@ -172,72 +172,87 @@ export async function runPackager(commandLine: string, host: string = "localhost
...
     packagerHost?: string,
+
     /**
      * Specifies the port of the packager.
      */
     packagerPort?: string,
+
     /**
-     * If false, start without debugging.
+     * Enable debugging.
```
akrantz, line 231: "If false, start without debugging; if true or undefined, start with debugging enabled."

akrantz: Please make sure that it continues to have the behavior to enable debugging if undefined.

packages/office-addin-debugging/src/start.ts
wandyezj: I see, perhaps the documentation should be updated to specify that if `app` is undefined it will prompt to specify the application?

packages/office-addin-debugging/src/start.ts
```diff
@@ -173,41 +173,124 @@ export async function runPackager(commandLine: string, host: string = "localhost
...
- * @param debuggingMethod The method to use when debugging.
- * @param sourceBundleUrlComponents Specify components of the source bundle url.
- * @param devServerCommandLine If provided, starts the dev server.
- * @param devServerPort If provided, port to verify that the dev server is running.
- * @param documentUrl If provided, the document to open for sideloading to web.
- * @param packagerCommandLine If provided, starts the packager.
- * @param packagerHost Specifies the host name of the packager.
- * @param packagerPort Specifies the port of the packager.
- * @param enableDebugging If false, start without debugging.
+ * `startDebugging` options.
```
akrantz, line 176: Consider removing the back ticks and making it `startDebugging() options` or `Options for startDebugging() function.`

packages/office-addin-debugging/src/start.ts
```diff
@@ -173,41 +173,124 @@ export async function runPackager(commandLine: string, host: string = "localhost
...
+     */
+    document?: string,
+}
+
+/**
+ * Start debugging
+ * @param options `startDebugging` options.
+ */
+export async function startDebugging(options: StartDebuggingOptions) {
+
```
akrantz, line 258: nit: I think this blank line can be removed.

akrantz (changes requested):
Almost there. Some minor tweaks suggested.

akrantz:
@wandyezj Could go through the comments and mark them resolved if you've addressed them? I think there are a few comments I made that are still outstanding.

akrantz:
Also, you should merge master into this.

packages/office-addin-debugging/src/start.ts
akrantz: @wandyezj Let me know if you disagree with this, but I don't see a change here. Did you want to keep the back ticks?

wandyezj: From what I've observed in other documentation it's standard to use backticks around the names of referred to functions or objects.

akrantz: Interesting. I haven't noticed this myself but would like to understand if this would be a better pattern to follow. Does this affect how it appears in VS Code IntelliSense?

akrantz (approved):

packages/office-addin-debugging/src/start.ts
```diff
@@ -173,41 +173,123 @@ export async function runPackager(commandLine: string, host: string = "localhost
...
+ * `startDebugging` options.
  */
-export async function startDebugging(manifestPath: string, appType: AppType, app: OfficeApp | undefined,
-    debuggingMethod: DebuggingMethod = defaultDebuggingMethod(),
+export interface StartDebuggingOptions {
+
+    /**
+     * The path to the manifest file.
+     */
+    manifestPath: string,
```
akrantz, line 183: Joe, based on our discussion earlier, should some of these be required parameters for the startDebugging() function instead of in the options object?

wandyezj: In this case, since there are so many options, I think it make sense to keep it as is.

TCourtneyOwen:
Before we merge this, I want to get a clear sense if this is a breaking change (i.e. will will need to update other template and tooling to accommodate this change. I am OK with a breaking change but just want to know what, if any, additional changes we will need to make to accommodate this change

akrantz:
When this gets published, the major version of this package should be updated.

Other packages which consume this will need to the latest version, and then update to call StartDebugging() with the options object instead of just the parameters. `startDebugging(manifestPath, etc.)` gets replaced with:

`startDebugging({ manifestPath, enableLiveReload: false, etc. });`

It's not a difficult change.

PR #362, opened 2021-01-07
Enable Outlook desktop sideloading for Windows

TCourtneyOwen:
With these changes:
- Office-Add-Scripts will add the manifest path to HKEY_CURRENT_USER\SOFTWARE\Microsoft\Office\16.0\Wef\OutlookSideloadManifestPath if the platform is desktop and app is Outlook
- Code also does a check to ensure that the manifest contains a valid Outlook Host element
- Once the value is set, startDebugging will launch Outlook desktop.
- Outlook desktop then has code to see if the reg value is set.  If it is it will show a warning dialog that Outlook will attempt to register an add-in:

image

- If the user says OK to the warning dialog, Outlook will then attempt to register the add-in. If the manifest is invalid for some reason of the manifest path is not correct, Outlook will surface an informative error to the user

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -281,6 +281,12 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
             }
         }
 
+        // enable Outlook sideloading
+        if (isDesktopAppType && app == OfficeApp.Outlook) {
+            await devSettings.enableOutlookSideloading(manifestPath);
+            console.log(`Enabled sideloading for Outlook.`);
+        }
+
         try {
             console.log(`Sideloading the Office Add-in...`);
             await sideloadAddIn(manifestPath, app, true, appType, document, false /* isTest */);
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -283,7 +283,7 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
 
         try {
             console.log(`Sideloading the Office Add-in...`);
-            await sideloadAddIn(manifestPath, app, true, appType, document, false /* isTest */);
+            await sideloadAddIn(manifestPath, appType, app, true, document, false /* isTest */);
         } catch (err) {
             throw new Error(`Unable to sideload the Office Add-in. \n${err}`);
         }
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -281,12 +281,6 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
             }
         }
 
-        // enable Outlook sideloading
-        if (isDesktopAppType && app == OfficeApp.Outlook) {
-            await devSettings.enableOutlookSideloading(manifestPath);
-            console.log(`Enabled sideloading for Outlook.`);
-        }
-
         try {
             console.log(`Sideloading the Office Add-in...`);
             await sideloadAddIn(manifestPath, app, true, appType, document, false /* isTest */);
--- a/packages/office-addin-dev-settings/src/cli.ts
+++ b/packages/office-addin-dev-settings/src/cli.ts
@@ -88,7 +88,8 @@ commander
     .action(commands.webView)
     .on("--help", () => {
         console.log("\nFor [web-view-type], choose one of the following values:\n");
-        console.log("\t'edge' for Microsoft Edge");
+        console.log("\t'edge' or 'edge-chromium' for Microsoft Edge (Chromium)");
+        console.log("\t'edge-legacy' for the legacy Microsoft Edge (EdgeHTML)");
         console.log("\t'ie' for Internet Explorer 11");
         console.log("\t'default' to remove any preference");
         console.log("\nOmit [web-view-type] to see the current setting.");
--- a/packages/office-addin-dev-settings/src/cli.ts
+++ b/packages/office-addin-dev-settings/src/cli.ts
@@ -57,13 +57,13 @@ commander
     .action(commands.runtimeLogging);
 
 commander
-    .command("sideload <manifest-path> <platform>")
+    .command("sideload <manifest-path> <app-type.")
     .description("Launch Office with the Office Add-in loaded.")
     .option("-a,--app <app>", `The Office app to launch. ("Excel", "PowerPoint", or "Word")`)
     .option("-d,--document <document>", `The location of the document to be sideloaded - this can be an absolute file path or url`)
     .action(commands.sideload)
     .on("--help", () => {
-        console.log("\nFor <platform>, choose one of the following values:\n");
+        console.log("\nFor <app-type>, choose one of the following values:\n");
         console.log("\t'desktop' for sideloading desktop add-ins");
         console.log("\t'web' for sideloading web add-ins");
     });
--- a/packages/office-addin-dev-settings/src/cli.ts
+++ b/packages/office-addin-dev-settings/src/cli.ts
@@ -57,12 +57,16 @@ commander
     .action(commands.runtimeLogging);
 
 commander
-    .command("sideload <manifest-path>")
+    .command("sideload <manifest-path> <platform>")
     .description("Launch Office with the Office Add-in loaded.")
     .option("-a,--app <app>", `The Office app to launch. ("Excel", "PowerPoint", or "Word")`)
-    .option("-p,--platform <platform>", `Plaform to launch sideloading for ("desktop" or "web")`)
     .option("-d,--document <document>", `The location of the document to be sideloaded - this can be an absolute file path or url`)
-    .action(commands.sideload);
+    .action(commands.sideload)
+    .on("--help", () => {
+        console.log("\nFor <platform>, choose one of the following values:\n");
+        console.log("\t'desktop' for sideloading desktop add-ins");
+        console.log("\t'web' for sideloading web add-ins");
+    });
 
 commander
     .command("source-bundle-url <manifest-path>")
--- a/packages/office-addin-dev-settings/src/commands.ts
+++ b/packages/office-addin-dev-settings/src/commands.ts
@@ -314,18 +314,18 @@ export async function runtimeLogging(command: commander.Command) {
   }
 }
 
-export async function sideload(manifestPath: string, platform: string, command: commander.Command) {
+export async function sideload(manifestPath: string, appType: string, command: commander.Command) {
   try {
     const app: OfficeApp | undefined = command.app ? parseOfficeApp(command.app) : undefined;
     const canPrompt = true;
     const document: string | undefined = command.document ? command.document : undefined;
     const isTest: boolean | undefined = command.test ? true : false;
 
-    if (platform !== AppType.Desktop && platform !== AppType.Web) {
-      throw new Error(`Unsupported sideload plaform argument: ${platform}`);
+    if (appType !== AppType.Desktop && appType !== AppType.Web) {
+      throw new Error(`Unsupported sideload plaform argument: ${appType}`);
     }
 
-    await sideloadAddIn(manifestPath, platform, app, canPrompt, document, isTest);
+    await sideloadAddIn(manifestPath, appType, app, canPrompt, document, isTest);
   } catch (err) {
     logErrorMessage(err);
   }
--- a/packages/office-addin-dev-settings/src/commands.ts
+++ b/packages/office-addin-dev-settings/src/commands.ts
@@ -155,14 +155,6 @@ export async function enableLiveReload(manifestPath: string) {
   }
 }
 
-export async function enableOutlookSideloading(manifestPath: string) {
-  try {
-    await devSettings.enableOutlookSideloading(manifestPath);
-  } catch (err) {
-    logErrorMessage(err);
-  }
-}
-
 export async function enableRuntimeLogging(path?: string) {
   try {
     const logPath = await devSettings.enableRuntimeLogging(path);
--- a/packages/office-addin-dev-settings/src/commands.ts
+++ b/packages/office-addin-dev-settings/src/commands.ts
@@ -311,15 +311,18 @@ export async function runtimeLogging(command: commander.Command) {
   }
 }
 
-export async function sideload(manifestPath: string, command: commander.Command) {
+export async function sideload(manifestPath: string, platform: string, command: commander.Command) {
   try {
     const app: OfficeApp | undefined = command.app ? parseOfficeApp(command.app) : undefined;
     const canPrompt = true;
     const document: string | undefined = command.document ? command.document : undefined;
     const isTest: boolean | undefined = command.test ? true : false;
-    const platform: AppType | undefined = command.platform ? command.platform : undefined;
 
-    await sideloadAddIn(manifestPath, app, canPrompt, platform, document, isTest);
+    if (platform !== AppType.Desktop && platform !== AppType.Web) {
+      throw new Error(`Unsupported sideload plaform argument: ${platform}`);
+    }
+
+    await sideloadAddIn(manifestPath, platform, app, canPrompt, document, isTest);
   } catch (err) {
     logErrorMessage(err);
   }
--- a/packages/office-addin-dev-settings/src/commands.ts
+++ b/packages/office-addin-dev-settings/src/commands.ts
@@ -244,12 +244,15 @@ export function parseWebViewType(webViewString?: string): devSettings.WebViewTyp
     case "internet explorer":
     case "internetexplorer":
       return devSettings.WebViewType.IE;
-    case "edge":
+    case "edgelegacy":
+    case "edge-legacy":
     case "spartan":
       return devSettings.WebViewType.Edge;
     case "edge chromium":
     case "edgechromium":
+    case "edge-chromium":
     case "anaheim":
+    case "edge":
       return devSettings.WebViewType.EdgeChromium;
     case "default":
     case "":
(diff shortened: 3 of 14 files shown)


TCourtneyOwen:
Investigating build error in CI loop

TCourtneyOwen:
I think the failure is due to a circular dependency between office-addin-dev-settings -> office-addin-debugging

akrantz:
There shouldn't be a circular dependency. office-addin-debugging can use stuff from office-addin-dev-settings, but not also the other way.

packages/office-addin-debugging/src/start.ts
```diff
@@ -281,6 +281,12 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
             }
         }
 
+        // enable Outlook sideloading
+        if (isDesktopAppType && app == OfficeApp.Outlook) {
```
akrantz, line 285: Could this not be integrated more into the sideloadOfficeAddIn() call below, so if the app is Outlook, it then does the different stuff? We really don't neeed to the console output to appear differently.  Does the "Enabled sideloading for Outlook" message mean much?

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -286,8 +301,12 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
 
   if (isDesktop && app) {
     if (isSideloadingSupportedForDesktopHost(app)) {
-      await registerAddIn(manifestPath);
-      sideloadFile = await generateSideloadFile(app, manifest, document);
+      if (app == OfficeApp.Outlook) {
```
akrantz, line 304: It looks like you could write the dev settings here just before you get the Outlook exe path and launch it.

akrantz: I also think you don't really need to check whether or not it is already in the registry. It can just write the registry value each time and it is OK if it is writing a previous value (as you can only sideload one add-in at a time).

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -3,6 +3,7 @@
 
 import * as fs from "fs";
 import * as jszip from "jszip";
+import { startDetachedProcess } from "office-addin-debugging";
```
akrantz, line 6: Is this the only reason for office-addin-dev-settings to use office-addin-debugging?  We might want to discuss moving the process/port stuff out of office-addin-debugging to a "lower level' package.

TCourtneyOwen: Yes so I can use startDetachedProcess.  I can work around this by creating a duplicate instance of this method in office-addin-dev-settings, but obviously I would rather not have to do this

akrantz:
Courtney, it would help if the description had a brief explanation of this work. I think what happens is that for outlook it just registers the add-in and launches Outlook, and then the app does the registration of the add-in after seeing the manifest listed in the registry.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: I think this may make it more difficult to test but I will have to see

akrantz: Once you see that the app is outlook, it could call to a separate function for doing the outlook sideloading.

packages/office-addin-dev-settings/src/sideload.ts
akrantz: Rather than copy it, you can just move process.ts from office-addin-debugging to office-addin-dev-settings, then have office-addin-debugging import it (and also re-export it for backward compatibility).

packages/office-addin-debugging/src/start.ts
TCourtneyOwen: sure we can do that - we really don't need the logging

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: I followed your guidance here and moved everything, including enableOutlookSideloading, into a separate sideloadOutlookAddin function

akrantz: Thanks. Did you push a new commit? I don't see one.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: Moved process.ts functions to office-addin-dev-settings, and then have existing process.ts in office-addin-debugging import those functions and call them for backwards compat

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: No not yet - wrapping up remaining comments

TCourtneyOwen:
Regarding this comment:  "I also think you don't really need to check whether or not it is already in the registry. It can just write the registry value each time and it is OK if it is writing a previous value (as you can only sideload one add-in at a time)."

The only place I check to see if the registry is already set is in our tests and I do this to ensure the registry is cleaned up before running the test so the test is more valid

akrantz:
OK, but for the tests, you could just have the test clear the registry value without checking as well. I guess it doesn't neccessarily hurt to have the isEnabled() function, but it might be simpler to not include it if it is really not needed.

One thing I wanted to chat with you about is whether the register/unregister add-in functions in dev settings should handle Outlook rather than having the separate function to enable sideload for Outlook. If you use the office-addin-dev-settings register command from the CLI, and then manually launch Outlook, would that still work? The way it is not I think that CLI wouldn't work, since it is only from office-addin-debugging start command (startDebugging function) which has the support for registering Outlook add-ins. 

Perhaps if the registration for Outlook is moved into the regular register/unregister sideload paths that would make sense. That may require some small tweaks that register/isRegistered/unregister functions work for Outlook but might make things more consistent, even though Outlook is a bit different when it comes to registration and sideload.  Feel free to give me a call to discuss.

TCourtneyOwen:
The one test failing in the Mac CI loop is an SSO test and presumably unrelated to this change.  Investigating. I am guess there is some process that's not closing successfully in a previous test and thus the test is unable to start the SSO service

TCourtneyOwen:
Yup just as I suspected, there was already a node process running on the port where the SSO service is supposed to be started so it couldn't start it and the test failed.  Killed the process, reran the job on Mac and it passes

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -199,12 +201,32 @@ function getWebExtensionPath(
 }
 
 function isSideloadingSupportedForDesktopHost(app: OfficeApp): boolean {
-  if (app === OfficeApp.Outlook || app === OfficeApp.Project || app === OfficeApp.OneNote) {
+  if (app === OfficeApp.Outlook && process.platform !== "win32" || app === OfficeApp.Project || app === OfficeApp.OneNote) {
```
akrantz, line 204: As we discussed, let's have a way to opt-in to the sideload to Outlook feature since it hasn't rolled out to production yet. For example, allow an environment variable to be set to enable the feature.

You might consider having the logic be for the true cases rather than the false cases.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -199,12 +201,32 @@ function getWebExtensionPath(
...
     return false;
   }
   return true;
 }
 
+function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
+  if (app === OfficeApp.Outlook || app === OfficeApp.Project) {
```
akrantz, line 211: I also think it would be better to have the logic reflect the apps that do support sideload and return true for them, or fallback to returning false at the end.

packages/office-addin-dev-settings/src/sideload.ts
akrantz: When you push new commits, it doesn't tell others, so it can help to provide a message when you want the additional changes review.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -287,20 +309,32 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
-      sideloadFile = await generateSideloadFile(app, manifest, document);
+      if (app == OfficeApp.Outlook) {
+        sideloadFile = await getOutlookExePath();
+      } else {
+        sideloadFile = await generateSideloadFile(app, manifest, document);
+      }
     } else {
       throw new Error(`Sideload is not supported for ${app} on ${AppType.Desktop}.`);
     }
   } else {
```
akrantz, line 320: I think it would be better to make this else be `else if (isWeb && app)`. You then could remove other checks for whether this is web from inside the clause.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -287,20 +309,32 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
-    if (platform && platform == AppType.Web && document === undefined) {
+    if (platform && platform == AppType.Web && document === undefined && app && isSideloadingSupportedForWebHost(app)) {
```
akrantz, line 321: You wouldn't need to check "app" here if done earlier. You can put the `isSideloadingSupportedForWebHost(app) check first and return the error if not, and then put the check for the document after it.

akrantz: You shouldn't need to check `platform && platform == AppType.Web) if you check for it earlier.

akrantz (changes requested):
This is close to being ready. Let's add the ability to have this off by default but a way to opt-in so we can merge it before the Outlook Win32 app support has rolled out to production.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: Are you referring to this part being "else is"?  I am not clear from your comment:

 else {
    if (platform && platform == AppType.Web && document === undefined && app && isSideloadingSupportedForWebHost(app)) {
      throw new Error(`For sideload to web, you need to specify a document url.`);
    }

    if (app && !isSideloadingSupportedForWebHost(app)) {
      throw new Error(`Sideload to web is not supported for ${app}.`);
    }
    
    const manifestFileName: string = path.basename(manifestPath);
    sideloadFile = await generateSideloadUrl(manifestFileName, manifest, document, isTest);
  }

akrantz: Sorry if not clear, there is `if (isDesktop && app) { ... }` and then after that can be `else if (isWeb && app)`.  I think at this point `app is required` so it might make sense to do a check `if (!app)` and throw an error about it. The point is that I think if you look over this it can be simplified.

akrantz: I think the `platform && platform == AppType.Web` can be simplified to `const isWeb == ...`. Of course, it also might be better to error out if there is no platform specified, and then use a switch statement for `switch (platform)` `case  AppType.Desktop:` and AppType.Web. Lots of ways this code could be improved.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: I will add a flag to opt out of Outlook sideload as we discussed

TCourtneyOwen:
It dawned on me that the sideload "platform" option is not really optional anymore, given that we now support desktop and web.  For the sake of backwards compat, I guess we can leave this as optional and have it always default to desktop if the platform isn't specified.  Thoughts on what we should do here?

akrantz:
If we set a default, then it will be defined whether by the user. I suspect that for the templates we already default to desktop, but the config value in the template could be switched. For example, "npm run start" is effectively "npm run start:desktop" and it is defined. But there are cases where web is a reasonable default too, for example on Linux where you can't use the desktop apps. I don't think we need to force a default, but rather should just have an error if the platform is not provided, and most likely there will already be some sort of value passed in. If we find otherwise later we can address it.

TCourtneyOwen:
Should we then update our sideload help to be the following:
.command("sideload <manifest-path> <platform>") 

and remove platform as an optional value here:
.option("-p,--platform <platform>", `Plaform to launch sideloading for ("desktop" or "web")`)

akrantz:
I think so. I was thinking more about start which I believe uses this pattern and makes sense to do also for sideload by itself.

sideload <manifest-path> [platform]

I think the question is whether we make platform optional and default to desktop (which would keep current behavior), rather than requiring it. 

I think that is better than having "--platform" as an option.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: Made some changes here - let me know if you aren't OK with them

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -309,25 +310,22 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
   if (isDesktop && app) {
     if (isSideloadingSupportedForDesktopHost(app)) {
       await registerAddIn(manifestPath);
-      if (app == OfficeApp.Outlook) {
+      if (app == OfficeApp.Outlook && isOutlookSideloadEnabled) {
```
akrantz, line 313: if the app is Outlook and sideload is not enabled, it shouldn't go to the else corresponding to the if when false. That would be non-outlook sideload. It should instead result in the error "Sideload is not supported for Outlook on Desktop."

akrantz (changes requested):
I need to still review the latest commit you pushed...

packages/office-addin-debugging/src/start.ts
```diff
@@ -283,7 +283,7 @@ export async function startDebugging(manifestPath: string, appType: AppType, app
 
         try {
             console.log(`Sideloading the Office Add-in...`);
-            await sideloadAddIn(manifestPath, app, true, appType, document, false /* isTest */);
+            await sideloadAddIn(manifestPath, appType, app, true, document, false /* isTest */);
```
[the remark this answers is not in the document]
TCourtneyOwen: I had to change the order because we are now making it a required param, and a required param cannot be after an optional param (i.e. after "app)

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -57,12 +57,16 @@ commander
...
-    .command("sideload <manifest-path>")
+    .command("sideload <manifest-path> <platform>")
     .description("Launch Office with the Office Add-in loaded.")
     .option("-a,--app <app>", `The Office app to launch. ("Excel", "PowerPoint", or "Word")`)
-    .option("-p,--platform <platform>", `Plaform to launch sideloading for ("desktop" or "web")`)
     .option("-d,--document <document>", `The location of the document to be sideloaded - this can be an absolute file path or url`)
-    .action(commands.sideload);
+    .action(commands.sideload)
+    .on("--help", () => {
+        console.log("\nFor <platform>, choose one of the following values:\n");
```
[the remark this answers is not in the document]
TCourtneyOwen: I believe this has been fixed - not sure if this is comment came in before or after my last commit

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -311,15 +311,18 @@ export async function runtimeLogging(command: commander.Command) {
   }
 }
 
-export async function sideload(manifestPath: string, command: commander.Command) {
+export async function sideload(manifestPath: string, platform: string, command: commander.Command) {
```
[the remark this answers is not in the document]
TCourtneyOwen: As far as I can tell, this breaking change does not impact our existing project templates. I did testing of Office-Add-Taskpane with this change, and "npm start" and "npm run start web" still work

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: Ah stupid mistake - will fix this

TCourtneyOwen: Moved this check up into isSideloadingSupportedForDesktopHost

packages/office-addin-dev-settings/src/commands.ts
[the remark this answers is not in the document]
akrantz: Sure, but this is an exported function so code that is using it will be broken by adding the parameter in the middle. I strive to follow semantic versioning properly, which means we should just bump the major version of this when we release it. Unless we add it to the end, which would make it a non-breaking change.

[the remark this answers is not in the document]
TCourtneyOwen: I personally would like for this to be a non-breaking change, but not sure if this is possible if we opt to make it a required param, which I think is the right thing to do.  Not sure I understand this comment, which would seem to suggest making it an optional param again:  "Unless we add it to the end, which would make it a non-breaking change"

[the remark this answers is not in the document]
akrantz: With JavaScript, any parameters not specified when a call is made have the value `undefined` so adding additional parameters won't break things. With TypeScript, since this could get a value undefined, it needs to be an optional param. Therefore, adding `platform?: string` on the end would be OK. However, I think commander may require the commander.Command argument to be at the end. I don't recall exactly. If it does require the `command` parameter at the end, then there's no way to make it a non-breaking change. However, as I look at this, since it is just the command implementation, and not the actual sideload() function from sideload.ts, I don't think it is as big of a deal to have this change since there's no reason to use this exported function from command.ts. This is a long-winded way to say that I now think it is OK to add this platform parameter here.

TCourtneyOwen:
I'd like to wrap this up today if possible. I think all outstanding issues have been addressed

akrantz (approved):

TCourtneyOwen:
Thanks Adam, I can go ahead and merge if @wandyezj hasn't merged his PR yet. Otherwise I will need to merge his changes first

@wandyezj if I merge this now, please be sure to merge my changes into your PR

PR #363, opened 2021-01-12
Add Edge Chromium documentation to Office-Addin-Dev-Settings

arttarawork:
The actual code to enable webview switching to Edge Chromium already exists, this just adds the documentation to make it easy for devs to look up.

### DIFF
--- a/packages/office-addin-dev-settings/README.md
+++ b/packages/office-addin-dev-settings/README.md
@@ -3,17 +3,20 @@
 Provides the ability to configure developer settings for Office Add-ins.
 
 ## Command-Line Interface
-* [appcontainer](#appcontainer)
-* [clear](#clear)
-* [debugging](#debugging)
-* [live-reload](#live-reload)
-* [register](#register)
-* [registered](#registered)
-* [runtime-log](#runtime-log)
-* [sideload](#sideload)
-* [source-bundle-url](#source-bundle-url)
-* [unregister](#unregister)
-* [webview](#webview)
+- [Office-Addin-Dev-Settings](#office-addin-dev-settings)
+  - [Command-Line Interface](#command-line-interface)
+    - [appcontainer](#appcontainer)
+    - [clear](#clear)
+- [](#)
+    - [debugging](#debugging)
+    - [live-reload](#live-reload)
+    - [register](#register)
+    - [registered](#registered)
+    - [runtime-log](#runtime-log)
+    - [sideload](#sideload)
+    - [source-bundle-url](#source-bundle-url)
+    - [unregister](#unregister)
+    - [webview](#webview)
 
 #
 
@@ -256,6 +259,6 @@ Syntax:
 
 `manifest`: path to manifest file. 
 
-`runtime`: Office runtime to load (currently accepts 'ie', 'edge', or 'default'). 
+`runtime`: Office runtime to load (currently accepts 'ie', 'edge', 'edgechromium', or 'default'). 
 
 #
--- a/packages/office-addin-dev-settings/README.md
+++ b/packages/office-addin-dev-settings/README.md
@@ -252,10 +252,10 @@ Switches the webview runtime in Office for testing and development scenarios.
 
 Syntax:
 
-`office addin-dev-settings webview <manifest> <runtime>`
+`office addin-dev-settings webview <manifest> <web-view-type>`
 
 `manifest`: path to manifest file. 
 
-`runtime`: Office runtime to load (currently accepts 'ie', 'edge', 'edgechromium', or 'default'). 
+`runtime`: Office webview to load (currently accepts 'ie', 'edge' or 'edge-chromium, 'edge-legacy', or 'default'). 
 
 #
--- a/packages/office-addin-dev-settings/README.md
+++ b/packages/office-addin-dev-settings/README.md
@@ -3,20 +3,17 @@
 Provides the ability to configure developer settings for Office Add-ins.
 
 ## Command-Line Interface
-- [Office-Addin-Dev-Settings](#office-addin-dev-settings)
-  - [Command-Line Interface](#command-line-interface)
-    - [appcontainer](#appcontainer)
-    - [clear](#clear)
-- [](#)
-    - [debugging](#debugging)
-    - [live-reload](#live-reload)
-    - [register](#register)
-    - [registered](#registered)
-    - [runtime-log](#runtime-log)
-    - [sideload](#sideload)
-    - [source-bundle-url](#source-bundle-url)
-    - [unregister](#unregister)
-    - [webview](#webview)
+* [appcontainer](#appcontainer)
+* [clear](#clear)
+* [debugging](#debugging)
+* [live-reload](#live-reload)
+* [register](#register)
+* [registered](#registered)
+* [runtime-log](#runtime-log)
+* [sideload](#sideload)
+* [source-bundle-url](#source-bundle-url)
+* [unregister](#unregister)
+* [webview](#webview)
 
 #
 
--- a/packages/office-addin-dev-settings/README.md
+++ b/packages/office-addin-dev-settings/README.md
@@ -256,6 +256,6 @@ Syntax:
 
 `manifest`: path to manifest file. 
 
-`runtime`: Office webview to load (currently accepts 'ie', 'edge' or 'edge-chromium, 'edge-legacy', or 'default'). 
+`runtime`: Office webview to load ('edge' or 'edge-chromium, 'edge-legacy', 'ie', or 'default'). 
 
 #
--- a/packages/office-addin-dev-settings/src/cli.ts
+++ b/packages/office-addin-dev-settings/src/cli.ts
@@ -84,8 +84,8 @@ commander
     .action(commands.webView)
     .on("--help", () => {
         console.log("\nFor [web-view-type], choose one of the following values:\n");
-        console.log("\t'edgechromium' for Microsoft Edge (Chromium)");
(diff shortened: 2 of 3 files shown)


packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -84,7 +84,8 @@ commander
     .action(commands.webView)
     .on("--help", () => {
         console.log("\nFor [web-view-type], choose one of the following values:\n");
-        console.log("\t'edge' for Microsoft Edge");
+        console.log("\t'edgechromium' for the new Chromium-based Microsoft Edge");
```
TCourtneyOwen, line 87: I think leave out "the new" because at some point it won't be new

arttarawork: "newer"? or leave it out completely

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -84,7 +84,8 @@ commander
...
+        console.log("\t'edge' for the older Microsoft Edge");
```
TCourtneyOwen, line 88: leave as is - remove "the older"

packages/office-addin-dev-settings/README.md
```diff
@@ -3,17 +3,20 @@
...
-* [debugging](#debugging)
-* [live-reload](#live-reload)
-* [register](#register)
-* [registered](#registered)
-* [runtime-log](#runtime-log)
-* [sideload](#sideload)
-* [source-bundle-url](#source-bundle-url)
-* [unregister](#unregister)
-* [webview](#webview)
+- [Office-Addin-Dev-Settings](#office-addin-dev-settings)
```
TCourtneyOwen, line 6: Why are we changing the formatting here?

arttarawork: This is weird, I didn't intend to change this. When I tried saving this portion of the readme back to it's original form, it automatically changed it to this format

packages/office-addin-dev-settings/src/cli.ts
akrantz: We should strive to be consistent with the naming. I think we should transition to calling the older version `legacy Microsoft Edge` which is how the https://www.microsoft.com/en-us/edge page puts it. For the new version, I think we should call it `Microsoft Edge (Chromium)`. It is also possible to refer to the legacy version as `Microsoft Edge (EdgeHTML)`.

packages/office-addin-dev-settings/README.md
arttarawork: image

arttarawork: Upon saving:

image

packages/office-addin-dev-settings/src/cli.ts
akrantz: I would recommend to update this to:
```
`edge-legacy` for the legacy Microsoft Edge (EdgeHTML)`
`edge` or `edge-chromium` for Microsoft Edge (Chromium)`
```

I think we could update the code to follow these strings for the versions. Notably, we should switch `edge` to refer to chromium now (it's simpler), and modify `edgechromium` to `edge-chromium` keeping both for backward compatibility but documenting the one with the hyphen.

I'd like your views on this, but I think this would make more sense going forward.

packages/office-addin-dev-settings/src/cli.ts
akrantz: This should match the other documentation in office-addin-dev-settings once that is decided.

packages/office-addin-dev-settings/README.md
akrantz: Is this change intended? I don't see a need to make changes here.

TCourtneyOwen: Can you just revert that change to the formatting and recommit?

packages/office-addin-dev-settings/README.md
```diff
@@ -256,6 +259,6 @@ Syntax:
 
 `manifest`: path to manifest file. 
 
-`runtime`: Office runtime to load (currently accepts 'ie', 'edge', or 'default'). 
+`runtime`: Office runtime to load (currently accepts 'ie', 'edge', 'edgechromium', or 'default').
```
akrantz, line 259: What is 'runtime' used for? This is specifying which webview type, not a runtime really. Can we change this to `<web-view-type>`?

packages/office-addin-dev-settings/README.md
arttarawork: Reverted these changes, but had to use notepad to make them. Weird, I know.

akrantz: You should be able to see these diffs locally even before you push with `git status` and `git difftool`. I like to use `git difftool` to review changes before committing or pushing, as I can make changes and catch stuff. Also, if you edit a file while viewing the diffs, the edits will be applied to the unstaged files, so you can undo specific things easily this way. Glad to show an example some time.

packages/office-addin-dev-settings/src/cli.ts
arttarawork: This should be an easy change, but my immediate concern is it might break some workflows for users. Do we do a minor version bump to account for this, then?

akrantz: It is fair to consider if this will impact users -- do we have any data about how much this command is used? Also, we should see if there is documentation about this and consider that we'll need to update docs to match. This is just for the `edge` tag as the others won't affect anything. We could leave `edge` for the legacy Edge but I really think it makes sense at some point to switch it. The semantic version of the package really won't make a different. We could make a major version bump as a breaking change I don't think this warrants it.  @TCourtneyOwen what's your view? Should we leave `edge` for Edge legacy?  We can still go ahead and add `edge-legacy` and `edge-chromium` and perhaps update docs to refer to them since it is more clear.

packages/office-addin-dev-settings/README.md
```diff
@@ -252,10 +252,10 @@ Switches the webview runtime in Office for testing and development scenarios.
 
 Syntax:
 
-`office addin-dev-settings webview <manifest> <runtime>`
+`office addin-dev-settings webview <manifest> <web-view-type>`
 
 `manifest`: path to manifest file. 
 
-`runtime`: Office runtime to load (currently accepts 'ie', 'edge', 'edgechromium', or 'default'). 
+`runtime`: Office webview to load (currently accepts 'ie', 'edge' or 'edge-chromium, 'edge-legacy', or 'default').
```
akrantz, line 259: I noticed a slight tweak that could be make here, can you remove `currently accepts` so inside parentheses is just:
```
(`edge` or 'edge-chromium', 'edge-legacy`, `ie`, or `default`)
```

Once we decide whether `edge` is for `edge-chromium` or `edge-legacy`, you can put it with the appropriate string.

I think it helps to reorder them from most recent to oldest.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -244,12 +244,15 @@ export function parseWebViewType(webViewString?: string): devSettings.WebViewTyp
     case "internet explorer":
     case "internetexplorer":
       return devSettings.WebViewType.IE;
-    case "edge":
+    case "edgelegacy":
```
akrantz, line 247: These changes look good. Thanks!

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -244,12 +244,15 @@ export function parseWebViewType(webViewString?: string): devSettings.WebViewTyp
...
+    case "edge-legacy":
     case "spartan":
```
TCourtneyOwen, line 249: Do we really want to accept multiple names for a single webview type?

akrantz: It's fine to have the flexibility. We were already doing this.

arttarawork: It's what we've been doing before, and all of these names refer to the same specific webview. I don't see the issue with accepting multiple names, but if there's something that it specifically affects, we can change it.

arttarawork: All these names get converted to the same WebViewType enum, so it gets handled the same way.

packages/office-addin-dev-settings/src/cli.ts
arttarawork: At this point, would the PR be ready for approval? @TCourtneyOwen @akrantz

akrantz: I can review the PR with the assumption we're going to switch `edge` to Edge Chromium, but we'll need to make a call there and then make the minor changes needed if we don't want that. 

@arttarawork It would be great if you could:
1. get data about the usage of the webview command to know if significant.
2. point out the relevant documentation so we can determine if the docs would be affected by this change.

TCourtneyOwen: Let me see if I can quickly get that data and I will send a query url with the results

TCourtneyOwen: It appears that we don't currently collect any usage data for office-addin-dev-settings.  Nor do we collect and related data in offfice-addin-debugging.  We should add a work item to add usage data to office-addin-dev-settings

TCourtneyOwen: As for existing documentation for switching the webview, I couldn't find anything when doing a search, so it's quite likely our ReadMe is our only documentation

TCourtneyOwen (approved):

PR #369, opened 2021-01-19
Putting office-addin-lint usage data changes in separate branch to make integration easier

arttarawork:
…ke integration easier

### DIFF
--- a/packages/office-addin-lint/src/commands.ts
+++ b/packages/office-addin-lint/src/commands.ts
@@ -23,7 +23,11 @@ export async function lint(command: commander.Command) {
     const pathToFiles: string = getPathToFiles(command);
     performLintCheck(pathToFiles);
   } catch (err) {
-    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
+    let outputExitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
+    if (outputExitCode != 1) {
+      logErrorMessage(`Unexpected error detected.\n${err}`);
+    }
+    process.exitCode = outputExitCode;
   }
 }
 
@@ -32,7 +36,11 @@ export async function lintFix(command: commander.Command) {
     const pathToFiles: string = getPathToFiles(command);
     performLintFix(pathToFiles);
   } catch (err) {
-    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
+    let outputExitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
+    if (outputExitCode != 1) {
+      logErrorMessage(`Unexpected error detected.\n${err}`);
+    }
+    process.exitCode = outputExitCode;
   }
 }
 
@@ -41,7 +49,10 @@ export async function prettier(command: commander.Command) {
     const pathToFiles: string = getPathToFiles(command);
     makeFilesPrettier(pathToFiles);
   } catch (err) {
-    logErrorMessage(`Unable to make code prettier.\n${err}`);
-    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.PrettierExitCode.CommandFailed;
+    let outputExitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
+    logErrorMessage(
+      outputExitCode == 1 ? `Unable to make code prettier.\n${err}` : `Unexpected error detected.\n${err}`
+    );
+    process.exitCode = outputExitCode;
   }
 }
--- a/packages/office-addin-lint/src/commands.ts
+++ b/packages/office-addin-lint/src/commands.ts
@@ -23,9 +23,7 @@ export async function lint(command: commander.Command) {
     const pathToFiles: string = getPathToFiles(command);
     performLintCheck(pathToFiles);
   } catch (err) {
-    // no need to display an error since there will already be error output;
-    // just return a non-zero exit code
-    process.exitCode = defaults.ESLintExitCode.CommandFailed;
+    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
   }
 }
 
@@ -34,9 +32,7 @@ export async function lintFix(command: commander.Command) {
     const pathToFiles: string = getPathToFiles(command);
     performLintFix(pathToFiles);
   } catch (err) {
-    // no need to display an error since there will already be error output;
-    // just return a non-zero exit code
-    process.exitCode = defaults.ESLintExitCode.CommandFailed;
+    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
   }
 }
 
@@ -46,6 +42,6 @@ export async function prettier(command: commander.Command) {
     makeFilesPrettier(pathToFiles);
   } catch (err) {
     logErrorMessage(`Unable to make code prettier.\n${err}`);
-    process.exitCode = defaults.PrettierExitCode.CommandFailed;
+    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.PrettierExitCode.CommandFailed;
   }
 }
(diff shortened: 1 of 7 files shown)


packages/office-addin-lint/src/commands.ts
```diff
@@ -27,7 +25,7 @@ export async function lint(command: commander.Command) {
   } catch (err) {
     // no need to display an error since there will already be error output;
     // just return a non-zero exit code
-    process.exitCode = 1;
+    process.exitCode = 2;
```
akrantz, line 28: Same comments as before. Add an enum to capture the exit code value meaning.

packages/office-addin-lint/src/defaults.ts
```diff
@@ -3,11 +3,11 @@
 
 import * as usageData from "office-addin-usage-data";
 
-export const lintFiles = "src/**/*.{ts,tsx,js,jsx}";
+export const lintFiles = "src\\**\\*.{ts,tsx,js,jsx}";
```
akrantz, line 6: Should keep forward slash. This should work on Windows too.

packages/office-addin-lint/src/lint.ts
```diff
@@ -43,13 +47,17 @@ export function getLintFixCommand(files: string): string {
...
 export function performLintFix(files: string) {
-  try{
+  try {
     const command = getLintFixCommand(files);
     execCommand(command);
     usageDataObject.sendUsageDataSuccessEvent("performLintFix");
   } catch (err) {
-    usageDataObject.sendUsageDataException("performLintFix", err);
-    throw err;
+    if (err.status && err.status == 1) {
```
akrantz, line 55: I'd like to understand better why an exception would be thrown if there is not really an error in running the command, but rather it identified a lint error. I am thinking that the command executing successfully should send a success event which includes a data point for the status so that it can report success for "no errors" (e.g. 0) and success for "contains lint errors" (whichever status code). I think if the data point is the status code or a string corresponding to the status code, then we can query it on the other side. What is less than clear is whether we are getting an exception if the command runs OK but there are lint errors that cannot be fixed.

packages/office-addin-lint/src/lint.ts
```diff
@@ -59,12 +67,16 @@ export function getPrettierCommand(files: string): string {
 }
 
 export function makeFilesPrettier(files: string) {
-  try{
+  try {
     const command = getPrettierCommand(files);
     execCommand(command);
     usageDataObject.sendUsageDataSuccessEvent("makeFilesPrettier");
```
akrantz, line 73: Same goes here. Need to better clarify when the prettier command succeed or has an exception, and include the status code with the success event. (If an exception is thrown, I would suspect the status code wouldn't apply and the err would contain the info.)

akrantz (changes requested):
Same feedback as before for the new PR.

packages/office-addin-lint/src/lint.ts
arttarawork: With the way I have written it, we send a success event in the event that there are linting errors, so I'm not sure what the issue is specifically.

packages/office-addin-lint/src/defaults.ts
arttarawork: Will change in latest commit now that forward slash is working on my windows machine.

packages/office-addin-lint/src/lint.ts
```diff
@@ -32,8 +32,12 @@ export function performLintCheck(files: string) {
     execCommand(command);
     usageDataObject.sendUsageDataSuccessEvent("performLintCheck");
   } catch (err) {
-    usageDataObject.sendUsageDataException("performLintCheck", err);
-    throw err;
+    if (err.status && err.status == ESLintExitCode.LintingError) {
```
TCourtneyOwen, line 35: why are we sending linting errors to usage data. Perhaps I am misunderstanding, but we don't really care about users' linting error and don't want to muddy up our own usage data with this type of data.  If I am misunderstanding, it's because there's no description for this PR. Please be sure to include this for a PR :-)

TCourtneyOwen (approved):

packages/office-addin-lint/src/defaults.ts
```diff
@@ -5,9 +5,21 @@ import * as usageData from "office-addin-usage-data";
 
 export const lintFiles = "src/**/*.{ts,tsx,js,jsx}";
 
+export enum ESLintExitCode {
+  Success = 0,
+  LintingError = 1,
```
akrantz, line 10: Based on https://eslint.org/docs/user-guide/command-line-interface#exit-codes

I think these values may be clearer. The first two are both "success" cases but I think it's better to focus on the fact that it indicates whether or not there were lint errors reported. For the third case, the command itself failed.

```
NoLintErrors = 0,
HasLintError = 1,
CommandFailed = 2
```

packages/office-addin-lint/src/defaults.ts
```diff
@@ -5,9 +5,21 @@ import * as usageData from "office-addin-usage-data";
...
+  ToolingError = 2
+}
+
+export enum PrettierExitCode {
```
akrantz, line 14: Similarly for Prettier: https://prettier.io/docs/en/cli.html#exit-codes

```
NoFormattingProblems = 0,
HasFormattingProblem = 1,
CommandFailed = 2
```

packages/office-addin-lint/src/lint.ts
akrantz: We want to report a successful command execution when there is a non-zero exit code (which results in an exception being thrown), and report the exit code, but report a failure when the status code indicates the command failed (which is a value of 2).

akrantz (changes requested):
Looks good. I suggest making the enum names clearer.

packages/office-addin-lint/src/commands.ts
```diff
@@ -25,7 +25,7 @@ export async function lint(command: commander.Command) {
   } catch (err) {
     // no need to display an error since there will already be error output;
     // just return a non-zero exit code
-    process.exitCode = defaults.ESLintExitCode.ToolingError;
+    process.exitCode = defaults.ESLintExitCode.CommandFailed;
```
akrantz, line 28: shouldn't this exit code reflect the command exit code?  Perhaps these should all just be `process.exitCode = err && typeof(err.status) == "number" ? err.status : ESLintExitCode.CommandFailed`. 

This is just that if you use the command and there are lint errors, the exit code should be 1 to reflect that.

arttarawork: Exit code issue after manual testing has been fixed 👍

packages/office-addin-lint/src/lint.ts
```diff
@@ -34,6 +34,7 @@ export function performLintCheck(files: string) {
   } catch (err) {
     if (err.status && err.status == ESLintExitCode.HasLintError) {
       usageDataObject.sendUsageDataSuccessEvent("performLintCheck", { exitCode: ESLintExitCode.HasLintError });
+      throw err;
```
akrantz, line 37: This makes sense. If you notice now, both the if and the else throw err. You can simplify this to always throw error after the if/else. The point here is that the exception is caught in order to send the usage data, but needs to be rethrown so the error is displayed by the top-level catch for the command.

akrantz: (This applies for of the places.)

packages/office-addin-lint/src/commands.ts
```diff
@@ -23,7 +23,11 @@ export async function lint(command: commander.Command) {
     const pathToFiles: string = getPathToFiles(command);
     performLintCheck(pathToFiles);
   } catch (err) {
-    process.exitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
+    let outputExitCode = err && typeof err.status == "number" ? err.status : defaults.ESLintExitCode.CommandFailed;
```
akrantz, line 26: I think you could change this a better way. I don't think adding the `Unexpected error detected.` message helps (but I might need to see it to make sure I understand what the output is when the command has an unexpected error).  The key thing here is that after catching it so the exit code can be set, the error should be retrhown so the error message is shown (by the top-level catch handler).

akrantz (changes requested):
Some comments on these changes.

akrantz (approved):
Looks good. Thanks for taking the time to go through all of these cases and get it working well.

TCourtneyOwen:
Btw, I am not sure why the Mac job is not kicking off in the CI loop

PR #371, opened 2021-01-25
Fix bug where Outlook exe path has a space in the path

TCourtneyOwen:
- This addresses the scenario where the Outlook exe path that is to be lauched has a space in it.

### DIFF
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -219,7 +219,11 @@ async function getOutlookExePath(): Promise<string | undefined> {
   try {
     const OutlookInstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\OUTLOOK.EXE`;
     const key = new registry.RegistryKey(`${OutlookInstallPathRegistryKey}`);
-    const outlookExePath: string | undefined = await registry.getStringValue(key, "");
+    let outlookExePath: string | undefined = await registry.getStringValue(key, "");
+    
+    if (outlookExePath) {
+      outlookExePath = `"${outlookExePath}"`;
+    }
 
     return outlookExePath;
   } catch (err) {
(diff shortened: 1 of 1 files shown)


packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -219,7 +219,11 @@ async function getOutlookExePath(): Promise<string | undefined> {
   try {
     const OutlookInstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\OUTLOOK.EXE`;
     const key = new registry.RegistryKey(`${OutlookInstallPathRegistryKey}`);
-    const outlookExePath: string | undefined = await registry.getStringValue(key, "");
+    let outlookExePath: string | undefined = await registry.getStringValue(key, "");
+    
+    if (outlookExePath) {
+      outlookExePath = `"${outlookExePath}"`;
```
akrantz, line 225: Personally, rather than putting the string with spaces in a quote here, I would do this differently. Just return the string for the path, which can have spaces, and at the place where it is used as an argument, add the quotes around the path so it handles the command line appropriately.

akrantz: That would be in the `startDetachedProcess(sideloadFile)` on line 337. You could either always escape the argument there, or see if the implementation should have special code to check if on Windows and file has a space and then add the quotes around it. Or maybe a different command should be used to launch the process, since this doesn't need to run via cmd.exe.

TCourtneyOwen:
I don't want to change the implementation of startDetachedProcess too much, as it's also used to start the dev-server. I can add the space check there but would prefer not to create a new method just to allow for launching the Outlook.exe

akrantz:
That why I suggest to just change this line in sideload.ts for the Outlook case:

 `startDetachedProcess(`"${sideloadFile}"`)`

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -334,6 +330,10 @@ export async function sideloadAddIn(manifestPath: string, appType: AppType, app?
 
   if (sideloadFile) {
     if (app == OfficeApp.Outlook) {
+      // check to see if Outlook exe path contains spaces and escape the path if it does.
+      if (sideloadFile.indexOf(' ') >= 0){
+        sideloadFile = `"${sideloadFile}"`;
```
akrantz, line 335: Do you have to check if there's a space in order to add double-quotes?  This is fine, but just wondering.

TCourtneyOwen: yes that's correct

akrantz (approved):

Issue #352, opened 2020-12-10
office-addin-manifest validate command should give an error if manifest path doesn't point to a file.


afisher:
# Expected behavior

Running `office-addin-manifest validate manifest` outputs the result of running against the validator.

# Current behavior

Running `office-addin-manifest validate manifest` outputs an error saying the service could not be contacted.

## Steps to Reproduce

`office-addin-manifest validate manifest`

Output is:
```
Error: Unable to contact the manifest validation service.
FetchError: request to https://validationgateway.omex.office.net/package/api/check failed, reason: socket hang up
```

## Context

* Operating System: Mac OS Catalina 10.15.7 
* Node version: v14.14.0
* Tool version: office-addin-manifest@1.5.12

afisher:
My mistake, the path to the manifest file was wrong, it should be `manifest.xml` in the command.

akrantz:
It would be useful for the command to detect that the manifest path does not correspond to a file and generate a good error message.

TCourtneyOwen:
I will create a bug for this

akrantz:
Why was this closed?

millerds:
Courtney merged a PR for this almost 10 months ago.

akrantz:
OK, great.

PR #372, opened 2021-01-27
validate command should throw error if manifest path isn't valid

TCourtneyOwen:
- Call readManifestFile to ensure manifestPath is valid before calling validateManifest. Catch block will catch error if manifest file path is not a valid file and return error to use (e.g. rror: ENOENT: no such file or directory, open 'C:\Office-Addin-Scripts\packages\office-addin-manifest\foo.xml')
- Fixes issue https://github.com/OfficeDev/Office-Addin-Scripts/issues/352

### DIFF
--- a/packages/office-addin-manifest/src/commands.ts
+++ b/packages/office-addin-manifest/src/commands.ts
@@ -129,28 +129,31 @@ export async function modify(manifestPath: string, command: commander.Command) {
 
 export async function validate(manifestPath: string, command: commander.Command) {
   try {
-    const validation: ManifestValidation = await validateManifest(manifestPath);
-
-    if (validation.isValid) {
-      console.log("The manifest is valid.");
-    } else {
-      console.log("The manifest is not valid.");
-    }
-    console.log();
-
-    if (validation.report) {
-      logManifestValidationErrors(validation.report.errors);
-      logManifestValidationWarnings(validation.report.warnings);
-      logManifestValidationInfos(validation.report.notes);
+    const manifest: manifestInfo.ManifestInfo = await manifestInfo.readManifestFile(manifestPath);
+    if (manifest) {
+      const validation: ManifestValidation = await validateManifest(manifestPath);
 
       if (validation.isValid) {
-        if (validation.report.addInDetails) {
-          logManifestValidationSupportedProducts(validation.report.addInDetails.supportedProducts);
+        console.log("The manifest is valid.");
+      } else {
+        console.log("The manifest is not valid.");
+      }
+      console.log();
+  
+      if (validation.report) {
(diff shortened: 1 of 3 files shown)


packages/office-addin-manifest/src/commands.ts
```diff
@@ -129,28 +129,31 @@ export async function modify(manifestPath: string, command: commander.Command) {
...
-    } else {
-      console.log("The manifest is not valid.");
-    }
-    console.log();
-
-    if (validation.report) {
-      logManifestValidationErrors(validation.report.errors);
-      logManifestValidationWarnings(validation.report.warnings);
-      logManifestValidationInfos(validation.report.notes);
+    const manifest: manifestInfo.ManifestInfo = await manifestInfo.readManifestFile(manifestPath);
```
akrantz, line 132: In validate.ts, there is a readManifestFile() which is commented out. Seems better to uncomment it there.

akrantz (changes requested):

packages/office-addin-manifest/src/validate.ts
```diff
@@ -65,47 +65,51 @@ export async function validateManifest(manifestPath: string): Promise<ManifestVa
...
+    
+                if (result) {
+                    switch (result.toLowerCase()) {
+                        case "accepted":
+                            validation.isValid = true;
+                            break;
+                    }
                 }
+            } else {
+                throw new Error("The manifest validation service did not return the expected response.");
```
akrantz, line 107: I don't think this else can even happen since readManifestFile() should return the manifest info or throw.

You don't actually need the return value from readManifestFile() at all.

Based on the original code where the line was commented out, you should just be able to uncomment the line and remove the variable assignment.

packages/office-addin-manifest/src/validate.ts
```diff
@@ -3,7 +3,7 @@
 
 import { createReadStream } from "fs";
 import fetch from "node-fetch";
-import { readManifestFile } from "./manifestInfo";
+import { ManifestInfo, readManifestFile } from "./manifestInfo";
```
akrantz, line 6: can remove ManifestInfo import.

packages/office-addin-manifest/src/validate.ts
```diff
@@ -101,10 +102,8 @@ export async function validateManifest(manifestPath: string): Promise<ManifestVa
                         break;
                 }
             }
-        } else {
-            throw new Error("The manifest validation service did not return the expected response.");
         }
-
+
```
akrantz, line 106: no whitespace change needed here.

akrantz (approved):
Looks good. There are a couple of suggestions of minor fixes that would be good to make.

akrantz (approved):

PR #374, opened 2021-01-29
fix startDebugging defaults when undefined is passed in

wandyezj:


TCourtneyOwen:
This works Joe, thank for getting the fix in.  I have a couple changes that I would like to make that Adam and I discussed, the most notable one being to remove manifestPath from the StartDebuggingOptions, as it's not really an option - it's required.  I have that change locally and will push it up to your branch

packages/office-addin-debugging/src/start.ts
```diff
@@ -385,6 +385,18 @@ export async function startDebugging(options: StartDebuggingOptions) {
     }
 }
 
+/**
+ * @param option 
+ * @param other 
+ * @returns other if option is undefined or option
+ */
+function definedOrOther<T>(option:T | undefined, other: T) {
```
akrantz, line 393: This function is not really needed. You could just as easily write:

```
debuggingMethod: options.DebuggingMethod || defaultDebuggingMethod(),
enableDebugging: options.EnableDebugging || true,
```

akrantz (changes requested):
This code can be simplified to not require an extra method. See comment.

packages/office-addin-debugging/src/start.ts
TCourtneyOwen: /Azurepipelines run

akrantz:
These changes have already been incorporated.

packages/office-addin-debugging/src/start.ts
wandyezj: @akrantz 

In this case perhaps, but it's easy to duplicate and to then encounter an edge case that does not perform as expected:

```typescript
const va = 0 || "hello"
console.log(va) // hello

const vb = false || "hello"
console.log(vb) // hello
```

akrantz: well really the point is to check whether undefined or not, rather than whether falsey or not. Personally the function didn't make that clear. Perhaps it was the name. 

Here's the best way in TypeScript:

`const value = options.DebuggingMethod ?? true;`

PR #375, opened 2021-01-29
Make app type unrequired for sideload

TCourtneyOwen:
Make appType param not required for sideload command
    
- Update sideload cli to reflect type is once again an option, not required 
- Update sideloadAddin to make appType optional and revert param ordering for method back to previous method signature so this is not a breaking change
- Update Readme to reflect that type is now an option
- Remove check for Outlook sideloading environment variable so we can enable Outlook sideloading (we don't have to include this in this PR but I thought it would be nice to get this enabled soon)
- Incorporate changes from StartDebuggingOptions to fix issue with default options not being set properly. Also remove manifestPath from StartDebuggingOptions as it's not really an option - it's required

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -271,12 +271,12 @@ export async function startDebugging(options: StartDebuggingOptions) {
         openDevTools,
         document,
     } = {
-        // Defaults
-        debuggingMethod: defaultDebuggingMethod(),
-        enableDebugging: true,
+        // Supplied Options
+        ...options,
 
-        // Override with supplied options
-        ...options
+        // Defaults when variable is undefined.
+        debuggingMethod: definedOrOther(options.debuggingMethod, defaultDebuggingMethod()),
+        enableDebugging: definedOrOther(options.enableDebugging, true),
     };
 
     try {
@@ -383,6 +383,18 @@ export async function startDebugging(options: StartDebuggingOptions) {
         usageDataObject.sendUsageDataException("startDebugging", err);
         throw err;
     }
+}
+
+/**
+ * @param option 
+ * @param other 
+ * @returns other if option is undefined or option
+ */
+function definedOrOther<T>(option:T | undefined, other: T) {
+    if (option === undefined) {
+        return other;
+    }
+    return option;
 }
 
 export async function waitUntil(callback: (() => Promise<boolean>), retryCount: number, retryDelay: number): Promise<boolean> {
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -178,11 +178,6 @@ export async function runPackager(commandLine: string, host: string = "localhost
 export interface StartDebuggingOptions {
 
     /**
-     * The path to the manifest file.
-     */
-    manifestPath: string,
-
-    /**
      * The type of application to debug.
      */
     appType: AppType,
@@ -254,9 +249,8 @@ export interface StartDebuggingOptions {
  * Start debugging
  * @param options startDebugging options.
  */
-export async function startDebugging(options: StartDebuggingOptions) {
+export async function startDebugging(manifestPath: string, options: StartDebuggingOptions) {
     const {
-        manifestPath,
         appType,
         app,
         debuggingMethod,
@@ -294,7 +288,7 @@ export async function startDebugging(options: StartDebuggingOptions) {
         console.log(enableDebugging
             ? "Debugging is being started..."
             : "Starting without debugging...");
-        console.log(`App type: ${appType.toString()}`);
+        console.log(`App type: ${appType}`);
 
         const manifestInfo = await readManifestFile(manifestPath);
 
@@ -312,7 +306,7 @@ export async function startDebugging(options: StartDebuggingOptions) {
         if (isDesktopAppType && isWindowsPlatform) {
             await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);
             if (enableDebugging) {
-                console.log(`Enabled debugging for add-in ${manifestInfo.id}. Debug method: ${debuggingMethod.toString()}`);
+                console.log(`Enabled debugging for add-in ${manifestInfo.id}. Debug method: ${DebuggingMethod[debuggingMethod]}`);
             }
         }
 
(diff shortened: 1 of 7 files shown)


akrantz (changes requested):

packages/office-addin-debugging/src/start.ts
```diff
@@ -294,7 +288,7 @@ export async function startDebugging(options: StartDebuggingOptions) {
         console.log(enableDebugging
             ? "Debugging is being started..."
             : "Starting without debugging...");
-        console.log(`App type: ${appType.toString()}`);
+        console.log(`App type: ${appType}`);
```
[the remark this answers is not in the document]
TCourtneyOwen: doing a check now for appType to see if undefined

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -57,16 +57,12 @@ commander
     .action(commands.runtimeLogging);
 
 commander
-    .command("sideload <manifest-path> <app-type.")
+    .command("sideload <manifest-path>")
```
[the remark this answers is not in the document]
akrantz: Is there a commit you haven't pushed yet?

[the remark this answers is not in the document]
TCourtneyOwen: Just did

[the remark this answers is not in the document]
akrantz: OK, I'm not seeing the changes related to my original comment here.

akrantz (changes requested):
good progress but still waiting on some remaining changes to be made.

packages/office-addin-dev-settings/src/cli.ts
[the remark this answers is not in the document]
TCourtneyOwen: Ugh forgot to add those to the CL

packages/office-addin-dev-settings/src/cli.ts
```diff
@@ -56,13 +56,17 @@ commander
...
+    commander
+    .command("sideload <manifest-path> [app-type]")
     .description("Launch Office with the Office Add-in loaded.")
-    .option("-a,--app <app>", `The Office app to launch. ("Excel", "Outlook", "PowerPoint", or "Word")`)    
+    .option("-a,--app <app>", `The Office app to launch. ("Excel", "Outlook", "PowerPoint", or "Word")`)
     .option("-d,--document <document>", `The location of the document to be sideloaded - this can be an absolute file path or url`)
-    .option("-t, --type <type>", `Application type to sideload ("desktop" or "web")`)
-    .action(commands.sideload);
+    .action(commands.sideload)
+    .on("--help", () => {
```
akrantz, line 65: I would revise the help text. It is not really a "desktop add-in" or a "web add-in". It is whether to load in the Office desktop app or in Office Online (web).

My suggestion:
```
[app-type] specifies the type of Office app:
   'desktop': Office app for Windows or Mac (default),
   'web': Office running in the web browser
 ```

File: packages/office-addin-dev-settings/src/commands.ts [hunk 4]
```diff
@@ -233,6 +233,19 @@ export async function liveReload(manifestPath: string, command: commander.Comman
   }
 }
 
+function parseAppType(appType: string | undefined): AppType | undefined {
```
akrantz, line 236: Don't we already have this function somewhere? Let's try to use one copy.

akrantz (changes requested):

packages/office-addin-dev-settings/src/commands.ts
TCourtneyOwen: If we use the one in office-addin-debugging, I think I will need to move it to office-addin-dev-settings to avoid a circular dependency.  That's fine with me as long as we are in agreement to move it to office-addin-dev-settings

akrantz: I think it makes sense to do that. Thanks.

TCourtneyOwen: Moving parseAppType to sideload.ts in office-addin-dev-settings.  I think it makes sense to put it in sideload.ts since this is where we also define the AppType enum

packages/office-addin-dev-settings/src/cli.ts
TCourtneyOwen: updated locally

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -28,6 +28,28 @@ export enum AppType {
...
 
+/**
+ * Parse the input text and get the associated AppType
+ * @param text app-type/platform text
+ * @returns AppType or undefined.
+ */
+export function parseAppType(text: string | undefined): AppType | undefined {
+  switch (text ? text.toLowerCase() : undefined) {
+    case "desktop":
+    case "macos":
```
akrantz, line 39: this should only have "desktop" and "web" and not the other platforms (but keep undefined case). For the record there is a separate function for parsing the platforms and you can map from platform to app type. office-addin-debugging start does this, but it's not needed for sideload.

akrantz (changes requested):
Looks good. See the one comment I made.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: Wait I am confused - do we not need all the stuff anymore about macOs, win32, ios and android? Are these effectively no longer used anywhere. If I remove them from here, will that cause problems for the office-addin-debugging call to parseAppType in packages/office-addin-debugging/src/commands.ts?

akrantz: If you look at this diff, the code you're moving doesn't have those other cases. Those cases are part of parsePlatform() in office-addin-debugging. We don't need all of the platforms here and we could add support for it later if for some reason we did, e.g. sideload to android.

akrantz: actually I think I am mistaken.

akrantz: This is the code I was thinking of: https://github.com/OfficeDev/Office-Addin-Scripts/blob/755485666db67538ed802277efad751f1c01565e/packages/office-addin-debugging/src/commands.ts#L43

However, you're right that this function did have the platform strings. Sorry for the confusion.

akrantz (approved):

PR #376, opened 2021-02-01
Fix issue with granting admin consent

TCourtneyOwen:
- Azure seems to have recently changed the naming convention for a tenant admin from "Company Administrator" to "Global Administrator" and as a result granting admin consent was failing for the SSO application created via configure-sso
- I believe this should fix https://github.com/OfficeDev/Office-Addin-Taskpane-SSO/issues/25


akrantz (approved):
I'm wondering why there is an "unresolved promise rejection" error showing when this error gets thrown. That usually means a missing await somewhere, but I didn't see one missing. I don't think that should hold up this fix but just thought I would mention it.

PR #379, opened 2021-02-05
Removing Debug Method console logging

igor-ribeiiro:
Removed the following code:

 // enable debugging
        if (isDesktopAppType && isWindowsPlatform) {
            await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);
            if (enableDebugging) {
                console.log(`Enabled debugging for add-in ${manifestInfo.id}. Debug method: ${DebuggingMethod[debuggingMethod]}`);
            }
        }

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -293,6 +293,9 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
         // enable debugging
         if (isDesktopAppType && isWindowsPlatform) {
             await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);
+            if (enableDebugging) {
+                console.log(`Enabled debugging for add-in ${manifestInfo.id}.`);
+            }
         }
 
         // enable live reload
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -290,14 +290,6 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
             await devSettings.ensureLoopbackIsEnabled(name);
         }
 
-        // enable debugging
-        if (isDesktopAppType && isWindowsPlatform) {
-            await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);
-            if (enableDebugging) {
-                console.log(`Enabled debugging for add-in ${manifestInfo.id}. Debug method: ${DebuggingMethod[debuggingMethod]}`);
-            }
-        }
-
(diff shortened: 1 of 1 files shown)


packages/office-addin-debugging/src/start.ts
```diff
@@ -290,14 +290,6 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
             await devSettings.ensureLoopbackIsEnabled(name);
         }
 
-        // enable debugging
-        if (isDesktopAppType && isWindowsPlatform) {
-            await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);
```
TCourtneyOwen, line 295: This should remain:

await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);

Please just remove the logging portion below

packages/office-addin-debugging/src/start.ts
```diff
@@ -290,14 +290,6 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
             await devSettings.ensureLoopbackIsEnabled(name);
         }
 
-        // enable debugging
```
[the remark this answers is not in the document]
akrantz: The reason that we're removing the logging of the debugging method is that the proxy debugging method that related to ReactNative is no longer being used, as ReactNative is moving towards direct debugging, so we really only have the one debugging method, so there is no need to log that.

akrantz (changes requested):
Thanks for the pull request! See the comments.

packages/office-addin-debugging/src/start.ts
akrantz: @TCourtneyOwen See my comments. We want the console.log() line, but just not the "Debug method: {method}" part.

packages/office-addin-debugging/src/start.ts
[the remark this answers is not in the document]
TCourtneyOwen: /Azurepipelines run

packages/office-addin-debugging/src/start.ts
```diff
@@ -293,9 +293,6 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
         // enable debugging
         if (isDesktopAppType && isWindowsPlatform) {
             await devSettings.enableDebugging(manifestInfo.id, enableDebugging, debuggingMethod, openDevTools);
-            if (enableDebugging) {
```
akrantz, line 296: Sorry if this has been unclear but we want this console.log message. Just remove the "Debug method: " part at the end.

akrantz (approved):

PR #381, opened 2021-02-06
improve npm scripts

akrantz:
Create "exec" script for running lerna exec.
Update other scripts to use "npm run exec".

Add scripts "install-packages" and "clean-install-packages" which run "npm install" or "npm ci", respectively, and then run "npm install" in the root afterward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "postinstall": "lerna bootstrap",
     "publish:next": "lerna publish --dist-tag next",
     "test": "lerna run test",
-    "update-latest-dist-tags": "npm run exec for /f \"usebackq\" %a in (`npm view . name`) do echo @for /f \"usebackq\" %b in (`npm view . dist-tags.next`) do @npm dist-tag add %a@%b",
+    "update-latest-dist-tags": "npm run exec for /f \"usebackq\" %a in (`npm view . name`) do @for /f \"usebackq\" %b in (`npm view . dist-tags.next`) do @npm dist-tag add %a@%b",
     "update-packages": "npm run exec -- npm update --dev"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,15 +2,18 @@
...
+    "clean-install-packages": "npm run exec -- npm ci & npm ci",
+    "dist-tags": "npm run exec -- npm dist-tag ls",
+    "install-packages": "npm run exec -- npm install & npm install",
+    "exec": "lerna exec --concurrency 1 --no-bail --no-sort --stream",
     "postinstall": "lerna bootstrap",
     "publish:next": "lerna publish --dist-tag next",
     "test": "lerna run test",
-    "update-latest-dist-tags": "lerna exec --concurrency 1 --no-sort --stream  for /f \"usebackq\" %a in (`npm view . name`) do @for /f \"usebackq\" %b in (`npm view . dist-tags.next`) do @npm dist-tag add %a@%b",
-    "update-packages": "lerna exec --concurrency 1 --no-sort --stream npm update --dev"
+    "update-latest-dist-tags": "npm run exec for /f \"usebackq\" %a in (`npm view . name`) do echo @for /f \"usebackq\" %b in (`npm view . dist-tags.next`) do @npm dist-tag add %a@%b",
```
akrantz, line 15: I added the `echo` for testing. Need to remove it.

TCourtneyOwen: Curious, what's the improvement we get from these changes?

akrantz: I added scripts "install-packages" and "clean-install-packages". These run "npm install" or "npm ci", respectively, then "npm install" in the root.

There are a bunch of switches for "lerna exec" that we now only have in one place for the "exec" script, and you can use that yourself for running other commands across the packages.

TCourtneyOwen (approved):

PR #385, opened 2021-02-09
Bugfix: Spaces in filename in lint commands

igor-ribeiiro:
**Summary**
The function that executes the commands can't handle spaces in files paths naturally:
[https://stackoverflow.com/questions/16395612/unable-to-execute-child-process-exec-when-path-has-spaces](https://stackoverflow.com/questions/16395612/unable-to-execute-child-process-exec-when-path-has-spaces)

So the solution is simply add a regex that escapes the spaces:
```replace(/ /g, '\\ ')```

**Unit Test:** 
created a new unit test to test file paths with spaces in it. The new unit test passes with my changes on the code and fails with them, which makes sense, as the code didn't work for spaces before. 

**End to End Test**
Ran npm link into two other projects: one with spaces in the name and the other without it.
Ran the following commands on both projects with success: 
- ```npm run lint```
- ```npm run lint:fix```
- ``` npm run prettier```

### CODE REVIEW

File: packages/office-addin-lint/src/lint.ts [hunk 1]
```diff
@@ -22,7 +22,7 @@ function getEsLintBaseCommand(): string {
 }
 
 export function getLintCheckCommand(files: string): string {
-  const eslintCommand: string = `${getEsLintBaseCommand()} ${files}`;
+  const eslintCommand: string = `${getEsLintBaseCommand()} ${files.replace(/ /g, '\\ ')}`; // Converting space to '\\'
   return eslintCommand;
```
TCourtneyOwen, line 26: Can we create a common function to convert the file path ? Something like normalFilePath(filePath: string) : string

packages/office-addin-lint/test/src/test.ts
```diff
@@ -13,10 +13,29 @@ describe("test cases", function() {
...
+
+        const lintFixCommand = prettierLint.getLintFixCommand(inputFile);
+        assert.strictEqual(lintFixCommand.indexOf(lintFixExpectedCommand) > 0, true, "Lint fix command does not match expected value.");
+
+        const prettierCommand = prettierLint.getPrettierCommand(inputFile);
+        assert.strictEqual(prettierCommand.indexOf(prettierExpectedCommand) > 0, true, "Prettier command does not match expected value.");
+    });
+
+    it("spaces in filename", async function() {
+        const inputFile = "./test/cases/basic/functions with space.ts";
```
TCourtneyOwen, line 28: Did we also verify for spaces in file path (e.g. "/test/cases/**basic cases**/functions with space.ts"?

TCourtneyOwen (approved):

akrantz:
Was this tested on both Windows and Mac?

TCourtneyOwen:
I can try it out on Mac

TCourtneyOwen:
On Mac, I don't hit the initial error we were seeing because the project name has spaces in it, but I am seeing another error that says "No files matching the pattern "src/**/*.tsx were found".  I need to investigate a bit more and see if this error also occurs in the case where the project doesn't have spaces

TCourtneyOwen:
The issue I am seeing on Mac where it says "No files matching the pattern "src/**/*.tsx were found" happens independent of Igor's changes so presumably linting has not worked properly on Mac all along

TCourtneyOwen:
If I update our listing script in Office-Add-Taskpane, for example, to the following, it works on Mac:

"lint": "office-addin-lint check --files \"src/**/*.ts\"",

So it seems to me that the problem lies in reading the default value of "src/**/*.{ts,tsx,js,jsx}"

I think it's the "{. }" around the extensions is causing problems

TCourtneyOwen:
I think we should merge your change and open a separate bug for the Mac issue.  You's fixes things on Windows and it looks like the Mac issue has always existed

TCourtneyOwen:
I opened a bug to address the linting issue on Mac, which was also identified by a customer at https://github.com/OfficeDev/generator-office/issues/564

igor-ribeiiro:
Makes sense, will merge the changes

PR #387, opened 2021-02-12
Adding Telemetry to Office-Addin-Dev-Settings

igor-ribeiiro:
Added functions to send data usage success and failure events to office-addin-dev-settings for all the commands.

Tests:
- success cases: just by running the command and then checking if the event was logged.
- fail cases: created a throw error at the beginning of each command then called them and checked the event was logged
- npm run tests worked successfully


packages/office-addin-dev-settings/package-lock.json
```diff
@@ -1,2810 +1,8 @@
 {
   "name": "office-addin-dev-settings",
   "version": "1.8.16",
-  "lockfileVersion": 2,
```
akrantz, line 4: Please use the latest npm. `npm install -g npm`. npm v7 now writes lock file version 2.

akrantz (changes requested):
This looks good. You need to update to NPM v7.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -23,6 +23,19 @@ export enum UsageDataLevel {
...
+/**
+ * Defines an error that is expected to happen given some situation
+ * @member err Message to be logged in the error
+ */
+export class ExpectedError extends Error {
+  constructor(err: string) {
+      super(err);
+
+      // Set the prototype explicitly.
+      Object.setPrototypeOf(this, ExpectedError.prototype);
```
akrantz, line 35: I'm not familiar with this but I wouldn't think this is the way to do this in TypeScript.

akrantz: The Error() constructor type definition is `Error(message: string | undefined)` so the constructor here should match that.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -26,28 +27,32 @@ export async function appcontainer(manifestPath: string, command: commander.Comm
           const allowed = await ensureLoopbackIsEnabled(manifestPath, askForConfirmation);
           console.log(allowed ? "Loopback is allowed." : "Loopback is not allowed.");
         } catch (err) {
-          throw new Error(`Unable to allow loopback for the appcontainer. \n${err}`);
+          throw new ExpectedError(`Unable to allow loopback for the appcontainer. \n${err}`);
```
akrantz, line 30: This is not an expected error.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -26,28 +27,32 @@ export async function appcontainer(manifestPath: string, command: commander.Comm
...
         }
       } else if (command.preventLoopback) {
         try {
           const name = await getAppcontainerNameFromManifestPath(manifestPath);
           await removeLoopbackExemptionForAppcontainer(name);
           console.log(`Loopback is no longer allowed.`);
         } catch (err) {
-          throw new Error(`Unable to disallow loopback. \n${err}`);
+          throw new ExpectedError(`Unable to disallow loopback. \n${err}`);
```
akrantz, line 38: This is not an expected error.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -26,28 +27,32 @@ export async function appcontainer(manifestPath: string, command: commander.Comm
...
         }
       } else {
         try {
           const name = await getAppcontainerNameFromManifestPath(manifestPath);
           const allowed = await isLoopbackExemptionForAppcontainer(name);
           console.log(allowed ? "Loopback is allowed." : "Loopback is not allowed.");
         } catch (err) {
-          throw new Error(`Unable to determine if appcontainer allows loopback. \n${err}`);
+          throw new ExpectedError(`Unable to determine if appcontainer allows loopback. \n${err}`);
```
akrantz, line 46: Not an expected error.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -26,28 +27,32 @@ export async function appcontainer(manifestPath: string, command: commander.Comm
...
         }
       }
-      usageDataObject.sendUsageDataSuccessEvent("appcontainer");
+      usageDataObject.reportSuccess("appcontainer");
     } catch (err) {
-      usageDataObject.sendUsageDataException("appcontainer", err);
+      if (err instanceof ExpectedError){
+        usageDataObject.reportExpectedError("appcontainer", err);
```
akrantz, line 52: perhaps we should change usageDataObject to be:

usageDataObject.reportError("appcontainer", err) with an optional isExpected parameter which could be explicitly set or would be implicity set based on whether it is ExpectedError. So there could be different usage. Here it can just pass err but for other usage it could be:

// this is an expected error
usageDataObject.reportError("eventName", err, true);

akrantz (changes requested):
See comments.

packages/office-addin-dev-settings/src/commands.ts
igor-ribeiiro: I've set the function name to be reportException, because a reportError function already existed and I didn't want to change its name

akrantz: Let's chat. There are a few things we should clarify. You're doing great. I am thinking that we need to understand better how the underlying telemetry data can distinguish the three cases (success, expected error, unexpected error). I think we may want to set Succeeded = true even for expected errors, so this is only false for unexpected errors. But then for all successful telemetry (when Succeeded == true), how would we tell between success and expected errors. Do we have ExpectedError = true? Perhaps.

I know currently you have Succeeded = true for success, and false for both expected and unexpected errors, and then Expected = true for expected errors. 

The only real reason to change this is that it aligns more with how other telemetry is done.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -290,45 +327,71 @@ export class OfficeAddinUsageData {
...
 
   /**
-   * Reports custom event object to Application Insights
+   * Reports custom exception event object to Application Insights
    * @param projectName Project name sent to Application Insights
+   * @param err Error or message about error sent to Application Insights
    * @param data Data object(s) sent to Application Insights
+   * @deprecated Use `reportUnexpectedError` instead.  
    */
-  public sendUsageDataEvent(data: object = {}) {
```
akrantz, line 308: I think this function should remain as the basic function for sending the telemetry event. The other reportSuccess() or reportError0- function can compose the data object and then call this function. We only need one place in which the actual sending is done.

packages/office-addin-dev-settings/src/commands.ts
igor-ribeiiro: Makes sense. It was like that before, but I changed it to put together with the report exception function and that changed these.
The way I am thinking of making the change you suggested is putting an ```if``` in the beginning of the ```reportException``` to check if the error is an expected error. If that's the case, we treat it as succeded as needed.

I like this way because the user only calls a single function (```reportException```) for all possible errros, but in the back, we are checking and treating the cases differently.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: It still exists!
It is just that the changes are getting pretty long and github is showing the changes in quite a weird way.

I will tag you in where you can find the function. I didn't even touch it.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -271,10 +285,33 @@ export class OfficeAddinUsageData {
     delete this.usageDataClient.context.tags["ai.user.accountId"]; // subscription
   }
 
+  /**
+   * Reports custom event object to Application Insights
+   * @param projectName Project name sent to Application Insights
+   * @param data Data object(s) sent to Application Insights
+   */
+  public sendUsageDataEvent(data: object = {}) {
```
igor-ribeiiro, line 293: @akrantz Here is the ```sendUsageDataEvent``` as normal

PR #388, opened 2021-02-17
Updating eslint versions for Office-Adding-Lint

igor-ribeiiro:
Updating ```eslint``` versions to update the version of ```lodash``` package used by them.

The ```lodash``` version 4.17.19 has a vulnerability issue and needs to be updated to avoid that.

Tested with:
1. ```npm install```
2. ```npm run build```
3. ```npm run test```
4. Checked that the ```lodash``` version was updated to the latest in package-lock.json

### DIFF
--- a/packages/office-addin-lint/package.json
+++ b/packages/office-addin-lint/package.json
@@ -22,15 +22,15 @@
     "Office Add-in"
   ],
   "dependencies": {
-    "@typescript-eslint/eslint-plugin": "^2.26.0",
-    "@typescript-eslint/parser": "^2.26.0",
+    "@typescript-eslint/eslint-plugin": "^4.15.1",
+    "@typescript-eslint/parser": "^4.15.1",
     "commander": "^6.2.0",
     "eslint": "^7.20.0",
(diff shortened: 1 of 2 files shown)


packages/office-addin-lint/package.json
```diff
@@ -25,9 +25,9 @@
     "@typescript-eslint/eslint-plugin": "^2.26.0",
```
akrantz, line 25: Are there newer versions of the typescript-eslint packages as well?

Should we also update the prettier version?

akrantz (changes requested):
See comments.

packages/office-addin-lint/package.json
igor-ribeiiro: Yes there are, they just don't need to be updated for the sake of fixing this ```lodash``` vulnerability bug.
I will update them for the sake of completeness.

akrantz:
@TCourtneyOwen It looks like there are test failures related to office-addin-sso.

TCourtneyOwen (approved):

PR #389, opened 2021-02-19
add --no-sideload option to start command (office-addin-debugging)

akrantz:


packages/office-addin-debugging/src/cli.ts
```diff
@@ -19,12 +19,12 @@ commander
     .option("--document <document>", "Document to be used for sideloading.")
     .option("--no-debug", "Start without debugging.")
     .option("--no-live-reload", "Do not enable live-reload.")
-    .option("--dev-tools", "Open the web browser developer tools when debugging (if supported)")
+    .option("--no-sideload", "Do not start the Office app and load the add-in.")
```
igor-ribeiiro, line 22: Shouldn't it just say something like: "Start the Office app, but do not load the add-in"?
The way that is currently written makes it sounds like nothing opens (neither the app nor the add-in)

TCourtneyOwen: should we add this to the ReadMe

packages/office-addin-debugging/src/start.ts
```diff
@@ -219,6 +219,12 @@ export interface StartDebuggingOptions {
      */
     enableLiveReload?: boolean,
 
+    /**
```
TCourtneyOwen, line 222: Will this be a breaking change given that we are introducing a new param?

akrantz: No, because we moved it to the options object, so the function params aren't changing.   😄

packages/office-addin-debugging/src/cli.ts
akrantz: Added

akrantz: @igor-ribeiiro It is written that way because the concept of sideloading covers both launching the app and loading the add-in. If sideload is turned off it will not launch the app when you start debugging.

TCourtneyOwen (approved):

PR #391, opened 2021-02-23
Adding possibility to report expected error

igor-ribeiiro:
Adding functions:
```reportSuccess```
```reportExpectedException```
```reportException```

And deprecating old ones.
Adding ```expectedError``` class.

Tested by calling each function and seeing the results in the logs.

### DIFF
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -24,6 +24,20 @@ export enum UsageDataLevel {
 }
 
 /**
+ * Defines an error that is expected to happen given some situation
+ * @member message Message to be logged in the error
+ */
+export class ExpectedError extends Error {
+  constructor(message: string | undefined) {
+      super(message);
+
+      // need to adjust the prototype after super()
+      // See https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work
+      Object.setPrototypeOf(this, ExpectedError.prototype);
+  }
+}
+
+/**
  * UpdateData options
  * @member groupName Group name for usage data settings (Optional)
  * @member projectName The name of the project that is using the usage data package.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/office-addin-usage-data/src/usageData.ts [hunk 1]
```diff
@@ -271,10 +285,39 @@ export class OfficeAddinUsageData {
     delete this.usageDataClient.context.tags["ai.user.accountId"]; // subscription
   }
 
+   /**
+   * Transform the project name by adddin '-test' suffix to it if necessary
+   */
+  private getEventName() {
```
akrantz, line 291: Could you please move the functions so they are in alphabetic order?

akrantz (changes requested):
Looks good. I think the diffs will be easier to read if the functions are in alphabetic order.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: I didn't even know this convention existed. 
Looking into the code, looks like the file doesn't follow this at all (I could change around the functions to fix that).
Until then, I've put it near the get function, which makes more sense.

akrantz (approved):
Looks good.

You could move the reportError(), reportExpectedError(), and reportSuccess() up higher so they come before the send functions and are in alphabetic order.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -272,44 +293,151 @@ export class OfficeAddinUsageData {
...
+  }
+
+  /**
+   * Reports custom exception event object to Application Insights
+   * @param method Method name sent to Application Insights
+   * @param err Error or message about error sent to Application Insights
+   * @param data Data object(s) sent to Application Insights
+   * @deprecated Use `reportUnexpectedError` instead.  
+   */
+  public sendUsageDataException(method: string, err: Error | string, data: object = {}) {
```
akrantz, line 340: This isn't yet in alphabetic order.

igor-ribeiiro: I have put the non-deprecated functions in alphabetic order, while the deprecated ones just after each one of them.
Do you think putting every function in alphabetic, not caring about deprecated better?

akrantz: Yes, just having them all in alphabetic order. Thanks.

akrantz (approved):
Looks good.

akrantz:
@TCourtneyOwen Please take a quick look. Let me know if you have any concerns.

TCourtneyOwen (approved):

PR #392, opened 2021-02-25
Adding telemetry to office-addin-dev-settings

igor-ribeiiro:
- Added telemetry with the new functions created (```reportSuccess``` and ```reportException```) to ```office-addin-dev-settings```.
- Transformed some already existing errors into ```ExpectedError```.
- Added telemetry to functions that are not directly called on ```cli.ts``` file by putting the string suffix "()" to the report message, as discussed with @akrantz.

Tested by running command and seeing the logs on Office-Addin-Usage-Data | Logs


akrantz (dismissed):

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -261,14 +295,16 @@ export function parseWebViewType(webViewString?: string): devSettings.WebViewTyp
     case undefined:
       return undefined;
     default:
-      throw new Error(`Please select a valid web view type instead of '${webViewString!}'.`);
+      throw new ExpectedError(`Please select a valid web view type instead of '${webViewString!}'.`);
```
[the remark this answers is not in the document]
igor-ribeiiro: Oh no, I added it just by copy-pasting from another place. 
It is not needed probably, I will update it.

packages/office-addin-dev-settings/src/commands.ts
```diff
@@ -61,7 +65,9 @@ export async function clear(manifestPath: string) {
     await devSettings.clearDevSettings(manifest.id!);
 
     console.log("Developer settings have been cleared.");
+    usageDataObject.reportSuccess("clear");
```
TCourtneyOwen, line 68: can we make this "clearDevSettings" so it's more obvious what method it's related to in our usage data?

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -70,8 +71,8 @@ export async function enableRuntimeLogging(path: string): Promise<void> {
 }
 
 export function getDeveloperSettingsRegistryKey(addinId: string): registry.RegistryKey {
-  if (!addinId) { throw new Error("addinId is required."); }
-  if (typeof addinId !== "string") { throw new Error("addinId should be a string."); }
+  if (!addinId) { throw new ExpectedError("addinId is required."); }
```
TCourtneyOwen, line 74: I wonder if we should make this more specific to let the user know that it's "manifest" ID is missing?

packages/office-addin-dev-settings/src/commands.ts
akrantz: It is the "clear" command for office-addin-dev-settings package. Which is provided in the data. Is this different than "start" for debugging or "webview" command?

packages/office-addin-dev-settings/src/dev-settings-windows.ts
akrantz: This is not neccessarily from the manifest. It is in an input parameter to the function.  We should keep both checks and I think the existing messaging is OK. If you want to tweak it you can make them:

"The addIn parameter is required." and "The addIn parameter should be a string."

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -71,7 +71,7 @@ export async function enableRuntimeLogging(path: string): Promise<void> {
 }
 
 export function getDeveloperSettingsRegistryKey(addinId: string): registry.RegistryKey {
-  if (!addinId) { throw new ExpectedError("addinId is required."); }
+  if (!addinId) { throw new ExpectedError("addinId is required. Manifest ID is missing."); }
```
akrantz: You just changed this before having a chance to read my comments. I don't think this is better. See my earlier comment about this.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
igor-ribeiiro: I guess this new message is a little bit more clear and I'm already setup to change this message, so I will update it to be more complete as you suggested

TCourtneyOwen (approved):

TCourtneyOwen (approved):

PR #393, opened 2021-02-25
Adding outlook version check

igor-ribeiiro:
Checking  the Outlook install version as part of the sideload command to determine if it's a valid version (needs to >= 16.0.13709.10000)

Tested with my computer with the version installed and tested with: 16.0.13628.20346
It successfully throws an ```ExpectedError``` in that case.

### DIFF
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -13,6 +13,7 @@ import {
   readManifestFile,
 } from "office-addin-manifest";
 import open = require("open");
+import semver = require('semver');
 import * as os from "os";
 import * as path from "path";
 import * as util from "util";
@@ -22,6 +23,7 @@ import { startDetachedProcess } from "./process";
 import { chooseOfficeApp } from "./prompt";
 import * as registry from "./registry";
 import { ExpectedError } from "office-addin-usage-data";
+
 
 const readFileAsync = util.promisify(fs.readFile);
 
@@ -216,8 +218,9 @@ async function getOutlookVersion(): Promise<string | undefined> {
     const OutlookInstallPathVersionRegistryKey = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Office\\ClickToRun\\Configuration`;
     const key = new registry.RegistryKey(`${OutlookInstallPathVersionRegistryKey}`);
     const outlookInstallVersion: string | undefined = await registry.getStringValue(key, "ClientVersionToReport");
-
-    return outlookInstallVersion;
+    const outlookSmallerVersion = outlookInstallVersion?.split(`.`, 3).join(`.`);
+    
+    return outlookSmallerVersion;
   }
   catch (err) {
     return undefined;
@@ -328,8 +331,8 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
       // for Outlook, open Outlook.exe; for other Office apps, open the document
       if (app == OfficeApp.Outlook) {
         const version = await getOutlookVersion();
-        if (version && version < "16.0.13709.10000") {
-          throw new ExpectedError(`Outlook install version should be 16.0.13709.10000 or greater. Current version: ${version}`);
+        if (version && semver.gt(version, "16.0.13709")) {
+          throw new ExpectedError(`Outlook install version should be 16.0.13709 or greater. Current version: ${version}`);
         }
         sideloadFile = await getOutlookExePath();
       } else {
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -213,8 +213,8 @@ function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
   return false;
 }
 
-function hasOfficeVersion(targetVersion: string, currentVersion: string): boolean {
-  return semver.gte(currentVersion, targetVersion)
+function hasOfficeVersion(targetVersion: string, currentVersion: string | undefined): boolean {
+  return currentVersion ? semver.gte(currentVersion, targetVersion) : false;
 }
 
 async function getOutlookVersion(): Promise<string | undefined> {
@@ -228,11 +228,6 @@ async function getOutlookVersion(): Promise<string | undefined> {
   catch (err) {
     return undefined;
   }
-}
-
-async function hasOutlookVersion(version: string): Promise<boolean> {
-  const currentVersion: string | undefined = await getOutlookVersion();
-  return currentVersion ? hasOfficeVersion(version, currentVersion) : false;
 }
 
 async function getOutlookExePath(): Promise<string | undefined> {
@@ -338,7 +333,8 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
       await registerAddIn(manifestPath);
       // for Outlook, open Outlook.exe; for other Office apps, open the document
       if (app == OfficeApp.Outlook) {
-        if (!(await hasOutlookVersion("16.0.13709"))) {
+        const version = await getOutlookVersion();
+        if (!hasOfficeVersion("16.0.13709", version)) {
           throw new ExpectedError(`The current version of Outlook does not support sideload. Please use version 16.0.13709 or greater.`);
         }
         sideloadFile = await getOutlookExePath();
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -213,8 +213,8 @@ function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
   return false;
 }
 
-function hasOfficeVersion(targetVersion: string, currentVersion: string | undefined): boolean {
-  return currentVersion ? semver.gte(currentVersion, targetVersion) : false;
+function hasOfficeVersion(targetVersion: string, currentVersion: string): boolean {
+  return semver.gte(currentVersion, targetVersion);
 }
 
 async function getOutlookVersion(): Promise<string | undefined> {
@@ -333,8 +333,8 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
       await registerAddIn(manifestPath);
       // for Outlook, open Outlook.exe; for other Office apps, open the document
       if (app == OfficeApp.Outlook) {
-        const version = await getOutlookVersion();
-        if (!hasOfficeVersion("16.0.13709", version)) {
+        const version: string | undefined = await getOutlookVersion();
+        if (version && !hasOfficeVersion("16.0.13709", version)) {
           throw new ExpectedError(`The current version of Outlook does not support sideload. Please use version 16.0.13709 or greater.`);
         }
         sideloadFile = await getOutlookExePath();
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -13,7 +13,7 @@ import {
   readManifestFile,
 } from "office-addin-manifest";
 import open = require("open");
-import semverlt = require('semver/functions/lt');
+import semver = require('semver');
 import * as os from "os";
 import * as path from "path";
 import * as util from "util";
@@ -330,7 +330,7 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
       // for Outlook, open Outlook.exe; for other Office apps, open the document
       if (app == OfficeApp.Outlook) {
         const version = await getOutlookVersion();
-        if (version && semverlt(version, "16.0.13709")) {
+        if (version && semver.lt(version, "16.0.13709")) {
           throw new ExpectedError(`The current version of Outlook (${version}) does not support sideload. Please use version 16.0.13709 or greater.`);
         }
         sideloadFile = await getOutlookExePath();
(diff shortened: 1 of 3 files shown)


packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -311,11 +326,16 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
 
       await registerAddIn(manifestPath);
-
       // for Outlook, open Outlook.exe; for other Office apps, open the document
-      sideloadFile = (app === OfficeApp.Outlook)
-        ? await getOutlookExePath()
-        : await generateSideloadFile(app, manifest, document);
+      if (app == OfficeApp.Outlook) {
+        const version = await getOutlookVersion() ?? "";
+        if (version < "16.0.13709.10000") {
```
TCourtneyOwen, line 332: This is a string value so I think we need to convert this to a number and  compare it to the number 16.0.13709.10000

akrantz: For version numbers, you need to compare each numeric portion separately.

akrantz:
Is this only going to work for C2R? I would recommend failing only if we are certain what the version is and that it is older, and if we cannot tell the version, do not fail.

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: Why can't I compare them as strings directly?
The versions are in alphabetical order!

Below there's a StackOverflow item on how to compare strings:
https://stackoverflow.com/questions/10198257/comparing-2-strings-alphabetically-for-sorting-purposes#:~:text=JavaScript%20compares%20strings%20character%20by%20character%20and%20%22a%22,all%20equal,%20lets%20move%20to%20the%20next%20character.

akrantz: Version numbers do not follow alphabetic order.

"16.0.30" is greater than "16.0.130" alphabetically, but 16.0.130 is a bigger version number.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -210,6 +211,20 @@ function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
...
+async function getOutlookVersion(): Promise<string | undefined> {
+  try {
+    const OutlookInstallPathVersionRegistryKey = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Office\\ClickToRun\\Configuration`;
+    const key = new registry.RegistryKey(`${OutlookInstallPathVersionRegistryKey}`);
+    const outlookInstallVersion: string | undefined = await registry.getStringValue(key, "ClientVersionToReport");
+
+    return outlookInstallVersion;
+  }
+  catch (err) {
+    const errorMessage: string = `Unable to find Outlook version location: \n${err}`;
```
akrantz, line 223: We need to be careful here. There are other ways to install Outlook, and we don't want the inability to get the version number to block usage there. I would recommend this function not throw an error but just return undefined if it cannot get the version.

igor-ribeiiro:
> Is this only going to work for C2R? I would recommend failing only if we are certain what the version is and that it is older, and if we cannot tell the version, do not fail.

What is C2R?
I just added an extra check to check if the version was found and it is defined before throwing the error.

akrantz (changes requested):
See comments. 

1. The version check needs to be based on the numeric components. It would be best to either use an existing function that can do this right, or write one yourself.
2. Sideload should not fail if we cannot get the version.
3. You should look at non C2R installs to see where the version is. It would be best if this was in one place regardless of install type.

packages/office-addin-dev-settings/src/sideload.ts
TCourtneyOwen: You can do something like we do in generator office for checking the version of node:

    if (parseInt(process.version.slice(1, process.version.indexOf('.'))) % 2 == 1) {
      console.log(yosay('generator-office does not support your version of Node. Please switch to the latest LTS version of Node.'));
      this._exitProcess();
    }
We really are only interested in this part of the version: 16.0.13709

TCourtneyOwen: Also you can change your local version to Production, by going to Outlook File--> Office Account--> Office Account and selecting the Office Insider button. This gives you the option to the "Change Channel" (I assume you are currently in the Dogfood Channel, but you can change it to "Microsoft Current Channel (OuterRing)" in the following dialog:

image
 
Be sure you check the privacy checkbox or in my experience the settings don't take effecr.

You can then go to Office Updates under Outlook File--> Office Account--> Office Account and do Update Now and this will install a Production version of Office so you will have a version >= 16.0.13709.10000

After you are done testing, you probably want to change your channel back "Dogfood Main Channel" for reapply the updates to install back to your original version

TCourtneyOwen:
"C2R" means "Click to Run". This is way most people install Office these days, as this is the way Office365 installs Office. However, some smaller set of people still prefer to install Office the old way through an MSI (Microsoft Installer). I am not sure if the current reg-key we are checking will capture the version of MSI installs

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: I just updated the PR by using a package named ```semver```, that compares the versions directly. So that should be taken care of now.

TCourtneyOwen:
I have done some research and haven't been able to find a way to determine an MSI install version in the registry. It's probably there somewhere, but I haven't found any postings yet about it

TCourtneyOwen:
I would be sure that your change works for the supported case by installing the latest Production build using the steps I outlined to change your install channel

igor-ribeiiro:
> I would be sure that your change works for the supported case by installing the latest Production build using the steps I outlined to change your install channel

Yeah, I just updated my outlook version, and adding the add-in worked!
So I Tested the error not passing before (with the old version) and now it properly working with the new code

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -210,6 +213,20 @@ function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
   return false;
 }
 
+async function getOutlookVersion(): Promise<string | undefined> {
+  try {
+    const OutlookInstallPathVersionRegistryKey = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Office\\ClickToRun\\Configuration`;
+    const key = new registry.RegistryKey(`${OutlookInstallPathVersionRegistryKey}`);
```
akrantz, line 219: There's no need for the OutlookInstallPathVersionRegistryKey variable.  Just put it directly:

`... new RegistryKey("HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Office\\ClickToRun\\Configuration");`

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -210,6 +213,20 @@ function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
...
+    const outlookInstallVersion: string | undefined = await registry.getStringValue(key, "ClientVersionToReport");
+    const outlookSmallerVersion = outlookInstallVersion?.split(`.`, 3).join(`.`);
```
akrantz, line 221: Why do you split and join here?

TCourtneyOwen:
When you say it worked, did sideloading of an addin in Outlook actually work for you? I am interested because this will validate that the Outlook sideloading feature is fully rolled out to Production. If it worked you should have gotten a prompt such as the following when Outlook boots:

image

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -311,11 +328,16 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
       await registerAddIn(manifestPath);
-
       // for Outlook, open Outlook.exe; for other Office apps, open the document
-      sideloadFile = (app === OfficeApp.Outlook)
-        ? await getOutlookExePath()
-        : await generateSideloadFile(app, manifest, document);
+      if (app == OfficeApp.Outlook) {
+        const version = await getOutlookVersion();
+        if (version && semver.lt(version, "16.0.13709")) {
+          throw new ExpectedError(`Outlook install version should be 16.0.13709 or greater. Current version: ${version}`);
```
akrantz, line 335: This error message would be better if it mentioned that sideload is not supported.

`The current version of Outlook (${version}) does not support sideload. Please use version 16.0.13709 or greater.`

akrantz (changes requested):
See comments.

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: Because the version that gets returned is in the format:  3 "a.b.c.d", but the API only supports version in the format: "a.b.c"
Given that we only care about the first 3 steps, I do this split join step.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -311,11 +328,16 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
...
 
       await registerAddIn(manifestPath);
-
       // for Outlook, open Outlook.exe; for other Office apps, open the document
-      sideloadFile = (app === OfficeApp.Outlook)
-        ? await getOutlookExePath()
-        : await generateSideloadFile(app, manifest, document);
+      if (app == OfficeApp.Outlook) {
+        const version = await getOutlookVersion();
+        if (version && semver.gt(version, "16.0.13709")) {
```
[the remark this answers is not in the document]
igor-ribeiiro: The error only occurs when the version is smaller right? So we need don't need to check the equal then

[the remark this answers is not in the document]
akrantz: Yeah, you have the logic reversedand you're using lt() now, so it should be good.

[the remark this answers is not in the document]
igor-ribeiiro: Just imported what is necessary now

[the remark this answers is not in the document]
igor-ribeiiro: Reverting. It is giving me an error when I try to import a specific function and I didn't find any example of that in their docs.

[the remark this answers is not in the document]
akrantz: OK

igor-ribeiiro:
> When you say it worked, did sideloading of an addin in Outlook actually work for you? I am interested because this will validate that the Outlook sideloading feature is fully rolled out to Production. If it worked you should have gotten a prompt such as the following when Outlook boots:

Yeah, it fully worked and I did get this prompt

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -329,9 +338,8 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
       await registerAddIn(manifestPath);
       // for Outlook, open Outlook.exe; for other Office apps, open the document
       if (app == OfficeApp.Outlook) {
-        const version = await getOutlookVersion();
-        if (version && semver.lt(version, "16.0.13709")) {
-          throw new ExpectedError(`The current version of Outlook (${version}) does not support sideload. Please use version 16.0.13709 or greater.`);
+        if (!(await hasOutlookVersion("16.0.13709"))) {
```
akrantz, line 341: It would be better to assign it to a variable first:

```
const version = await getOutlookVersion();

if (!hasOfficeVersion(...
```

I don't think really need the hasOutlookVersion() function.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -213,8 +213,8 @@ function isSideloadingSupportedForWebHost(app: OfficeApp): boolean {
   return false;
 }
 
-function hasOfficeVersion(targetVersion: string, currentVersion: string): boolean {
-  return semver.gte(currentVersion, targetVersion)
+function hasOfficeVersion(targetVersion: string, currentVersion: string | undefined): boolean {
```
akrantz, line 216: This function should only take string, and not take undefined.

You should put the logic about the fact that the getOutlookVersion() function can return undefined outside of here.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -338,7 +333,8 @@ export async function sideloadAddIn(manifestPath: string, app?: OfficeApp, canPr
       await registerAddIn(manifestPath);
       // for Outlook, open Outlook.exe; for other Office apps, open the document
       if (app == OfficeApp.Outlook) {
-        if (!(await hasOutlookVersion("16.0.13709"))) {
+        const version = await getOutlookVersion();
+        if (!hasOfficeVersion("16.0.13709", version)) {
```
akrantz, line 337: Do the check about whether version is undefined here.

It would be a good practice to put type info on the version variable, e.g. `const version: string | undefined = `
It will still be implicit from the return of getOutlookVersion().
You don't want to be able to pass undefined into hasOfficeVersion(). Rather just check here.

`if (version && !hasOfficeVersion())`

In this specific case, we want the fact that if the version cannot be retrieved we still allow it to sideload.

igor-ribeiiro: Sure, thanks for the feedback!
Just made those changes :D

So, in general, we like to treat the undefined outside a function?

akrantz (dismissed):

TCourtneyOwen (dismissed):

TCourtneyOwen (approved):

PR #395, opened 2021-03-01
Added the new telemetry syntax to the following packages, which are all that had previously telemetry implemented: 

igor-ribeiiro:
Added the new telemetry syntax to the following packages, which are all that had previously telemetry implemented: 
- office-addin-debugging 
- office-addin-dev-certs 
- office-addin-lint 
- office-addin-manifest 
- office-addin-sso

### DIFF
--- a/packages/office-addin-lint/src/lint.ts
+++ b/packages/office-addin-lint/src/lint.ts
@@ -37,7 +37,7 @@ export function performLintCheck(files: string) {
     usageDataObject.reportSuccess("performLintCheck()", { exitCode: ESLintExitCode.NoLintErrors });
   } catch (err) {
     if (err.status && err.status == ESLintExitCode.HasLintError) {
-      usageDataObject.reportExpectedException("performLintCheck()", "Has lint error", { exitCode: ESLintExitCode.HasLintError });
+      usageDataObject.reportExpectedException("performLintCheck()", err, { exitCode: ESLintExitCode.HasLintError });
     } else {
       usageDataObject.reportException("performLintCheck()", err);
     }
@@ -57,7 +57,7 @@ export function performLintFix(files: string) {
     usageDataObject.reportSuccess("performLintFix()", { exitCode: ESLintExitCode.NoLintErrors });
   } catch (err) {
     if (err.status && err.status == ESLintExitCode.HasLintError) {
-      usageDataObject.reportExpectedException("performLintFix()", "Has lint error", { exitCode: ESLintExitCode.HasLintError });
+      usageDataObject.reportExpectedException("performLintFix()", err, { exitCode: ESLintExitCode.HasLintError });
     } else {
       usageDataObject.reportException("performLintFix()", err);
     }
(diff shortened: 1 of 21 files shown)


packages/office-addin-lint/src/lint.ts
```diff
@@ -34,12 +34,12 @@ export function performLintCheck(files: string) {
   try {
     const command = getLintCheckCommand(files);
     execCommand(command);
-    usageDataObject.sendUsageDataSuccessEvent("performLintCheck", { exitCode: ESLintExitCode.NoLintErrors });
+    usageDataObject.reportSuccess("performLintCheck()", { exitCode: ESLintExitCode.NoLintErrors });
   } catch (err) {
     if (err.status && err.status == ESLintExitCode.HasLintError) {
-      usageDataObject.sendUsageDataSuccessEvent("performLintCheck", { exitCode: ESLintExitCode.HasLintError });
+      usageDataObject.reportExpectedException("performLintCheck()", "Has lint error", { exitCode: ESLintExitCode.HasLintError });
```
akrantz, line 40: Why is this string "Has lint error"? Sorry I didn't look at the code and cannot tell here when looking at the diff.

igor-ribeiiro: To send an "expectedException", it is required to pass the ```err``` parameter to the functions.
So I passed this string to represent the expected error.

akrantz (commented):
Looks good overall. It is interesting that there was a use case to call reportExpectedException(). 

I'm curious about how it is used with regard to the second string "Has lint problem" or "Has formatting issue". Just wondering if that string should be different or not...

igor-ribeiiro:
> Looks good overall. It is interesting that there was a use case to call reportExpectedException().
> 
> I'm curious about how it is used with regard to the second string "Has lint problem" or "Has formatting issue". Just wondering if that string should be different or not...

I also liked to see that the code already had a use case for the  ```reportExpectedException``` function.
This string parameter will be the error message that will show on the logs.

So, when looking through the logs, you could see that is had ```Succeed = true```, while being an error. Then you could look into this message and see why this expected error happened.

packages/office-addin-lint/src/lint.ts
akrantz: Based on the function definitiion:
`public reportExpectedException(method: string, err: Error | string, data: object = {}): void`

I would expect the second parameter to be `err` and not `"Has lint error"`. The exitCode actually distinguishes the cases because exitCode 1 means there is a lint error and 2 for unexpected error.

Let me know if I'm missing something.

igor-ribeiiro: Sure, a mistake by me. Err should be passed instead of the string

akrantz:
@igor-ribeiiro Does this remove all existing usage of the deprecated usage data methods now?

akrantz (approved):
Looks good to me. Should probably wait for approval from @millerds too before merging.

igor-ribeiiro:
> @igor-ribeiiro Does this remove all existing usage of the deprecated usage data methods now?

Was missing some tests in the usage-data package. Now, there are 0 uses of the deprecated functions within this repository.

akrantz (approved):
I'd like to have @millerds also approve. 

A tip when you squash: you can edit the commit message and copy the PR description rather than using the default text where it puts the individual commit text (which isn't useful here).

millerds (approved):

PR #397, opened 2021-03-05
custom functions metadata generation doesn't require file output

akrantz:
The custom functions metadata plugin was always writing to an output file, but it doesn't need to do so. It just needs to emit the json metadata and webpack can either write it out to file or serve it up when using webpack-dev-server.

I have made a breaking change to custom-functions-metadata where the generate() function is renamed to generateCustomFunctionsMetadata() and does not take an output file path. 

The custom-functions-metadata CLI command `generate` still supports an output file path, but it is now optional. If provides, it writes to the file, and if not, the metadata is output to the console.


packages/custom-functions-metadata/test/src/testCases.ts
```diff
@@ -72,6 +79,10 @@ describe("test cases", function() {
                         }
 
                         assert.strictEqual(actualMetadata, expectedMetadata, "metadata does not match expected");
+                        if (actualMetadata !== expectedMetadata) {
```
akrantz, line 82: this should be removed. It was for debugging.

packages/custom-functions-metadata/test/src/testCases.ts
```diff
@@ -48,7 +48,7 @@ describe("test cases", function() {
                         const expectedErrorsFile = path.join(testCaseDirPath, `expected.${scriptType}.errors.txt`);
                         const actualMetadataFile = path.join(testCaseDirPath, `actual.${scriptType}.json`);
                         const expectedMetadataFile = path.join(testCaseDirPath, "expected.json");
-                        const expectedMetadata: string | undefined = readFileIfExists(expectedMetadataFile);
+                        const expectedMetadata: string | undefined = readFileIfExists(expectedMetadataFile) || "";
```
igor-ribeiiro, line 51: Why is the: ```|| ""``` needed here?

packages/custom-functions-metadata/test/src/test.ts
```diff
@@ -129,34 +110,7 @@ describe("test errors", function() {
...
-    });
-});
-describe("delete test files", function() {
-    describe("deleting files", function() {
-        it("files to delete", function() {
-            const outputJavaScript = "testjs.json";
-            const outputTypeScript = "test.json";
-            fs.unlinkSync(outputJavaScript);
-            fs.unlinkSync(outputTypeScript);
+             assert.equal(generateResult.metadataJson, "", "should not be any metadata");
```
igor-ribeiiro, line 113: Couldn't we use ```undefined``` instead of the empty ```""``` string?

igor-ribeiiro (changes requested):
Just small changes related to the use of empty string vs undefined.

packages/custom-functions-metadata/test/src/testCases.ts
akrantz: If there are errors the actualMetadata will be an empty string now. Previously, it would have not written the file, so reading the file would return undefined.  Therefore, when there is an error, the actualMetadata will be an empty string, and if the expectedMetadataFile doesn't exist -- which it shouldn't -- then the expectedMetadata is also an empty string.

packages/custom-functions-metadata/test/src/test.ts
akrantz: The actual IGenerateResult.metadataJson is a string, and will be empty when there are errors. It's not optional so there is no need to also have it be undefined.

igor-ribeiiro (dismissed):

packages/custom-functions-metadata/test/src/testCases.ts
```diff
@@ -48,7 +48,7 @@ describe("test cases", function() {
                         const expectedErrorsFile = path.join(testCaseDirPath, `expected.${scriptType}.errors.txt`);
                         const actualMetadataFile = path.join(testCaseDirPath, `actual.${scriptType}.json`);
                         const expectedMetadataFile = path.join(testCaseDirPath, "expected.json");
-                        const expectedMetadata: string | undefined = readFileIfExists(expectedMetadataFile) || "";
+                        const expectedMetadata = readFileIfExists(expectedMetadataFile) || "";
```
igor-ribeiiro, line 51: Why are we removing this explicit variable assign?

akrantz: I looked over the diff and noticed that the `| undefined` should be removed. But rather than leave `: string`, I decided to just remove it.

I would say that in general, it is better to put the types on variables. 

In this case, because the other variables don't have it, I decided to deviate from this.

igor-ribeiiro: Makes sense

igor-ribeiiro (approved):

millerds (approved):

PR #398, opened 2021-03-08
Update Webpack to fix elliptic 6.5.3 vulnerability issue

igor-ribeiiro:
Elliptic 6.5.3 has a vulnerability issue and needs to be removed from the code. It currently is imported by ```webpack``` in package ```custom-functions-metadata-plugin```. Thus ```webpack``` needed to be updated to a major version, and that's why there are a lot of changes.

Tests:
- Tested by running and building the excel-custom-functions project


akrantz:
Does this mean that Excel-Custom-Functions must use webpack v5?  If this requires webpack 5, then I suggest the major version be bumped when published. 

You should publish to the next dist-tag, then validate by installing custom-functions-metadata-plugin@next into Excel-Custom-Functions and making sure it works.

igor-ribeiiro:
> Does this mean that Excel-Custom-Functions must use webpack v5? If this requires webpack 5, then I suggest the major version be bumped when published.
> 
> You should publish to the next dist-tag, then validate by installing custom-functions-metadata-plugin@next into Excel-Custom-Functions and making sure it works.

I didn't even think about that, but yes, the webpack version should be updated as well in Excel-Custom-Functions, given that the current version uses a vulnerable package. Currently only the webpack version inside the plugin has been updated

akrantz:
Please resolve merge conflicts. These changes are fine. 

I really want to understand if the consumers of CustomFunctionsMetadataPlugin in webpack config means that webpack 5 must now be used. 

We will want to do a major version bump of the custom-functions-metadata-plugin package when "npm run publish:next" is run,

akrantz (changes requested):
Ping me after you merge.

igor-ribeiiro:
> Please resolve merge conflicts. These changes are fine.
> 
> I really want to understand if the consumers of CustomFunctionsMetadataPlugin in webpack config means that webpack 5 must now be used.
> 
> We will want to do a major version bump of the custom-functions-metadata-plugin package when "npm run publish:next" is run,

I just resolved the merge conflicts.
Sure, I will use the publish with the @next tag to be extra safe.

The consumers of CustomFunctionsMetadataPlugin, like the Excel Custom Functions, can just keep using their webpack versions normally if they are dev dependencies.

millerds (dismissed):

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -27,17 +29,14 @@ class CustomFunctionsMetadataPlugin {
...
-                    source() { return generateResult.metadataJson; },
-                    size() { return generateResult.metadataJson.length; },
-                };
+                compilation.assets[this.options.output] = new sources.RawSource(generateResult.metadataJson);
             }
         });
 
         compiler.hooks.compilation.tap(pluginName, (compilation, params) => {
-            compilation.moduleTemplates.javascript.hooks.render.tap(pluginName, (source, module) => {
+            compilation.moduleTemplates.javascript.hooks.render.tap(pluginName, (source : any, module : any) => {
```
akrantz, line 39: I'm getting build errors for these changes...

akrantz (changes requested):
This does not build locally for me. 

```
[0]
[0] > custom-functions-metadata-plugin@1.0.44 tsc
[0] > tsc -p tsconfig.json
[0]
[0] src/customfunctionsplugin.ts(7,49): error TS7016: Could not find a declaration file for module 'webpack'. 'c:/dev/git/Office-Addin-Scripts/packages/custom-functions-metadata-plugin/node_modules/webpack/lib/webpack.js' implicitly has an 'any' type.
[0]   Try `npm install @types/webpack` if it exists or add a new declaration (.d.ts) file containing `declare module 'webpack';`
[0] src/customfunctionsplugin.ts(30,46): error TS7006: Parameter 'compilation' implicitly has an 'any' type.
[0] src/customfunctionsplugin.ts(38,53): error TS7006: Parameter 'compilation' implicitly has an 'any' type.
[0] src/customfunctionsplugin.ts(38,66): error TS7006: Parameter 'params' implicitly has an 'any' type.
[0] npm ERR! code 2
[0] npm ERR! path c:\dev\git\Office-Addin-Scripts\packages\custom-functions-metadata-plugin
[0] npm ERR! command failed
[0] npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c tsc -p tsconfig.json
```

packages/custom-functions-metadata-plugin/package.json
```diff
@@ -26,11 +26,10 @@
   "dependencies": {
     "custom-functions-metadata": "^1.0.44",
     "es6-promise": "^4.2.8",
-    "webpack": "^4.42.1"
+    "webpack": "^5.24.4"
   },
   "devDependencies": {
     "@types/node": "^12.12.34",
-    "@types/webpack": "^4.41.10",
```
akrantz, line 33: I figured out that @types/webpack dev dependency is now missing.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
igor-ribeiiro: Wait, these changes seem to fix your bug!
Like, I added the ```any``` property on my own to fix this problem, so it is weird that this is happening

akrantz (approved):
OK, we worked through stuff so it is building now.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
akrantz: After getting the latest Node and NPM, I can now build everything.

packages/custom-functions-metadata-plugin/package.json
akrantz: webpack v5 no longer needs separate @types/webpack.

PR #399, opened 2021-03-09
Update version of pug

millerds:
Update to fix vulnerability


TCourtneyOwen (commented):
Actually we probably only want to include the package.json and package-lock.json change in this change.

akrantz:
@TCourtneyOwen What is pug for?

packages/eslint-config-office-addins/package-lock.json
```diff
@@ -1,1309 +1,1309 @@
...
-      }
-    },
-    "yargs-parser": {
-      "version": "20.2.6",
-      "resolved": "https://registry.npmjs.org/yargs-parser/-/yargs-parser-20.2.6.tgz",
-      "integrity": "sha512-AP1+fQIWSM/sMiET8fyayjx/J+JmTPt2Mr0FkrgqB4todtfa53sOsrSAcIrJRD5XS20bKUwaDIuMkWKCEiQLKA==",
-      "dev": true
-    }
-  }
+	"name": "eslint-config-office-addins",
```
akrantz, line 2: Why is the indent level different?

akrantz (changes requested):
Could you see about making the indent level the same so there aren't a bunch of whitespace changes?

packages/eslint-config-office-addins/package-lock.json
millerds: That was automatic . . . I'll have to manually take a look.

TCourtneyOwen:
pug is used in setting up the express server, although I can't recall why this is needed - this code came from the original template Ricky made.  https://www.npmjs.com/package/pug. 

Interestingly though, I just commented out this line from office-addin-sso (the only place pug is used), built, and linked to the Office-Addin-SSO template, and everything works fine:  this.appInstance.set('view engine', 'pug');

So, we could probably just remove pug and remove this like from %:\Office-Addin-Scripts\packages\office-addin-sso\src\app.ts

akrantz:
I vote to remove pug...

PR #401, opened 2021-03-10
Fixing lint problems with endline on windows

igor-ribeiiro:
linting started failing in Office-Addin-Taskpane (and presumably all our project repos) recently with errors (Delete ␍ prettier/prettier) about extra spaces on Windows.

This fix solves that linting problem. 

Tests:
- Linked this package to taskpane project
- Ran npm run lint there and these endline errors disappeared

### DIFF
--- a/packages/office-addin-lint/package.json
+++ b/packages/office-addin-lint/package.json
@@ -1,6 +1,6 @@
 {
   "name": "office-addin-lint",
-  "version": "1.1.1",
+  "version": "1.1.2",
   "description": "Provide eslint and prettier integration",
   "main": "./lib/main.js",
   "scripts": {
@@ -28,8 +28,8 @@
     "eslint": "^7.20.0",
     "eslint-config-prettier": "^7.2.0",
     "eslint-plugin-prettier": "^3.3.1",
-    "office-addin-prettier-config": "^1.0.14",
-    "office-addin-usage-data": "^1.0.24",
+    "office-addin-prettier-config": "^1.0.15",
+    "office-addin-usage-data": "^1.0.25",
     "prettier": "^2.2.1"
   },
   "devDependencies": {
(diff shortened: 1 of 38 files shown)


TCourtneyOwen:
Did you run "npm run lint" on Office-Addin-Taskpane with these changes?

igor-ribeiiro:
> Did you run "npm run lint" on Office-Addin-Taskpane with these changes?

Yes, that's how I tested

packages/office-addin-lint/package.json
```diff
@@ -79,6 +79,12 @@
     "extends": [
       "prettier",
       "prettier/@typescript-eslint"
+    ],
+    "prettier/prettier": [
```
akrantz, line 83: It would be good to discuss the ramification of this change. Without this option, I believe it normalizes to use LF line endings. By setting "endOfLine" to "auto", how does this affect Windows and non-Windows platforms? Does it change to normalize to CRLF, or does it use whatever is there?

Also, I think this is related to VS Code settings. Is there a change that could be made in VS Code so it allows using LF line endings rather than changing this?

akrantz: I will say that for the other React-based web app that I am working on, it was changed to use "endOfLine": "auto" so this is probably a worthwhile change. I just want to probe deeper and make sure we are making an informed decision on this.

akrantz: Also, did prettier recently change the line ending behavior?

akrantz: https://prettier.io/docs/en/options.html#end-of-line

akrantz: When I look at this, I think it is good to change to "auto".

akrantz (changes requested):
I think this change is in the wrong place. I don't think package.json should be changed for changing the default that is applied when office-addin-lint is used. (I could be mistaken.)

Also, there is no need to have changes to the package-lock.json file just because package.json was changed.

akrantz:
The lock file only needs to change when there are dependency changes. For this PR, there aren't any.

We also need to all consistently use NPM v7 as some people might be using v6, which I think has caused churn. There also seems to be an issue because if you have .npmrc with a private NPM registry, that can cause changes to the lock file.

igor-ribeiiro:
@akrantz I think the problems are solved now right? So could we merge this PR or there's something more should we discuss?

akrantz:
I think this is outdated now.

PR #404, opened 2021-03-18
Adding npmv7 support

igor-ribeiiro:
Updated the codebase to support npmv7 env variables changes. Checked for all occurrences of the variable of type:
```npm_package_scripts_*``` which don't get support anymore.

With this change, the templates will have to be updated to not use hyphens in the script's name anymore, because typescript can't have variables with hyphens. There will a separate PR on the templates to change the hyphens to underscores.

The package-lock.json changes are all created by running ```npm ci``` on the root with the latest npm version, so I guess it should be added in the PR.

Tests:
Linked it to Taskpane project and ran ```npm start```. It ran properly starting the dev server it was not started and running correctly if it was already started.

### CODE REVIEW

File: packages/office-addin-debugging/package.json [hunk 1]
```diff
@@ -29,7 +29,8 @@
     "office-addin-dev-settings": "^1.8.19",
     "office-addin-manifest": "^1.5.19",
     "office-addin-node-debugger": "^0.4.41",
-    "office-addin-usage-data": "^1.2.0"
+    "office-addin-usage-data": "^1.2.0",
+    "read-package-json-fast": "^2.0.2"
```
akrantz, line 33: Let's not add this to every package. You should have the functionality in one package that can then be used by the others.

akrantz (changes requested):
I still need to look through this. Let's discuss about the design.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -1,121 +1,123 @@
...
-        }
-
-        await stopDebugging(manifestPath);
-        usageDataObject.reportSuccess("stop");
-    } catch (err) {
-        usageDataObject.reportException("stop", err);
-        logErrorMessage(`Unable to stop debugging.\n${err}`);
-    }
-}
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
akrantz, line 1: There are whitespace issues that cause the diff to show changes on every line. Please update so I can actually tell what changed here.

packages/office-addin-debugging/src/commands.ts
```diff
@@ -1,121 +1,123 @@
...
+            throw new ExpectedError("--dev-server-port should be between 0 and 65535.");
+        }
+    }
+
+    return devServerPort;
+}
+
+export async function start(manifestPath: string, platform: string | undefined, command: commander.Command) {
+    try {
+        const pkg = await readJson(`${process.env.npm_package_json}`)
```
akrantz, line 46: I think you could make this work more simply and consolidate it into office-addin-cli. You can have a function which can take the string like "npm_package_config_app_type_to_debug", and if the package.json hasn't been parsed, you can parse it for the env vars that used to be set (npm_package_config_* and npm_package_script_*).

The goal here would be an existing piece of code that does `process.env.npm_packge_config_app_type_to_debug` should be replaced with `getPackageJsonConfig("app_type_to_debug") or getPackageJsonScript("dev_server").

packages/office-addin-debugging/package.json
igor-ribeiiro: I don't think I got it. I only added this one new package to the ```office-addin-debugging``` package and not any of the others.

packages/office-addin-debugging/src/commands.ts
igor-ribeiiro: I don't know why these files were created. I didn't even ran ```lint```, just the regular ```install``` and ```build``` and those changes were created.
What could it be?

packages/office-addin-debugging/src/commands.ts
igor-ribeiiro: The config variable still work, so I just didn't touch them. The npm team said they would keep it for now for easier upgrades.
I can make this ```getPackageJsonScript``` and it should be better to make the code more scalable long term, I agree

packages/office-addin-debugging/package.json
akrantz: This won't be specific to debugging. This should be usable for other places either now or in the future.

igor-ribeiiro: Being in ```office-addin-cli``` with some default functions there fixes this problem right?

packages/office-addin-cli/src/parse_json_script_variable.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+const readJson = require('read-package-json-fast');
+
+export async function parseJSONScriptVariable(variable: string): Promise <string | undefined> {
+    // NPMv7 Support
+    if(process.env.npm_package_json) {
+        const pkg = await readJson(`${process.env.npm_package_json}`);
```
akrantz, line 9: So you read and parse the package.json every time?  Maybe there's a better way...

igor-ribeiiro: Sure, my mistake. Just fixed it so it is only read once!

packages/office-addin-cli/src/parse_json_script_variable.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+const readJson = require('read-package-json-fast');
+
+let jsonPackage : any = undefined;
+
+export async function parseJSONScriptVariable(variable: string): Promise <string | undefined> {
```
akrantz, line 8: Consider the name of this function. If you were reading it, would you know what json is being parsed. Is parsing even that important here? When I consider this, I try to say to myself: what does this function do? It returns the value of the script from **package.json**. So a better name would be `getScriptFromPackageJson()` or `getPackageJsonScript()`. The other thing is it does not specify the package.json file, but this is implicit because of the NPM definition of `process.env.npm_package_json`. Do we want it to be implicit? There is no mention of NPM definition of environment variables. Should there be?

Perhaps it would be better to characterize this function as reading an NPM environment variable, e.g. `getNpmEnvVariable(name: string)` which takes "npm_package_script_*" and knows to look it up. This most closely replaces the direct behavior of `process.env.npm_package_script_dev_server` with `getNpmEnvVariable("npm_package_script_dev_server")` which is a simple replacement.

In terms of the implementation, if the env var name is passed in, the first thing it should do is see if that is defined and return it if so. 

In the next part, if it has not already done so, it needs to read and parse the package.json file. If we really only need to support reading the scripts, I would go ahead and make a helper function which takes the file path, parses the scripts, and generates the script env vars, so you can just keep an object with the script environment variables. It seems that rather than hold onto the contents of package.json or the script values, it could instead just write the environment variables. If it does this, then later checks would just find them defined and do nothing. That has the advantage of not really holding anything extra.

akrantz (changes requested):
See comments. 

Also, you shouldn't add unneccessary changes for package-lock.json files. The only package.json file that has a dependency change is office-addin-cli.

packages/office-addin-cli/src/parse_json_script_variable.ts
igor-ribeiiro: I just did those changes. Changed the function name to a more meaningful name and also the code just storing what is needed.

packages/office-addin-cli/src/get_package_json_script.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export async function getPackageJsonScript(variable: string): Promise <string | undefined> {
+    // NPMv7 Support
+    if(process.env.npm_package_json) {
+        if (!jsonScripts) {
+            await parseJsonPackageToJsonScripts(process.env.npm_package_json);
+        }
+
+        variable.replace(/_/g, "-"); // Replacing all underscore with hyphens
```
akrantz, line 19: move the replace() down to the NPM v6 env lookup.

packages/office-addin-cli/src/get_package_json_script.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+const readJson = require('read-package-json-fast');
+
+let jsonScripts : any = undefined;
+
+async function parseJsonPackageToJsonScripts(jsonPath: string) {
+    jsonScripts = (await readJson(`${jsonPath}`)).scripts;
```
akrantz, line 9: instead of a side effect, have the scripts returned from this function.

packages/office-addin-cli/test/test.ts
```diff
@@ -6,6 +6,7 @@ import * as mocha from "mocha";
 import * as sinon from "sinon";
 import * as log from "../src/log";
 import * as parse from "../src/parse";
+import * as jsonScript from "../src/get_package_json_script";
```
akrantz, line 9: camelCase for the file name is better.

packages/office-addin-cli/test/test.ts
```diff
@@ -74,4 +75,45 @@ describe("office-addin-cli tests", function() {
       });
     });
   });
+
+  describe("get_package_json_script.ts", function() {
+    describe("getPackageJsonScript() for npmv6", function() {
+      it("npmv6 - exists", async function() {
+        process.env["npm_package_scripts_test"] = "npmv6 1";
```
akrantz, line 82: use before/after to keep the process.env unchanged after running the tests.

packages/office-addin-cli/src/getPackageJsonScript.ts
```diff
@@ -0,0 +1,29 @@
...
+async function parseJsonPackageToJsonScripts(jsonPath: string) {
+    return (await readJson(`${jsonPath}`)).scripts;
+}
+
+/**
+ * Gets a script command from package json during a npm execution
+ * @param scriptName The name of the script
+ */
+export async function getPackageJsonScript(scriptName: string): Promise <string | undefined> {
+    // NPMv7 Support
```
akrantz, line 17: if (!scriptName) {
   return undefined;
}

packages/office-addin-cli/src/getPackageJsonScript.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (process.env.npm_package_json) {
+        if (!jsonScripts) {
+            jsonScripts = await parseJsonPackageToJsonScripts(process.env.npm_package_json);
+        }
+
+        return jsonScripts[scriptName];
+    }
+
+    // NPMv6 Support
+    scriptName.replace(/-/g, "_"); // Replacing all hyphens with underscores
```
akrantz, line 27: I mentioned in the previous comment to add a check at the top to handle when scriptName is undefined/null.

packages/office-addin-cli/test/test.ts
```diff
@@ -74,4 +75,59 @@ describe("office-addin-cli tests", function() {
...
+    });
+
+    describe("getPackageJsonScript() for npmv7", function() {
+      let npm_package_json_backup: string | undefined;
+      before(async function() {
+        npm_package_json_backup = process.env.npm_package_json;
+      });
+      it("npmv7 - no hyphen", async function() {
+        process.env.npm_package_json = "./test/test.json";
+        assert.strictEqual(await jsonScript.getPackageJsonScript("test"), "1");
```
akrantz, line 106: You could add test cases for the script name being: undefined, ""

akrantz:
For the package-lock.json files, once #413 is merged, you should remove the package-lock.json files changed, and then regenerate the updated package-lock.json files for the package.json changes.

igor-ribeiiro:
> For the package-lock.json files, once #413 is merged, you should remove the package-lock.json files changed, and then regenerate the updated package-lock.json files for the package.json changes.

Just did that!

akrantz (approved):

PR #415, opened 2021-03-28
fix getPackageScript() to handle hyphen in name for NPM v6

akrantz:
Also made improvements to have better type info and tests.


packages/office-addin-cli/src/npmPackage.ts
```diff
@@ -0,0 +1,36 @@
...
+
+let scripts: ScriptsObject | undefined;
+
+export function clearCachedScripts() {
+    scripts = undefined;
+}
+
+async function readPackageJsonScripts(jsonPath: string) {
+    const packageJson = await readPackageJson(`${jsonPath}`);
+    return packageJson.scripts || {};
```
akrantz, line 14: If package.json doesn't contain "scripts", it will ensure an empty object is returned.

packages/office-addin-cli/src/npmPackage.ts
```diff
@@ -0,0 +1,36 @@
...
+    // NPM v7 Support
+    if (process.env.npm_package_json) {
+        if (!scripts) {
+            scripts = await readPackageJsonScripts(process.env.npm_package_json);
+        }
+        return scripts[scriptName];
+    }
+
+    // NPM v6 Support
+    return process.env[`npm_package_scripts_${scriptName.replace(/-/g, "_")}`];
```
akrantz, line 35: The fix was to use the return value of replace().

packages/office-addin-cli/src/read-package-json-fast.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "read-package-json-fast" {
+    declare type ScriptsObject =  { [key: string]: string };
+    export default function readPackageJson(filePath: string): Promise<{ scripts: ScriptsObject, [key: string]: unknown}>;
```
akrantz, line 3: This provides better type info. The parsed package.json will return an object containing a scripts property, which is a an object containing keys and values that are strings. The object can contain other string properties with any type of property.

packages/office-addin-cli/test/test.json
```diff
@@ -1,10 +1,10 @@
 {
     "scripts": {
-      "test": "1",
-      "test-test": "2",
-      "test_test": "3",
-      "test test": "4",
-      "test-tes-te": "5"
+      "one": "1",
```
akrantz, line 3: There was a problem with the test using "test" for the script name. It turns out that when "npm test" is run, it was picking up the value from office-addin-cli package.json which contains a script named "test". So I changed to use names which would not exist.

packages/office-addin-cli/src/main.ts
```diff
@@ -2,5 +2,5 @@
 // Licensed under the MIT license.
 
 export * from "./log";
+export * from "./npmPackage";
```
akrantz, line 5: I renamed to "npmPackage.ts" since it is a more concise and description name for the category of functionality.

packages/office-addin-cli/src/npmPackage.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import readPackageJson, { ScriptsObject } from "read-package-json-fast";
+
+let scripts: ScriptsObject | undefined;
```
akrantz, line 6: Try not to use "any". You can use "unknown", but it is better to be able to describe type info more accurately. Here I was able to define ScriptsObject as the type, which is an object whose keys and values are strings.

igor-ribeiiro: Nice! I found a solution in the internet that was using any, but this ScriptsObject is way better!

packages/office-addin-cli/src/read-package-json-fast.d.ts
igor-ribeiiro: I have never created a .d.ts file, I thought they were just auto generated, I will search and learn more about what this file does.

packages/office-addin-cli/tsconfig.json
```diff
@@ -3,6 +3,7 @@
         "target": "es6",
         "module": "commonjs",
         "moduleResolution": "node",
+        "esModuleInterop": true,
```
igor-ribeiiro, line 6: What does this config option do?

igor-ribeiiro (approved):

packages/office-addin-cli/src/read-package-json-fast.d.ts
akrantz: JavaScript has no type info by default. TypeScript adds type info which helps catch errors during compilation rather than at runtime. The d.ts files can be generated by the TypeScript compiler, because after compilation the .js files that are output have no type info, and the type info from the original .ts files can be output to .d.ts. 

Since read-package-json-fast does not have type info, I've created read-package-json-fast to provide that type info.
https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html

packages/office-addin-cli/tsconfig.json
akrantz: See https://www.typescriptlang.org/tsconfig#esModuleInterop. This isn't required as part of these changes, but it is better if this is set to true. 

With TypeScript code, you shouldn't use `const package = require("package")`. There are a few forms of import statements that can apply depending on how the package exports its modules. You can read about modules here: https://www.typescriptlang.org/docs/handbook/modules.html.

PR #418, opened 2021-03-29
Update lerna to v4.0.0

akrantz:


millerds:
Looks like I did this later without realizing you already had it in progress . . . we don't need this anymore.

igor-ribeiiro:
> Looks like I did this later without realizing you already had it in progress . . . we don't need this anymore.

@millerds Should we close this PR then?

PR #420, opened 2021-03-30
handle modules with no _name

andrewkittredge:
I do not know why they have no name.


akrantz:
This code is different now and this change is no longer applicable.

PR #422, opened 2021-04-07
Use loader to inject function associations

### LINKED ISSUE #421: custom-functions-metadata-plugin triggering deprecation warnings on Webpack 5
Just a heads up from the future. In my setup with Webpack 5, I'm getting these warnings:

```
(node:8984) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
...
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
        Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
        Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
```

Plus, this piece of code:

```
                if (module._source && module._source._name.endsWith(inputFilePath)) {
                    generateResult.associate.forEach((item: { id: string; functionName: string; }) => {
                        module._source._value += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
                    });
                }
```

With webpack 5 on Windows, the condition never passes and the `CustomFunctions.associate` calls are never added to the output. Probing around, it looks like `module._source._name` is morphed into `webpack://data:text/javascript,__webpack_public_path__ = __webpack_base_uri__ = htmlWebpackPluginPublicPath;` by the time that callback is involved.

villasv:
~This PR is the seed for solving https://github.com/OfficeDev/Office-Addin-Scripts/issues/421. I mean seed because I couldn't test it. I wasn't able to setup the project on my machine (running npm ci fails), so I tried my best to guess some implementation details. I hope you're able to test it and fix whatever is still missing.~

This PR fixes #421.

I followed [this recommendation](https://github.com/webpack/webpack/issues/10081) and moved the code injection to a custom loader. I also updated the hooks and taps to not rely on deprecated APIs.

### DIFF
--- a/packages/custom-functions-metadata-plugin/package.json
+++ b/packages/custom-functions-metadata-plugin/package.json
@@ -26,9 +26,11 @@
   "dependencies": {
     "custom-functions-metadata": "^1.0.48",
     "es6-promise": "^4.2.8",
+    "loader-utils": "^2.0.0",
     "webpack": "^5.24.4"
   },
   "devDependencies": {
+    "@types/loader-utils": "^2.0.2",
     "@types/node": "^12.12.34",
     "concurrently": "^6.0.0",
     "rimraf": "^3.0.2",
--- a/packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
+++ b/packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
@@ -9,7 +9,6 @@ const pluginName = "CustomFunctionsMetadataPlugin";
 
 type Options = {input: string, output: string};
 
-
 class CustomFunctionsMetadataPlugin {
     private options: Options;
 
--- a/packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
+++ b/packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
@@ -2,7 +2,6 @@
 // Licensed under the MIT license.
 
 import { generateCustomFunctionsMetadata, IGenerateResult } from "custom-functions-metadata";
-import * as fs from "fs";
 import * as path from "path";
 import { Compiler, sources, WebpackError, NormalModule } from "webpack";
 
@@ -10,13 +9,6 @@ const pluginName = "CustomFunctionsMetadataPlugin";
 
 type Options = {input: string, output: string};
 
-declare global {
-    namespace NodeJS {
-        interface Global {
-            generateResults: Record<string, IGenerateResult> | undefined;
-        }
-    }
-}
 
 class CustomFunctionsMetadataPlugin {
     private options: Options;
@@ -25,9 +17,9 @@ class CustomFunctionsMetadataPlugin {
         this.options = options;
     }
 
+    public static generateResults: Record<string, IGenerateResult> = {};
+
     public apply(compiler: Compiler) {
-        const outputPath: string = (compiler.options && compiler.options.output) ? compiler.options.output.path || "" : "";
-        const outputFilePath = path.resolve(outputPath, this.options.output);
         const inputFilePath = path.resolve(this.options.input);
         let generateResult: IGenerateResult;
 
@@ -40,10 +32,7 @@ class CustomFunctionsMetadataPlugin {
                 generateResult.errors.forEach((err: string) => compilation.errors.push(new WebpackError(inputFilePath + " " + err)));
             } else {
                 compilation.assets[this.options.output] = new sources.RawSource(generateResult.metadataJson);
-                if (!global.generateResults) {
-                    global.generateResults = {};
-                }
-                global.generateResults[this.options.input] = generateResult;
+                CustomFunctionsMetadataPlugin.generateResults[this.options.input] = generateResult;
             }
 
             NormalModule.getCompilationHooks(compilation).beforeLoaders.tap(pluginName, (_, module) => {
@@ -62,4 +51,4 @@ class CustomFunctionsMetadataPlugin {
     }
 }
 
-module.exports = CustomFunctionsMetadataPlugin;
+export = CustomFunctionsMetadataPlugin;
--- a/packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
+++ b/packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
@@ -4,11 +4,19 @@
 import { generateCustomFunctionsMetadata, IGenerateResult } from "custom-functions-metadata";
 import * as fs from "fs";
 import * as path from "path";
-import { Compiler, sources, WebpackError } from "webpack";
+import { Compiler, sources, WebpackError, NormalModule } from "webpack";
 
 const pluginName = "CustomFunctionsMetadataPlugin";
 
 type Options = {input: string, output: string};
+
+declare global {
+    namespace NodeJS {
+        interface Global {
+            generateResults: Record<string, IGenerateResult> | undefined;
+        }
+    }
+}
 
 class CustomFunctionsMetadataPlugin {
     private options: Options;
@@ -23,24 +31,31 @@ class CustomFunctionsMetadataPlugin {
         const inputFilePath = path.resolve(this.options.input);
         let generateResult: IGenerateResult;
 
-        compiler.hooks.compile.tap(pluginName, async () => {
+        compiler.hooks.beforeCompile.tapPromise(pluginName, async () => {
             generateResult = await generateCustomFunctionsMetadata(inputFilePath, true);
         });
 
-        compiler.hooks.emit.tap(pluginName, (compilation) => {
+        compiler.hooks.compilation.tap(pluginName, (compilation) => {
             if (generateResult.errors.length > 0) {
                 generateResult.errors.forEach((err: string) => compilation.errors.push(new WebpackError(inputFilePath + " " + err)));
             } else {
                 compilation.assets[this.options.output] = new sources.RawSource(generateResult.metadataJson);
+                if (!global.generateResults) {
+                    global.generateResults = {};
+                }
+                global.generateResults[this.options.input] = generateResult;
             }
-        });
 
-        compiler.hooks.compilation.tap(pluginName, (compilation, params) => {
-            compilation.moduleTemplates.javascript.hooks.render.tap(pluginName, (source : any, module : any) => {
-                if (module._source && module._source._name.endsWith(inputFilePath)) {
-                    generateResult.associate.forEach((item: { id: string; functionName: string; }) => {
-                        module._source._value += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
-                    });
+            NormalModule.getCompilationHooks(compilation).beforeLoaders.tap(pluginName, (_, module) => {
+                if (module.userRequest.endsWith(inputFilePath)) {
+                    module.loaders.push({
+                        loader: require.resolve("./loader.js"),
+                        options: {
+                            input: this.options.input,
+                        },
+                        ident: null,
+                        type: null,
+                    })
                 }
             });
         });
--- a/packages/custom-functions-metadata-plugin/src/loader.ts
+++ b/packages/custom-functions-metadata-plugin/src/loader.ts
@@ -1,13 +1,16 @@
 import { getOptions } from "loader-utils";
 import { IAssociate } from "custom-functions-metadata";
+import CustomFunctionsMetadataPlugin from "./customfunctionsplugin";
 
-module.exports = function addFunctionAssociations(this: any, source: string): string {
+function addFunctionAssociations(this: any, source: string): string {
     const input = getOptions(this).input as string;
-    if (!global.generateResults || !global.generateResults[input]) {
+    if (!CustomFunctionsMetadataPlugin.generateResults[input]) {
         return source;
     }
-    global.generateResults[input].associate.forEach((item: IAssociate) => {
+    CustomFunctionsMetadataPlugin.generateResults[input].associate.forEach((item: IAssociate) => {
         source += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
     });
     return source;
 }
+
+export = addFunctionAssociations;
--- a/packages/custom-functions-metadata-plugin/src/loader.ts
+++ b/packages/custom-functions-metadata-plugin/src/loader.ts
@@ -1,0 +1,13 @@
+import { getOptions } from "loader-utils";
+import { IAssociate } from "custom-functions-metadata";
+
+module.exports = function addFunctionAssociations(this: any, source: string): string {
+    const input = getOptions(this).input as string;
+    if (!global.generateResults || !global.generateResults[input]) {
+        return source;
+    }
+    global.generateResults[input].associate.forEach((item: IAssociate) => {
+        source += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
+    });
+    return source;
+}
--- a/packages/office-addin-dev-settings/src/registry.ts
+++ b/packages/office-addin-dev-settings/src/registry.ts
@@ -184,7 +184,7 @@ export async function getStringValue(key: RegistryKey, value: string): Promise<s
 }
 
 export async function getValue(key: RegistryKey, value: string): Promise<RegistryValue | undefined> {
-  return new Promise<RegistryValue>((resolve, reject) => {
+  return new Promise<RegistryValue | undefined>((resolve, reject) => {
     const onError = (err: any, item?: winreg.RegistryItem) => {
       if (err) {
         resolve(undefined);
--- a/packages/custom-functions-metadata-plugin/package-lock.json
+++ b/packages/custom-functions-metadata-plugin/package-lock.json
@@ -91,6 +91,12 @@
         "regenerator-runtime": "^0.13.4"
       }
     },
+    "@types/anymatch": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/@types/anymatch/-/anymatch-1.3.1.tgz",
+      "integrity": "sha512-/+CRPXpBDpo2RK9C68N3b2cOvO0Cf5B9aPijHsoDQTHivnGSObdOF2BRQOYjojWTDy6nQvMjmqRXIxH55VjxxA==",
+      "dev": true
+    },
     "@types/eslint": {
       "version": "7.2.7",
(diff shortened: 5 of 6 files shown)


packages/custom-functions-metadata-plugin/package.json
```diff
@@ -26,7 +26,8 @@
   "dependencies": {
     "custom-functions-metadata": "^1.0.48",
     "es6-promise": "^4.2.8",
-    "webpack": "^5.24.4"
+    "webpack": "^5.24.4",
+    "loader-utils": "*",
```
igor-ribeiiro, line 30: This ```,``` needs to be removed. Hopefully you can build the project by now to detect those kind of errors

packages/custom-functions-metadata-plugin/src/loader.js
```diff
@@ -0,0 +1,9 @@
+const { getOptions } = require('loader-utils');
+
+module.exports = function addFunctionAssociations(source) {
```
igor-ribeiiro, line 3: Why is this function needed? It was created but it isn't called anywhere.

packages/custom-functions-metadata-plugin/src/loader.js
```diff
@@ -0,0 +1,9 @@
+const { getOptions } = require('loader-utils');
```
igor-ribeiiro, line 1: It would be great for this file to be of .ts typescript to be consistent with the rest of the project.
Just need to change .js to .ts and add the type for the variables.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -23,24 +23,25 @@ class CustomFunctionsMetadataPlugin {
...
 
-        compiler.hooks.compilation.tap(pluginName, (compilation, params) => {
-            compilation.moduleTemplates.javascript.hooks.render.tap(pluginName, (source : any, module : any) => {
-                if (module._source && module._source._name.endsWith(inputFilePath)) {
-                    generateResult.associate.forEach((item: { id: string; functionName: string; }) => {
-                        module._source._value += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
-                    });
+            NormalModule.getCompilationHooks(compilation).beforeLoaders.tap(pluginName, (_, module) => {
+                if (module.userRequest.endsWith(inputFilePath)) {
+                    module.loaders.push({
```
igor-ribeiiro, line 39: The following properties are missing from this object: ident, type

igor-ribeiiro (changes requested):
The project lock files got corrupted and I just sent a PR: [https://github.com/OfficeDev/Office-Addin-Scripts/pull/423](url) updating then, so you should be able to install and compile the project right after it is approved and merged.
Sorry for it not working beforehand.

igor-ribeiiro:
@villasv it might take a little longer to fix the install issue, but a quick workaround is to simply delete the package-lock.json file from custom-functions-metadata-plugin and install again and that fixes the problem

villasv:
No problem, I think we'll be able to work in parallel. After deleting all package-locks I was able to `npm install`. Once the package-lock fixes come in, I'll rebase and wrap up.

packages/custom-functions-metadata-plugin/src/loader.js
villasv: It defines the loader, which is passed directly as a file reference here: https://github.com/OfficeDev/Office-Addin-Scripts/pull/422/files#diff-addd1af0a602b17de45903dbd9a9d3726d67753c6ac81a238e31dc21361860bdR45

This could be an anonymous export too, though. I just named it out of habit.

villasv:
I was able to build the entire project, though I had to fix an extra bit: https://github.com/OfficeDev/Office-Addin-Scripts/pull/422/files#diff-dc5de9d974097b759edb365814ad4bc3236ea051eebe52498eea83e0eaad1f22R187

I'm not adding the package-lock.json files to the PR because I'm guessing they'll have to be fixed on your end first.

villasv:
It was a bit difficult to test (had to copy built files into my project manually), but I can confirm that the plugin works with my Webpack 5 project. I get no warnings and the `CustomFunction.associate` calls are added as intended.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -4,12 +4,16 @@
...
 import * as path from "path";
-import { Compiler, sources, WebpackError } from "webpack";
+import { Compiler, sources, WebpackError, NormalModule } from "webpack";
 
 const pluginName = "CustomFunctionsMetadataPlugin";
 
 type Options = {input: string, output: string};
 
+declare global {
+    var generateResults: Record<string, IGenerateResult> | undefined;
```
igor-ribeiiro, line 14: Use ```let``` here instead of ```var```  to be more consistent with ES6 conventions

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -23,24 +27,29 @@ class CustomFunctionsMetadataPlugin {
...
             generateResult = await generateCustomFunctionsMetadata(inputFilePath, true);
         });
 
-        compiler.hooks.emit.tap(pluginName, (compilation) => {
+        compiler.hooks.compilation.tap(pluginName, (compilation) => {
             if (generateResult.errors.length > 0) {
                 generateResult.errors.forEach((err: string) => compilation.errors.push(new WebpackError(inputFilePath + " " + err)));
             } else {
                 compilation.assets[this.options.output] = new sources.RawSource(generateResult.metadataJson);
+                if (!global.generateResults) global.generateResults = {};
```
igor-ribeiiro, line 39: Let's indent this to be like:
```
if (!global.generateResults) { 
    global.generateResults = {};
}
```
To be more consistent with the rest of the codebase.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -23,24 +27,29 @@ class CustomFunctionsMetadataPlugin {
...
-        compiler.hooks.compilation.tap(pluginName, (compilation, params) => {
-            compilation.moduleTemplates.javascript.hooks.render.tap(pluginName, (source : any, module : any) => {
-                if (module._source && module._source._name.endsWith(inputFilePath)) {
-                    generateResult.associate.forEach((item: { id: string; functionName: string; }) => {
-                        module._source._value += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
-                    });
+            NormalModule.getCompilationHooks(compilation).beforeLoaders.tap(pluginName, (_, module) => {
+                if (module.userRequest.endsWith(inputFilePath)) {
+                    module.loaders.push({
+                        loader: require.resolve("./loader.js"),
```
igor-ribeiiro, line 46: It is weird that this works, given that the filename is ```loader.ts```. Could it be changed to it or not?

packages/custom-functions-metadata-plugin/src/loader.ts
```diff
@@ -0,0 +1,11 @@
+import { getOptions } from "loader-utils";
+import { IAssociate } from "custom-functions-metadata";
+
+module.exports = function addFunctionAssociations(this: any, source: string): string {
+    const input = getOptions(this).input as string;
+    if (!global.generateResults || !global.generateResults[input]) return source;
```
igor-ribeiiro, line 6: Let's indent this to be like:
```
if (!global.generateResults || !global.generateResults[input]) {
    return source;
}
```
To be more consistent with the rest of the codebase.

igor-ribeiiro (changes requested):
Just a few minor requests to follow some conventions.
Tested the code on webpackv4 and it is working

igor-ribeiiro:
The tests are currently failing because the project wasn't installing properly. Let's wait for https://github.com/OfficeDev/Office-Addin-Scripts/pull/424 to be pushed and then we can merge this PR

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
villasv: It could not. It works because this only runs inside a Webpack build process, which will have this plugin installed with its transpiled files. So at runtime, `loader.ts` won't exist.

packages/office-addin-dev-settings/src/registry.ts
```diff
@@ -184,7 +184,7 @@ export async function getStringValue(key: RegistryKey, value: string): Promise<s
 }
 
 export async function getValue(key: RegistryKey, value: string): Promise<RegistryValue | undefined> {
-  return new Promise<RegistryValue>((resolve, reject) => {
+  return new Promise<RegistryValue | undefined>((resolve, reject) => {
```
akrantz, line 187: Why is `| undefined` being added here?

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -4,12 +4,16 @@
...
 import * as fs from "fs";
 import * as path from "path";
-import { Compiler, sources, WebpackError } from "webpack";
+import { Compiler, sources, WebpackError, NormalModule } from "webpack";
 
 const pluginName = "CustomFunctionsMetadataPlugin";
 
 type Options = {input: string, output: string};
 
+declare global {
```
akrantz, line 13: This is to have a global map of results for each inputFile, right?  Does `declare global` need to be used? (I don't recall what that does off the top of my head.) I was thinking this could be a const object which is initialized to an empy object (rather than line 40 below) and then line 42 would put the results into the object.

akrantz (changes requested):
See comments.

packages/office-addin-dev-settings/src/registry.ts
villasv: Because 3 lines below there's `resolve(undefined);`

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
villasv: I though this would add the `generateResults` prop to the `typeof global` interface, but it looks like I was mistaken.  I'm going to change this into a declaration merging into the Global interface.

villasv: I wasn't able to avoid using `declare global` (which I got from [the docs here](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#global-augmentation)). Maybe we could avoid it if we create a dedicated `.d.ts` file?

villasv:
I rebased and squashed commits. Also merged into the Global interface instead of declaring a variable, it looks a bit more self-explanatory.

packages/office-addin-dev-settings/src/registry.ts
akrantz: OK, this is fine. Sorry, I was reviewing too quickly and didn't see the outer context was returning undefined.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
akrantz: I was thinking this could just be 

`const generateResults: { [key: string], IGenerateResult } = { };`

but I could be mistaken.

villasv: We have to insert it on the `global` interface so we can store it like this:

```
global.generateResults[this.options.input] = generateResult;
```

So we can use it from the loader:
```
    global.generateResults[input].associate.forEach((item: IAssociate) => {
        source += `\nCustomFunctions.associate("${item.id}", ${item.functionName});`;
    });
```

Adding it to the global felt pretty hacky, but I just couldn't find a different way to share data between the plugin and the loader (which is required dynamically and its options are all `string | boolean | null`) without serialization.

Now given that this has to be a global, AFAIK the only way to patch the `Global` interface is either using `declare global` + interface merging or a custom `.d.ts` file.

villasv: As an alternative, we could just make the loader build `generateResults` again. It would be done twice: at the plugin to generate the JSON metadata; at the loader to generate the associate calls.

I'd be fine with that too, but I don't know if that would be too much of a performance tax. Of even if a small performance penalty, if it's worth it to avoid using a global var.

akrantz: No, we don't want to do that. I'm still not clear why you can't just declare a const variable.

villasv: If we declare a const variable inside `customfunctionsplugin.ts`, will the function inside `loader.ts` be able to access its value?

igor-ribeiiro (approved):

akrantz:
@igor-ribeiiro This needs to be tested with Office-Addin-TaskPane template (using webpack 4).  I also wonder if we could just move the templates to use webpack 5.

igor-ribeiiro:
@akrantz I tested that these changes work with webpackv4 on Excel Custom Functions

akrantz:
Please also verify with Office-Addin-TaskPane.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
akrantz: Sorry I haven't had time to try that out. It was just a suggestion to see if that could be simplified, but I don't think it has to block the PR. If you're able to try it and get it to work, that would be great. If not, I think it's OK.

igor-ribeiiro:
> Please also verify with Office-Addin-TaskPane.

Office-Addin-Taskpane doesn't use custom-functions package, so it shouldn't matter to them. I tried to test it before but when I discovered this there wasn't a test to be made.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
villasv: Tried it, couldn't make it work

```
[0] src/customfunctionsplugin.ts(37,29): error TS2339: Property 'generateResults' does not exist on type 'Global'.      
[0] src/customfunctionsplugin.ts(38,28): error TS2339: Property 'generateResults' does not exist on type 'Global'.      
[0] src/customfunctionsplugin.ts(40,24): error TS2339: Property 'generateResults' does not exist on type 'Global'.      
[0] src/loader.ts(6,17): error TS2339: Property 'generateResults' does not exist on type 'Global'.
[0] src/loader.ts(6,44): error TS2339: Property 'generateResults' does not exist on type 'Global'.
[0] src/loader.ts(9,12): error TS2339: Property 'generateResults' does not exist on type 'Global'.
```

villasv: If I just remove `global` references and try to get that variable from the scope, it also fails
```
[0] src/loader.ts(6,10): error TS2304: Cannot find name 'generateResults'.
[0] src/loader.ts(6,30): error TS2304: Cannot find name 'generateResults'.
[0] src/loader.ts(9,5): error TS2304: Cannot find name 'generateResults'.
```

akrantz: I have a change which compiles.

akrantz: image
image

I could push a commit with this changes if they look OK to you.

akrantz: I didn't try it out.

akrantz (changes requested):
Waiting to hear back whether my suggested change to avoid the global will work...

akrantz:
Sorry @igor-ribeiiro, my last comment was dumb. Of course, this only needs to work with Excel-Custom-Functions.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
villasv: Ah, I see. You mean to export the variable. Because the plugin I used as base didn't do it, I assumed I shouldn't do it (it's obviously simpler). It's an hypothesis weakly held, though. I'll test it soon.

villasv: We've got a situation here.
```
module.exports = CustomFunctionsMetadataPlugin;
```
This prevents us from exporting anything else. We can either change this into a named export (breaking change) or expose the `generateResults` as a static member.

akrantz: OK, I think you can make it a static member.

villasv:
I converted the `generateResults` into a static member, but to import the CustomFunctionsPlugin class from `loader.ts`, I had to convert its export syntax from `module.exports =` into `export =` and activate the `esModuleInterop` config.

akrantz:
Thanks! I am not an expert on modules, but I think this should be OK. It won't be a breaking change, right?

villasv:
No breaking change. `export = ` transpiles into `module.exports = ` too, it's just TypeScript's syntax for it and carries the type information.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
```diff
@@ -2,49 +2,53 @@
...
-import * as fs from "fs";
 import * as path from "path";
-import { Compiler, sources, WebpackError } from "webpack";
+import { Compiler, sources, WebpackError, NormalModule } from "webpack";
 
 const pluginName = "CustomFunctionsMetadataPlugin";
 
 type Options = {input: string, output: string};
 
+
```
akrantz, line 12: minor, but there's an extra line here now that can be removed.

akrantz (approved):
Thanks so much for taking the time to work through this.

There's one minor suggestion about an extra blank line that can be removed.

@igor-ribeiiro Please validate that this is working well for you before merging. Thanks.

packages/custom-functions-metadata-plugin/src/customfunctionsplugin.ts
villasv: Oops, missed that one when deleting the interface

villasv, line 12:
```suggestion
```

igor-ribeiiro (approved):

PR #423, opened 2021-04-09
Updating lock json  for CF metadataplugin

igor-ribeiiro:
The project wasn't installing with the current lock json files. 
After the updates the project is now installing successfully.

Running "npm install" in the custom-functions-metadata-package directory results in the error message:
```ERR! Cannot read property 'match' of undefined```

It appears that the ```package-lock.json``` file is corrupt and needs to be regenerated.
To fix this, I deleted the ```package-lock.json``` file and ran ```npm install``` to generate it again.


akrantz (changes requested):
You need to provide clearer details about the problem and why this is a solution. Which project are you referring to when you say it "doesn't install". This Office-Addin-Scripts repository, or a Yo Office project using this. Which version of Node and NPM are you using?

It is nearly impossible to review the diffs of package-lock.json changes.  Why do these changes fix the problem? What is the problem that it is fixing?

igor-ribeiiro:
> You need to provide clearer details about the problem and why this is a solution. Which project are you referring to when you say it "doesn't install". This Office-Addin-Scripts repository, or a Yo Office project using this. Which version of Node and NPM are you using?
> 
> It is nearly impossible to review the diffs of package-lock.json changes. Why do these changes fix the problem? What is the problem that it is fixing?

Sure, sorry for not providing that many details.
The problem is that this project Office-Addin-Scripts doesn't build. The package that has the problem is the custom functions metadata plugin. When using npm install, it throws an error in lock-json file and simply updating it solves this issue.

akrantz:
Did you check the other packages as well?

igor-ribeiiro:
> Did you check the other packages as well?

I'm thinking this change alone didn't fix it. When I retry the npm install it gives me the same error again!
I will do some more testing and investigate this bug, for now I will make this PR a draft.

PR #424, opened 2021-04-10
Fix pinned versions and remove extraneous pkgs

villasv:
> This PR duplicates the attempt of #423, while instead of doing a complete re-install of all packages, I try to change only the offending lockfiles. I though it was going to be a small fix, but in the end, it was pretty big... so after all I'd say it's probably for the best to just nuke all `package-lock.json`s and `npm install` everything anew. At least now we can know the particular packages that were causing issues, though.

In the current state of the `master` branch, running `npm ci` for the first time will generate errors like this:
```
ERR! Cannot read property 'match' of undefined
```
After examining the pinned versions and lockfiles for the packages `custom-function-metadata` and `custom-function-metadata-plugin` (using [this tip](https://stackoverflow.com/a/64369937/3935325)), I noticed a few inconsistencies. Some packages in the lockfile were not really asked by package.json, while some others were pinned in a version incompatible with the one specified in the lockfile. I've tried to make atomic commits so they can be closely inspected.

At the very end, I ran `npm ci` one last time and noticed quite a lot of changes to commit. I think it's important to note that they're basically all code removals.

Now I can reliably run `npm ci` and `npm build` without errors in a freshly cloned state.


packages/custom-functions-metadata/package.json
```diff
@@ -29,7 +29,7 @@
     "assert": "^1.5.0",
     "commander": "^6.2.0",
     "nconf": "^0.10.0",
-    "office-addin-cli": "^1.1.1",
```
igor-ribeiiro, line 32: Do we need this downgrade?

packages/custom-functions-metadata/package.json
```diff
@@ -41,7 +41,7 @@
     "@types/xregexp": "^3.0.30",
     "concurrently": "^6.0.0",
     "mocha": "^7.1.1",
-    "office-addin-lint": "^1.1.6",
```
igor-ribeiiro, line 44: And also this one, is it needed?

igor-ribeiiro (changes requested):
I just tested your changes and indeed it works! Thanks for the contributions!
I just made some few questions regarding to some version downgrades that you've made and that's it.

akrantz:
I think one part of the problem is that office-addin-cli was pointing to older versions for some of the packages. Please make sure that all of the packages in the project which have dependencies on other packages in the project are pointing to the latest version.

packages/custom-functions-metadata/package.json
villasv: We don't, I just wanted to make the PR as readable as possible, so I chose to make package.json consistent with what's on package-lock.json, not the other way around. I'll change this to latest and update the branch.

packages/custom-functions-metadata/package.json
villasv: Same as the other.

villasv:
@akrantz could be, but there were parts of package-lock.json that were simply corrupted (like [this](https://github.com/OfficeDev/Office-Addin-Scripts/pull/424/commits/e2c5abbe4badaea334d85e58834f1071541c8bf1#diff-9a76a31f79289612798064331ebdc77a4746b23c01958f20477c2696b7e5085aL8746)). I can't say for sure what caused this situation, but I'm guessing there must have been a merge conflict accident somewhere.

I'll update lint and cli to latest and see how it goes.

villasv:
So, here's the thing. This project uses lerna, which means that packages that have other packages as dependencies don't have to install them via npm. Lerna takes care of symlinking them together, which is why at the current state of the PR, the project installs and builds fine.

But, each of your packages is also publicly available on npm with the same name. This means that you can manually go inside one of the packages and `npm i office-addin-cli@latest`, and it will pass over lerna to pick up that dependency from the npm registry.

This is why the PR is currently deleting 70k LOC, I'm trusting lerna to resolve dependencies, but I'm not sure if that's what you really want (because the previous state of the repo was fecthing from the public registry), so I'm asking for guidance on it.

akrantz:
Thanks for working through this. I think with lerna the packages should have dependencies on the other packages. "npm install" in the root has a post-install script to run "lerna bootstrap" which symlinks packages in the repo together. But I think each package.json should have a dependency for other packages that it uses from office-addin-scripts repo. Whenever you do "npm install" operations in each package, be sure to then run "npm install" in the root afterward.

villasv:
Ok, if I understood correctly, you would rather have lerna symlinks between packages instead of fetching sister packages from the public registry, so I've updated the package-locks accordingly. Project is installing and building normally for me.

But you do have a few npm scripts in the root package.json that will undo this. `npm install-packages`  will perform `npm install` on each package individually, and so npm will fetch sister packages from the public registry instead of using symlinks. I'd recommend removing these and instructing contributors to rely only on running `npm ci` on the root package.

villasv:
Oops, sorry, I forgot to commit one file 🤦 . Here it goes.

akrantz:
Sorry if this was unclear. Each package should have package.json dependencies on other packages, just like if lerna is not used. The symlinking is only to make it easier to work with changes inside the monorepo before publish. On publish, the packages will have dependencies on one another just like normal.

office-addin-cli is used by most of the other packages. So each package.json that uses it should have a dependency on it which should match the version in office-addin-cli package.json.

akrantz:
I'm confused by the last commit subject. The packages in the mono-repo should be in the package-lock.json file. You should be able to run "npm run install-packages" from the root to ensure it runs "npm install" for each package, and then runs "npm install" in the root.

akrantz:
It appears that office-addin-usage-data has an incorrect version in its dependency on office-addin-cli. 

image

villasv:
I think I'm the one not being sufficiently clear 😅 

Yes, each package has sister packages listed in package.json as dependencies. This is what lerna uses to create symlinks, so we're on the same page here.

What I'm asking is if sister packages should also be listed in package-lock.json. This isn't required, because lerna won't update package-lock.json by default. The current state of the PR relies on lerna symlinks to not have sister packages listed on package-lock.json.

This isn't default behavior for lerna repos because pretty often, some packages are no publicly available, so it's not even possible to add them to package-lock.json. But because all your packages are public, it is indeed possible to have them inside the package-lock.json file. This is totally optional. If you do this, you might get warnings like these:

`lerna WARN EREPLACE_EXIST office-addin-prettier-config is already installed for custom-functions-metadata. Replacing with symlink...`

akrantz:
The sister packages SHOULD be in package-lock.json.  The warning about replacing with symlinks is OK.

villasv:
> > It appears that office-addin-usage-data has an incorrect version in its dependency on office-addin-cli.

Indeed, but that's also the case on the master branch. Its version is so old that I thought it was intentional. I'll update it too.

> The sister packages SHOULD be in package-lock.json. The warning about replacing with symlinks is OK.

Okay, then I'll fixup lockfiles to this pattern at once.

villasv:
All done. No errors on `npm ci`, `npm run exec -- npm ci & npm ci` or `npm run build`.

igor-ribeiiro (approved):

PR #425, opened 2021-04-12
office-addin-manifest: update package-lock.json file

akrantz:
With the fix to the package-lock.json files, after cleaning my node_modules folders and doing "npm install" for each package and then the root, I see one update for the office-addin-manifest lock file.


akrantz:
After doing "npm install" in the packages\office-addin-manifest folder, and then at the root, there are no changes needed to the lock file. Closing.

PR #428, opened 2021-04-15
Making File Paths more Strictly

igor-ribeiiro:
What was done:
- Added some Expected Errors on some errors that were missing.
- Adding better debug message to `Sideload is not supported` error.
- Changing the `maskFilePaths` function to trim absolute files up to extension.
  - Before: C:\path\filename.ext would be trimmed to C:\\filename.ext
  - Now: C:\path\filename.ext would be trimmed to C:\\.ext
  - This needed to be done because sometimes the filename may contain an UID
- Adding more unit tests for this `maskFilePaths` function.

### DIFF
--- a/packages/office-addin-dev-settings/src/appcontainer.ts
+++ b/packages/office-addin-dev-settings/src/appcontainer.ts
@@ -143,7 +143,7 @@ export async function getAppcontainerNameFromManifest(manifestPath: string): Pro
   const sourceLocation = manifest.defaultSettings ? manifest.defaultSettings.sourceLocation : undefined;
 
   if (sourceLocation === undefined) {
-    throw new Error(`The source location could not be retrieved from the manifest.`);
+    throw new ExpectedError(`The source location could not be retrieved from the manifest.`);
   }
 
   return getAppcontainerName(sourceLocation, false);
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -57,7 +57,7 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
   const templatePath = document && document !== "" ? path.resolve(document) : getTemplatePath(app, addInType);
 
   if (!templatePath) {
-    throw new ExpectedError("Sideload is not supported.");
+    throw new ExpectedError(`Sideload is not supported for apptype: ${addInType}.`);
   }
 
   const templateBuffer = await readFileAsync(templatePath);
--- a/packages/office-addin-manifest/src/addInTypes.ts
+++ b/packages/office-addin-manifest/src/addInTypes.ts
@@ -50,7 +50,7 @@ export function parseAddInType(value: string): AddInType {
   const addInType = toAddInType(value);
 
   if (!addInType) {
-    throw new Error(`${value} is not a valid Office add-in type.`);
+    throw new ExpectedError(`${value} is not a valid Office add-in type.`);
   }
 
   return addInType;
--- a/packages/office-addin-manifest/src/addInTypes.ts
+++ b/packages/office-addin-manifest/src/addInTypes.ts
@@ -1,6 +1,8 @@
 // copyright (c) Microsoft Corporation. All rights reserved.
 // licensed under the MIT license.
 
+import { ExpectedError } from "office-addin-usage-data";
+
 /**
  * The types of Office add-ins.
  */
@@ -48,7 +50,7 @@ export function parseAddInType(value: string): AddInType {
   const addInType = toAddInType(value);
 
   if (!addInType) {
-    throw new Error(`${value} is not a valid Office add-in type.`);
+    throw new ExpectedError(`${value} is not a valid Office add-in type.`);
   }
 
   return addInType;
--- a/packages/office-addin-manifest/src/addInTypes.ts
+++ b/packages/office-addin-manifest/src/addInTypes.ts
@@ -50,7 +50,7 @@ export function parseAddInType(value: string): AddInType {
   const addInType = toAddInType(value);
 
   if (!addInType) {
-    throw new ExpectedError(`${value} is not a valid Office add-in type.`);
+    throw new Error(`${value} is not a valid Office add-in type.`);
   }
 
   return addInType;
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -272,11 +272,10 @@ export class OfficeAddinUsageData {
   public maskFilePaths(err: Error): Error {
     try {
       const regexRemoveUserFilePaths = /[\/\\](.*)[\/\\]/gmi;
-      const regexRemoveRelativeUserFilePathsFromStack = /\/(.*)\//gmi;
       const regexRemoveAbsoluteUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gmi;
-
-      err.message = err.message.replace(regexRemoveUserFilePaths, "");
-      err.stack = err.stack.replace(regexRemoveRelativeUserFilePathsFromStack, "");
+      const regexRemoveFirstFilePathFromStack = /\\(.*)\./i;
+      err.message = err.message.replace(regexRemoveUserFilePaths, "\\");
+      err.stack = err.stack.replace(regexRemoveFirstFilePathFromStack, "\\.");
       err.stack = err.stack.replace(regexRemoveAbsoluteUserFilePathsFromStack, "");
       return err;
     } catch (err) {
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -272,11 +272,12 @@ export class OfficeAddinUsageData {
   public maskFilePaths(err: Error): Error {
     try {
       const regexRemoveUserFilePaths = /[\/\\](.*)[\/\\]/gmi;
-      const regexRemoveUserFilePathsFromStack = /\w:[\\\/](?:[^[\\\/]\s]+[\\\/])+/gmi;
+      const regexRemoveRelativeUserFilePathsFromStack = /\/(.*)\//gmi;
+      const regexRemoveAbsoluteUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gmi;
 
       err.message = err.message.replace(regexRemoveUserFilePaths, "");
-      err.stack = err.stack.replace(regexRemoveUserFilePaths, "");
-      err.stack = err.stack.replace(regexRemoveUserFilePathsFromStack, "");
+      err.stack = err.stack.replace(regexRemoveRelativeUserFilePathsFromStack, "");
+      err.stack = err.stack.replace(regexRemoveAbsoluteUserFilePathsFromStack, "");
       return err;
     } catch (err) {
       this.reportError("maskFilePaths", err);
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -271,8 +271,9 @@ export class OfficeAddinUsageData {
    */
   public maskFilePaths(err: Error): Error {
     try {
-      const regexRemoveUserFilePaths = /\/(.*)\//gmi;
-      const regexRemoveUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gmi;
+      const regexRemoveUserFilePaths = /[\/\\](.*)[\/\\]/gmi;
+      const regexRemoveUserFilePathsFromStack = /\w:[\\\/](?:[^[\\\/]\s]+[\\\/])+/gmi;
+
       err.message = err.message.replace(regexRemoveUserFilePaths, "");
       err.stack = err.stack.replace(regexRemoveUserFilePaths, "");
       err.stack = err.stack.replace(regexRemoveUserFilePathsFromStack, "");
--- a/packages/office-addin-usage-data/test/test.ts
+++ b/packages/office-addin-usage-data/test/test.ts
@@ -159,7 +159,7 @@ describe("Test office-addin-usage data-package", function() {
     });
   });
   describe("Test maskFilePaths method", () => {
-    it("should return a parsed file path error", () => {
+    it("should parse error file paths with slashs", () => {
       addInUsageData.setUsageDataOff();
       const compareError = new Error();
       compareError.name = "TestData-test";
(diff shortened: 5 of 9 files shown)


packages/office-addin-dev-settings/src/appcontainer.ts
```diff
@@ -143,7 +143,7 @@ export async function getAppcontainerNameFromManifest(manifestPath: string): Pro
   const sourceLocation = manifest.defaultSettings ? manifest.defaultSettings.sourceLocation : undefined;
 
   if (sourceLocation === undefined) {
-    throw new Error(`The source location could not be retrieved from the manifest.`);
+    throw new ExpectedError(`The source location could not be retrieved from the manifest.`);
```
millerds, line 146: How are we defining what is an expected error and what isn't?

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -57,7 +57,7 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
   const templatePath = document && document !== "" ? path.resolve(document) : getTemplatePath(app, addInType);
 
   if (!templatePath) {
-    throw new ExpectedError("Sideload is not supported.");
+    throw new ExpectedError(`Sideload is not supported for apptype: ${addInType}.`);
```
millerds, line 60: Was it just the apptype we were looking at?  Are there some other pieces of information that would help us better understand the code path that was taken?

Also, do we just have access to the error message or can we add information to the property bags we were looking at?

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -271,11 +271,12 @@ export class OfficeAddinUsageData {
...
   public maskFilePaths(err: Error): Error {
     try {
-      const regexRemoveUserFilePaths = /\/(.*)\//gmi;
-      const regexRemoveUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gmi;
-      err.message = err.message.replace(regexRemoveUserFilePaths, "");
-      err.stack = err.stack.replace(regexRemoveUserFilePaths, "");
-      err.stack = err.stack.replace(regexRemoveUserFilePathsFromStack, "");
+      const regexRemoveUserFilePaths = /[\/\\](.*)[\/\\]/gmi;
+      const regexRemoveAbsoluteUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gmi;
+      const regexRemoveFirstFilePathFromStack = /\\(.*)\./i;
```
millerds, line 276: I'm not clear on what the second two expressions are for.

packages/office-addin-dev-settings/src/appcontainer.ts
igor-ribeiiro: I think there is some subjectivity in this.
But the idea is that ExpectedError = errors that are expected to happen given a specific flow of actions

akrantz: Unexpected errors are things where the code itself has a problem, e.g. our code crashes. If the outcome is because of the inputs from the user, then it is an expected error. We cannot do anything to resolve the error. When we calculate reliability, we can know the overall success/total (ICE %), and we can know the unexpected errors / total (ACE%). We want our ACE% to be 100% ideally, e.g. our code has no issues, and we can monitor ICE% to see if there areas where customers are running into problems preventing them from being successful.

packages/office-addin-dev-settings/src/sideload.ts
akrantz: We shouldn't add addInType to the message here. It's not useful info for the customer. There are 3 add in types, TaskPane, Content, and MailApp. It's not really the add-in type which determines if sideload is supported or not. 

What is the scenario where the templatePath is not being returned from getTemplatePath?

igor-ribeiiro: I think the apptype for this specific error should be enough right? The main problem with this error is that it didn't give enough information about the error (which is this case).

What do you mean by property bags? It is something like additional properties? It is possible to add these when the report errors function is called.

packages/office-addin-usage-data/src/usageData.ts
akrantz: Wouldn't it be better to mask the user folder name, e.g. a path like "c:\users\adamk\..." would be turned into "c:\users\****\..."

I personally cannot read these regexes to know easily what they are doing. Can we please add unit tests for these so we can have tests which show what they do and that it is the desired behavior?

packages/office-addin-usage-data/test/test.ts
```diff
@@ -162,13 +163,20 @@ describe("Test office-addin-usage data-package", function() {
       addInUsageData.setUsageDataOff();
```
akrantz, line 163: This is a poorly written unit test. We should have a test case group for the maskFilePaths() function, and separate test cases. It should follow a pattern of arrange, act, assert. It would be great to see improvements made here.

packages/office-addin-dev-settings/src/appcontainer.ts
akrantz: It's not really subjective. An error which is the result of the inputs from the user are expected errors. We cannot change our code to influence the outcome. An unexpected error is a crash in our code.

packages/office-addin-dev-settings/src/sideload.ts
akrantz: Igor, see my previous comment. It's not the app type which is the issue here but more of a question of why getTemplate didn't return a string.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: The first expression is for using on the error message.

The second and third are for using in the stack of the error. The second is to remove absolute paths from the stack and only keep the internal paths, while the third is to trim the first file path to only have the extension in it (because of UID problems in the filename).

packages/office-addin-dev-settings/src/appcontainer.ts
millerds: The reason I asked is because I have seen differing points of view on this in recent months.  Martyn seems to prefer a stricter approach where there aren't many cases where things are "expected" (the context in that discussion was auth and things like needing to refresh from a token was not considered expected).

In this case I see that the defaultSettings were undefined and so the source Location can be retrieved, but that doesn't explain why it's a problem the user created.

akrantz: I see. I think it would be better to have a separate check to throw an Error if manifest.defaultSettings is not defined.

akrantz: Actually, I take that back. Both of these are just based on the manifest being correct. If the manifest is missing a DefaultSettings section or a SourceLocation field, that would cause the error, which is ExpectedError because it is based on user input.

packages/office-addin-dev-settings/src/sideload.ts
millerds: Igor, if you remember our working session there were filed called "customDimensions" and "details" that contained data like the ExpectedError boolean.  Instead of putting these values in the error message (unless they are also helpful to the user) they should be in data fields like that.

This error gets logged when getTemplatePath doesn't return a path.  That method uses a combination of OfficeApp and AddInType to return the path and there is not default response.  It also looks like the matrix of cases small (includes just three hosts and two types).  Logging both of these values would help us understand what cases are coming back as undefined triggering the error.

In the last 7 day this is by far our top error (with hundreds of hits) and it's too generic to determine what is causing it.  Determining that would lead us to understand what cases lead the customers to hit that error and consider if there is something we can do to improve their experience.  I'm not sure if the same information being put in the displayed error message would be helpful to the customers or not.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: @akrantz yes, regex are pretty hard to read, but there are unit tests for they already and I added some more inside this PR.
By the users you mean to mask by users and not just by C:?

millerds: What does the "user" have to do with any of these . . . seems like the operate on any file path?
Why specifically absolute files paths?
Why just the "first file path" . . . seems like it would affect any file path?

packages/office-addin-dev-settings/src/appcontainer.ts
millerds: Could it be a result of errors in the manifest parsing code rather then the manifest itself?  Is the manifest malformed because of something we did or the customer explicitly did?  These may be somethings we really can't determine . . . I just want us to be careful about our assumptions on what is expected vs. not expected.

akrantz: Not likely.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: The first file path only applies to the first one because the is no `g` flag. You can that the other 2 regex have it, so there are applied to all matchs.

The user is not applied right now, I just check if the path is in the format `LETTER:` like C:

Absolute file path: C:\path_to_file
Internal path: /src/index.js

packages/office-addin-usage-data/test/test.ts
igor-ribeiiro: Sure, I will create a new test group for this function

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: I should be able to add those fields to the error details, it is just on another part of the code. In this case, it is in the `startDebugging` function inside office-addin-debugging.

But we should make the error that the user see better right? If it is not the apptype, which more information should we give to the user?

igor-ribeiiro: @millerds I just added the new dimensions to the reported error message: app, apptype and document.

igor-ribeiiro: @akrantz Could you suggest a better message to be sent to the user? It needs to help him identify the cause of the issue and the current one is bad at that (I say bad because me and Darren spend a good amount of time trying to figure it out).

packages/office-addin-manifest/src/addInTypes.ts
```diff
@@ -48,7 +50,7 @@ export function parseAddInType(value: string): AddInType {
   const addInType = toAddInType(value);
 
   if (!addInType) {
-    throw new Error(`${value} is not a valid Office add-in type.`);
+    throw new ExpectedError(`${value} is not a valid Office add-in type.`);
```
millerds, line 53: Is this really expected?  This method is just converting a string to enum and has no idea on the context of where that string came from.

packages/office-addin-usage-data/src/usageData.ts
millerds: Why do we need a separate expression for the first case vs the others?

packages/office-addin-manifest/src/addInTypes.ts
akrantz: Yes, it is parsing what the user typed for the argument.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: @millerds for the first case I'm only leaving the file extension, to avoid having the UID.
For all the other on the stack I'm leaving the filename, as it is needed to debug (just the extension would not be enough).
This should fix most of the problems, as I think the UID is the most common on the first file of the stack error

packages/office-addin-manifest/src/addInTypes.ts
millerds: That may be in intended use, but the method could be used in unexpected ways.

akrantz: If this is called with a string that doesn't return an add-in type, is this a problem with our code? No. It reflects in the telemetry that an error is expected in this situation given the inputs.

millerds: It may not be a problem with the code in this method, but it could be a problem with some other code of ours that starts to call this method.  There isn't enough context in this method to indicate expected or not . . . that would have to be determined by the caller.

igor-ribeiiro: I guess if another caller would use that function in which it could raise an unexpected error, this caller could just catch this error and then parse it to be a normal error.
I feel most of the time this function should return an expected error given that the main error is passing an invalid string.

millerds: That requires the caller to know that the method is returning an expected error and that it needs to be changed to be something not expected.  Seems like a caller that wants it to be expected should be the one to catch it and make is expected.  Again . . . the method has no context to determine expected or not.

akrantz: The caller will get an exception that there is an error. The ExpectedError is so our telemetry can distinguish that the error is expected (given the input).

igor-ribeiiro: I fell like putting to be just error is the safe approach, given that we generally won't be looking that much into expected errors in the telemetry.
Because of all of this discussion, I fell like just `error` would the right way.

packages/office-addin-manifest/src/addInTypes.ts
```diff
@@ -50,7 +50,7 @@ export function parseAddInType(value: string): AddInType {
   const addInType = toAddInType(value);
 
   if (!addInType) {
-    throw new ExpectedError(`${value} is not a valid Office add-in type.`);
+    throw new Error(`${value} is not a valid Office add-in type.`);
```
akrantz: @millerds @igor-ribeiiro Let's chat offline. I really think this should be ExpectedError.

packages/office-addin-manifest/src/addInTypes.ts
akrantz: Let's discuss offline.

packages/office-addin-manifest/src/addInTypes.ts
igor-ribeiiro: I feel like this shouldn't be a big deal, that's why I changed to error, because it is the safer approach.

igor-ribeiiro: @millerds

millerds (approved):

packages/office-addin-manifest/src/addInTypes.ts
igor-ribeiiro: @millerds @akrantz I would like to make it be an expected error for now just to be with convention with the rest of the project. We could change it later in another PR

millerds (approved):

PR #430, opened 2021-04-29
Test change to so we can test the pipeline - won't be checked in

TCourtneyOwen:


TCourtneyOwen:
Hmm.. not sure what's going on here - the pipeline simply won't start. However if I manually trigger a job on the pipeline for this repo, it starts and passes

igor-ribeiiro:
> Hmm.. not sure what's going on here - the pipeline simply won't start. However if I manually trigger a job on the pipeline for this repo, it starts and passes

Even if you start the pipeline manually it doesn't show as if the tests passed here in github.
How can I trigger manually the tests for [https://github.com/OfficeDev/Office-Addin-Scripts/pull/428](https://github.com/OfficeDev/Office-Addin-Scripts/pull/428) and them force merge it? I say force merge because I think the merging will be blocked until all tests have passed.

I will also raise this problem as a problem for the team.

akrantz:
The pipelines are not running right now. There are config changes needed.

packages/office-addin-test-helpers/src/testHelpers.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 import * as fetch from "isomorphic-fetch";
-export const defaultPort: number = 4201;
+export const defaultPort: number = 4202;
```
akrantz, line 4: what is this change for?

.azure-devops/build-pipeline.yaml
```diff
@@ -25,7 +25,7 @@ resources:
 jobs:
 - job: Windows
   pool:
-    name: OPX Windows 1903 Dogfood Pool
+    name: OPX Windows 1903 Dogfood Pool - GitHub
```
akrantz, line 28: This will need to be checked in eventually. You can run the pipeline against a branch manually instead of creating a PR that you will abandon.

packages/office-addin-test-helpers/src/testHelpers.ts
igor-ribeiiro: This PR was created just to test if the pipelines were running.
This changes won't be merged, it is just there when the PR was created

TCourtneyOwen:
I see that the pipeline ran. I made a tweak to it so maybe that fixed it unless someone manually ran this job?

PR #432, opened 2021-05-06
Update agent pool in yaml file to point to new GitHub pool

igor-ribeiiro:


akrantz (approved):

TCourtneyOwen:
It doesn't look like the pipeline is actually running here

PR #449, opened 2021-05-11
Merge linter settings

millerds:
Current the project has a eslint-config-office-addins and eslint-plugin-office-addins.  The latter seems to have shareable configs as well.  On top of this the places that reference these configs seem to have a lot of common settings.  This will update the config in the plugin to match the common cases (which can be overridden in a project that is taking them) and then have the other packages here point to it.


akrantz (approved):
I cannot tell whether this is truly correct or not but the changes here seem reasonable to me.
I trust that you've gone to each of the project types and validated that the linter config and prettier is used by introducing errors and seeing that the linter flags/fixes them and prettier works as well.

igor-ribeiiro (approved):

PR #450, opened 2021-05-15
Linter prototype rule

millerds:
This add the first linter rule and test infrastructure to the plugin.  The rule isn't hooked up to the config so a local change is necessary to enable the rule:  
- add the rule to the list of rules in main.ts of the plugin
- or add the rule to the template eslintrc file

A local change to include the "bad code" is also needed to trigger the rule.
Broader hookup to the linter config will happen when we are ready to "ship" our first set of rules.

### DIFF
--- a/packages/eslint-plugin-office-addins/jest.config.js
+++ b/packages/eslint-plugin-office-addins/jest.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
+};
--- a/packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
@@ -1,0 +1,28 @@
+import { TSESTree  } from "@typescript-eslint/typescript-estree";
+
+export = {
+  name: "no-context-sync-in-loop",
+  meta: {
+    type: <"problem" | "suggestion" | "layout"> "problem",
+    messages: {
+      loopedSync: "Calling context.sync() inside a loop can lead to bad performance"
+    },
+    docs: { 
+      description: "Calling context.sync() inside of a loop dramatically increates the time the code runs the more iterations that are run",
+      category: <"Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"> "Best Practices",
+      recommended: <false | "error" | "warn"> false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/concepts/correlated-objects-pattern?view=powerpoint-js-1.1"
+    },
+    schema: []
+  },
+  create: function(context: any) {
+    return {
+      "CallExpression[callee.property.name='run'] :matches(ForStatement, ForInStatement, WhileStatement, DoWhileStatement, ForOfStatement) CallExpression[callee.object.name='context'][callee.property.name='sync']"(node: TSESTree.CallExpression) {
+        context.report({
+          node: node.callee,
+          messageId: "loopedSync"
+        })
+      }
+    }
+  }
+}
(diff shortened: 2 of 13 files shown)


packages/eslint-plugin-office-addins/jest.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json'],
+};
```
igor-ribeiiro, line 5: Did you check the other attributes we had on the previous code?
[Plugin Custom Functions](https://github.com/OfficeDev/Office-Addin-Scripts/blob/master/packages/eslint-plugin-excel-custom-functions/jest.config.js)

Just because it had some extra configs that could be useful

packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export = {
+  name: "no-context-sync-in-loop",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"problem",
+    messages: {
+      loopedSync: "Calling context.sync() inside a loop can lead to bad performance",
+    },
+    docs: {
+      description: "Calling context.sync() inside of a loop dramatically increates the time the code runs the more iterations that are run",
```
igor-ribeiiro, line 11: increates -> increases

packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
```diff
@@ -0,0 +1,30 @@
+import { TSESTree } from "@typescript-eslint/typescript-estree";
+
+export = {
+  name: "no-context-sync-in-loop",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"problem",
+    messages: {
+      loopedSync: "Calling context.sync() inside a loop can lead to bad performance",
+    },
+    docs: {
```
igor-ribeiiro, line 10: This docs section is just for metadata? Or the client does see some of that?

packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
igor-ribeiiro, line 11: I think the message can be changed to something more like:
"Each call of context.sync() in a loop dramatically increases the execution time of the code."

igor-ribeiiro (changes requested):

packages/eslint-plugin-office-addins/jest.config.js
millerds: I looked at those and decided to go with the simpler basic config here (notice the other one doesn't us the preset) and add the others as we discover a need for them.

packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
millerds: Thanks! 

. . . we really need a PM or documentation expert to chime in on the exact text.

igor-ribeiiro: > Thanks!
> 
> . . . we really need a PM or documentation expert to chime in on the exact text.

Yeah, let's raise this concern on the daily, it would be great to have the documentation written the scearios

packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
millerds: I believe it's original design for for eslint documentation generation, but at least some of this (the url specifically) is getting used by VS Code.  Still have to figure out the rest.

packages/eslint-plugin-office-addins/src/rules/no-context-sync-in-loop.ts
millerds: Need a design/PM/Doc writer to give us specific text.

The performance impact of each individual call isn't really noticeable . . . and this rule and its documentation is specifically about loops.  The other rule in our list (calling context.sync()) would be we were we talk about the number of times it's called.

igor-ribeiiro (approved):

PR #454, opened 2021-05-19
Office ready rule

igor-ribeiiro:
Creating rule for:
"
Calling Office.initialize instead of Office.onReady
"

The idea is for this just to be a suggestion on you to use the onReady instead of initialize.
It doesn't check for the app being Outlook for now, because I couldn't make it work (the async problem).


packages/eslint-plugin-office-addins/src/rules/office-ready.ts
```diff
@@ -0,0 +1,30 @@
+import { TSESTree } from "@typescript-eslint/typescript-estree";
+
+export = {
+  name: "office-ready",
+  meta: {
```
millerds, line 5: Take a look at the preferred rule naming conventions from eslint.  Should probably be something like "use-office-on-ready" or "no-office-initialize"

packages/eslint-plugin-office-addins/src/rules/office-ready.ts
```diff
@@ -0,0 +1,30 @@
...
+    type: <"problem" | "suggestion" | "layout">"suggestion",
+    messages: {
+      officeOnReady: "Prefer calling Office.onReady() instead of Office.initialize",
+    },
+    docs: {
+      description: "It is a bad practice to call Office.initialize instead of Office.onReady",
+      category: <"Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors">"Best Practices",
```
millerds, line 12: I would describe it more like "Office.onReady() is more flexible the Office.initalize" rather then indicating it's a "bad practice".

This will likely be updated later when we get better messages.

packages/eslint-plugin-office-addins/test/rules/office-ready.test.ts
```diff
@@ -0,0 +1,81 @@
...
+const ruleTester = new ESLintUtils.RuleTester({
+  parser: '@typescript-eslint/parser',
+});
+
+ruleTester.run('office-ready', rule, {
+  valid: [ 
+    {
+      code: `
+        Office.onReady();`
+    },
```
millerds, line 13: Nit: don't need the extra newline

packages/eslint-plugin-office-addins/test/rules/office-ready.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      code: `
+        Office.onReady(function(info) {
+          if (info.host === Office.HostType.Excel) {
+              // Do Excel-specific initialization (for example, make add-in task pane's
+              // appearance compatible with Excel "green").
+          }
+          if (info.platform === Office.PlatformType.PC) {
+              // Make minor layout changes in the task pane.
+          }
+          console.log("Office.js is now ready");
```
millerds, line 24: probably don't need these extra if statements.

packages/eslint-plugin-office-addins/src/rules/office-ready.ts
igor-ribeiiro: I followed this convention: [Rules Convention](https://eslint.org/docs/developer-guide/working-with-rules#rule-naming-conventions). It doesn't explicit say what to do in case of a "positive rule".
In this case, I do think the "no-office-initialize" could be proper and I will be using that now

packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
```diff
@@ -0,0 +1,30 @@
...
+export = {
+  name: "no-office-initialize",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"suggestion",
+    messages: {
+      noOfficeInitialize: "Prefer calling Office.onReady() instead of Office.initialize",
+    },
+    docs: {
+      description: "Office.onReady() is more flexible the Office.initalize",
+      category: <"Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors">"Best Practices",
```
millerds, line 12: use "than" instead of "the"

igor-ribeiiro: Messages are hard! hhahaha

packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
```diff
@@ -0,0 +1,30 @@
+import { TSESTree } from "@typescript-eslint/typescript-estree";
+
+export = {
+  name: "no-office-initialize",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"suggestion",
+    messages: {
+      noOfficeInitialize: "Prefer calling Office.onReady() instead of Office.initialize",
+    },
```
millerds, line 9: How about "Office.onReady() is preferred over Office.initialize"

packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
millerds: Yes they are :-)

millerds (approved):

PR #456, opened 2021-05-23
Update README.md

ralphtheninja:
Add missing await.

I know this is on top level, but passing options to `https.createServer()` is incorrect, since it's a `Promise`.


igor-ribeiiro (approved):

igor-ribeiiro:
> Add missing await.
> 
> I know this is on top level, but passing options to `https.createServer()` is incorrect, since it's a `Promise`.

Thanks for the change, we appreciate it

PR #457, opened 2021-05-24
Adding test config for office-addin-lint

millerds:
Adding another config in eslint-plugin-office-addings and a command argument in office-addin-lint to use it.  This will also us to add rules to the test config and then once they are in then the test project only have to adjust their call to use the test lint config instead of symlinking between the different parts locally (plugin, lint, and template).

### DIFF
--- a/packages/office-addin-lint/src/lint.ts
+++ b/packages/office-addin-lint/src/lint.ts
@@ -21,7 +21,7 @@ function normalizeFilePath(filePath: string): string {
   return filePath.replace(/ /g, "\\ "); // Converting space to '\\'
 }
 
-function getEsLintBaseCommand(useTestConfig: boolean): string {
+function getEsLintBaseCommand(useTestConfig: boolean = false): string {
   const configFilePath = useTestConfig ? eslintTestConfigPath : eslintConfigPath
   const eslintBaseCommand: string = `node ${eslintFilePath} -c ${configFilePath} --resolve-plugins-relative-to ${__dirname}`;
   return eslintBaseCommand;
@@ -32,7 +32,7 @@ export function getLintCheckCommand(files: string, useTestConfig: boolean = fals
   return eslintCommand;
 }
 
-export function performLintCheck(files: string, useTestConfig: boolean) {
+export function performLintCheck(files: string, useTestConfig: boolean = false) {
   try {
     const command = getLintCheckCommand(files, useTestConfig);
     execCommand(command);
@@ -52,7 +52,7 @@ export function getLintFixCommand(files: string, useTestConfig: boolean = false)
   return eslintCommand;
 }
 
-export function performLintFix(files: string, useTestConfig: boolean) {
+export function performLintFix(files: string, useTestConfig: boolean = false) {
   try {
     const command = getLintFixCommand(files, useTestConfig);
     execCommand(command);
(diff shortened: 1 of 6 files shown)


packages/office-addin-lint/src/lint.ts
```diff
@@ -20,19 +21,20 @@ function normalizeFilePath(filePath: string): string {
   return filePath.replace(/ /g, "\\ "); // Converting space to '\\'
 }
 
-function getEsLintBaseCommand(): string {
-  const eslintBaseCommand: string = `node ${esLintFilePath} -c ${esLintConfigPath} --resolve-plugins-relative-to ${__dirname}`;
+function getEsLintBaseCommand(useTestConfig: boolean): string {
```
igor-ribeiiro, line 24: Can we put a default value for `useTestConfig` here to false?
It would be great to not need to always specify that it is not a test.

packages/office-addin-lint/src/lint.ts
```diff
@@ -20,19 +21,20 @@ function normalizeFilePath(filePath: string): string {
...
 
-export function getLintCheckCommand(files: string): string {
-  const eslintCommand: string = `${getEsLintBaseCommand()} ${normalizeFilePath(files)}`;
+export function getLintCheckCommand(files: string, useTestConfig: boolean = false): string {
+  const eslintCommand: string = `${getEsLintBaseCommand(useTestConfig)} ${normalizeFilePath(files)}`;
   return eslintCommand;
 }
 
-export function performLintCheck(files: string) {
+export function performLintCheck(files: string, useTestConfig: boolean) {
```
igor-ribeiiro, line 35: And also here (put the default to false on `useTestConfig`)

packages/office-addin-lint/src/lint.ts
```diff
@@ -45,14 +47,14 @@ export function performLintCheck(files: string) {
...
 
-export function getLintFixCommand(files: string): string {
-  const eslintCommand: string = `${getEsLintBaseCommand()} --fix ${normalizeFilePath(files)}`;
+export function getLintFixCommand(files: string, useTestConfig: boolean = false): string {
+  const eslintCommand: string = `${getEsLintBaseCommand(useTestConfig)} --fix ${normalizeFilePath(files)}`;
   return eslintCommand;
 }
 
-export function performLintFix(files: string) {
+export function performLintFix(files: string, useTestConfig: boolean) {
```
igor-ribeiiro, line 55: Here too

packages/office-addin-lint/test/src/test.ts
```diff
@@ -11,12 +11,16 @@ describe("test cases", function() {
     it("eslint and prettier commands", async function() {
         const inputFile = "./test/cases/basic/functions.ts";
         const lintExpectedCommand = "./test/cases/basic/functions.ts";
+        const lintTestExpectedCommand = ".eslintrc.test.json"
```
igor-ribeiiro, line 14: I found this name quite misleading. It is not really a Command, more like a file inside the command.
Would the change `lintTestExpectedCommand` to `lintTestExpectedCommandFile`.

igor-ribeiiro (changes requested):

igor-ribeiiro (approved):

PR #465, opened 2021-06-01
Eslint call load on will be read objects

igor-ribeiiro:
This is the rule implemented: "Not loading objects returned by get*() before reading them"


packages/eslint-plugin-office-addins/src/main.ts
```diff
@@ -48,6 +48,7 @@ module.exports = {
         "office-addins/no-context-sync-in-loop": "warn",
         "office-addins/no-office-initialize": "warn",
         "office-addins/test-for-null-using-isNullObject": "error",
+        "office-addins/load-object-before-read": "error",
```
millerds, line 51: lets keep these in alphabetical order

igor-ribeiiro:
@millerds just a gentle reminder for us to continue this PR

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+    function getValueThatHadToBeLoaded(referenceNode: Reference): string | undefined {
+      if(referenceNode.identifier.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression
+        && referenceNode.identifier.parent?.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+        return referenceNode.identifier.parent.property.name;
+      }
+      return undefined;
+    }
+
+    function isLoadFunction(reference: Reference): boolean {
+      if(reference.identifier.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression
```
millerds, line 36: Should pass in the actual node you are testing instead of an indirection to it (makes it more generic)

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+        && reference.identifier.parent.property.type === TSESTree.AST_NODE_TYPES.Identifier
+        && reference.identifier.parent.property.name === "load") {
+        return true;
+      }
+      return false;
+    }
+
+    function wasCreatedByGetFunction(referenceNode: Reference): boolean {
+      const variable = referenceNode.resolved;
```
millerds, line 45: Should pass in the actual node you are testing instead of an indirection to it (makes it more generic).

I also think the name should be a little clearer ("what are you testing was created by a get function).

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+  create: function (context: any) {
+    function getValueThatHadToBeLoaded(referenceNode: Reference): string | undefined {
+      if(referenceNode.identifier.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression
+        && referenceNode.identifier.parent?.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+        return referenceNode.identifier.parent.property.name;
+      }
+      return undefined;
+    }
+
+    function isLoadFunction(reference: Reference): boolean {
```
millerds, line 35: Should pass in the actual node you are testing instead of an indirection to it (makes it more generic)

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+      >"Possible Errors",
+      recommended: <false | "error" | "warn">false,
+      url: 
+        "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#load",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    function getValueThatHadToBeLoaded(referenceNode: Reference): string | undefined {
+      if(referenceNode.identifier.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression
```
millerds, line 28: Value?  Seems like "property" would be better.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+      return getFunctionFound;
+    }
+
+    function isLoaded(referenceNode: Reference): boolean {
+      const variable = referenceNode.resolved;
+      let loadFound = false;
+      const valueRead = getValueThatHadToBeLoaded(referenceNode);
+
+      variable?.references.forEach((reference: Reference) => {
+        if(reference === referenceNode) return;
```
millerds, line 69: I'm not following this very well, but it seems like you are doing repeating loops through references to find specific instances.  Is it possible to simply the pass across the variables and references so there isn't so many loops (more performant)?

millerds, line 69: Aren't you supposed to return a boolean?

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+      let getFunctionFound = false;
+      variable?.references.forEach((reference: Reference) => {
+          if(reference.identifier.parent?.type === TSESTree.AST_NODE_TYPES.VariableDeclarator
+            && reference.identifier.parent.init?.type === TSESTree.AST_NODE_TYPES.CallExpression
+            && reference.identifier.parent.init.callee.type === TSESTree.AST_NODE_TYPES.MemberExpression
+            && reference.identifier.parent.init.callee.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+              const functionName = reference.identifier.parent.init.callee.property.name;
+              if(functionName === "getSelectedRange"
+                || functionName === "getItem" 
+                || functionName === "getRange") {
```
millerds, line 55: Seems like this big if statement could be extracted to a "isGetFunction" to make the code a bit clearer.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+    function wasCreatedByGetFunction(referenceNode: Reference): boolean {
+      const variable = referenceNode.resolved;
+      let getFunctionFound = false;
+      variable?.references.forEach((reference: Reference) => {
+          if(reference.identifier.parent?.type === TSESTree.AST_NODE_TYPES.VariableDeclarator
+            && reference.identifier.parent.init?.type === TSESTree.AST_NODE_TYPES.CallExpression
+            && reference.identifier.parent.init.callee.type === TSESTree.AST_NODE_TYPES.MemberExpression
+            && reference.identifier.parent.init.callee.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+              const functionName = reference.identifier.parent.init.callee.property.name;
+              if(functionName === "getSelectedRange"
```
millerds, line 53: Are you sure these are the only functions?  Given the rule I made seems like there should at least be some "*OrNullObject" variations.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+          && reference.identifier.parent.parent.arguments[0].value === valueRead
+          && reference.identifier.range[1] < referenceNode.identifier.range[1]) {
+          loadFound = true;
+        }
+      });
+
+      return loadFound;
+    }
+
+    function findVariablesInScope(scope: any) {
```
millerds, line 81: This sounds like a generic find function.  Should us a name more specific to this purpose.

packages/eslint-plugin-office-addins/test/rules/load-object-before-read.test.ts
```diff
@@ -0,0 +1,80 @@
...
+        var variableName = property.G2;`
+    },
+    {
+      code: `
+        var selectedRange = context.workbook.getSelectedRange();
+        selectedRange.load('values');
+        if(selectedRange.values === [2]){}`
+    }
+  ],
+  invalid: [
```
millerds, line 34: Should have a test case involving an assignment (and reassignment) instead of just declaration

-  var myRange;
   myRange = context.workbook.worksheets.getSelectedRange();
- var myRange = context.workbook.worksheets.getSelectedRange();
   myRange = context.workbook.worksheets.getItem(sheetName).getRange(rangeAddress);

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+        return true;
+      }
+      return false;
+    }
+
+    function wasCreatedByGetFunction(referenceNode: Reference): boolean {
+      const variable = referenceNode.resolved;
+      let getFunctionFound = false;
+      variable?.references.forEach((reference: Reference) => {
+          if(reference.identifier.parent?.type === TSESTree.AST_NODE_TYPES.VariableDeclarator
```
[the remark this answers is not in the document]
igor-ribeiiro: The only example that I am thinking of declaring a variable that is not currently support is:
a = b = 2;
I think this will not work for the b variable

Do you have any other examples or it is just this? I will try to tackle this problem.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: No, I'm not sure, I reached to riccki to see if he has a complete list of the get functions

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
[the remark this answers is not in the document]
millerds: var a = get(); // VariableDeclarator,

var a;
a = get();  // not a VariableDeclarator, but rather an AssignmentExpression.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: This is not really a return as normal. This is acting as a `continue` statement. 
I think this is a confusing thing in Javascript, but it is explained here:
[Return in ForEach](https://stackoverflow.com/questions/22585032/does-the-foreach-loop-allow-using-break-and-continue#:~:text=As%20already%20answered%2C%20you%20cannot%20use%20continue%20or,continue%20or%20throw%20new%20Error%20%28%29%20to%20break.)

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,107 @@
...
+        });
+      return getFunctionFound;
+    }
+
+    function isLoaded(referenceNode: Reference): boolean {
+      const variable = referenceNode.resolved;
+      let loadFound = false;
+      const valueRead = getValueThatHadToBeLoaded(referenceNode);
+
+      variable?.references.forEach((reference: Reference) => {
```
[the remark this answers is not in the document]
igor-ribeiiro: I was doing it in N^2 when looking back.
I fixed it so now it runs in linear time

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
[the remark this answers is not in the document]
igor-ribeiiro: Thanks, I added these cases to be tested and fixed them

igor-ribeiiro:
@millerds finished working on the requested comments. Now, this PR is ready to review again.
Currently the corner case of:
`var a = b = context.getRange();`
Is not getting covered. This case is when the assignment happens inside another assignment. This can happen recursively as well and the code doesn't support it for now. If you think I should add I can work on that.

The formatting probably has a lot of mistakes. I hope after we merge the changes from https://github.com/OfficeDev/Office-Addin-Scripts/pull/471 we can just run lint and prettier to fix them without having to worry about it now.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,146 @@
...
+    function findLoadBeforeRead(scope: Scope) {
+      scope.variables.forEach((variable: Variable) => {
+        let loadLocation: Map <string, number> = new Map<string, number>();
+        let getFound: boolean = false;
+        variable.references.forEach((reference: Reference) => {
+          const node: TSESTree.Node = reference.identifier;
+            if (isVariableDeclaration(node)
+              || isAssignmentExpression(node)) {
+              if (isGetVariableDeclaration(node)
+                || isGetAssignmentExpression(node)) {
```
millerds, line 102: Looks like you are duplicating work here.  You check for isVariableDeclaration and then you check for isGetVariableDeclaration.  You should be able to simplify those such that there is a single check for the get method that is used in both cases and you don't have to recheck the same things.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,146 @@
...
+        "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#load",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    function getPropertyThatHadToBeLoaded(node: TSESTree.Node): string | undefined {
+      if(node.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression
+        && node.parent?.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+        return node.parent.property.name;
+      }
```
millerds, line 33: You're really doing all the testing on the parent node so you should pass that in instaed.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: Yeah,  did notice that. It is just one small overlap on `node. parent?.type` check.
The problem that I was having and the reason I chose to make the code like this is that Typescript keeps complaining about accessing a node type without checking for the type first. So I needed to make the condition on both of the functions.

The choice was: or leave it like that with function calls and this small code duplication (one check).
Or not use functions and put all the if checks here

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: Yes, I'm doing all the computations on the parent, which I think is also done on some other functions.
I could pass the parent beforehand, but then I will have to update the main code to see that and also check if the parent is not null when calling the function.

Do you thing it makes sense doing those steps everytime the functions is called? 

In general, I'm making the function receive a node reference everytime and this is pretty consistent accross all the functions in this file and I think this make it easier when calling the functions.

millerds: You don't have to check the parent is null in the calling code . . . you can do it as part of the functions still.  See isNullObjectNode in the test-for-null-using-isNullObject rule.  

You are receiving a node either way . . . it's just a matter of which node you are looking at.  You can also type the input parameter to something more specific the just "node" and then you don't have to include that in the if statement.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
millerds: In the case of isNullObjectNode in test-for-null-using-isnullObject rule I ended up putting them in a complex if clause and had one shared method (taking advantage of a shared node type to examine that is more generic) for looking at the expression to determine if it was what I was looking for.  I could simplify a bit more by putting each half of the main if clause in a function, but I still wouldn't be testing the type twice.

Even if you didn't want to change much here you could simply remove the outer if clause and it would be the same.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,297 @@
...
+            return true;
+          }
+      }
+      return false;
+    }
+
+    function getLoadedPropertyName(node: TSESTree.Node): string {
+      if(node.parent?.parent?.type === TSESTree.AST_NODE_TYPES.CallExpression
+        && node.parent.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.Literal) {
+          return node.parent.parent.arguments[0].value as string;
```
millerds, line 239: You should simplify instead of having to look at parents of parents ect as that narrows the use of this method to a very specific type of heirachy.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,297 @@
...
+              || isGetAssignmentExpression(node)) {
+                getFound = true;
+                return;
+            } else {
+              getFound = false;
+            }
+          }
+          
+          if(!getFound) {
+            return;
```
millerds, line 262: If all the getFound is doing is determining return or not, you could just return from inside the if clause.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,297 @@
...
+      return "error in getLoadedPropertyName";
+    }
+
+    function findLoadBeforeRead(scope: Scope) {
+      scope.variables.forEach((variable: Variable) => {
+        let loadLocation: Map <string, number> = new Map<string, number>();
+        let getFound: boolean = false;
+        variable.references.forEach((reference: Reference) => {
+          const node: TSESTree.Node = reference.identifier;
+          if (isVariableDeclaration(node)
```
millerds, line 250: Could probably also check the "getFound" here and not do the other checks if it was already seen.

packages/eslint-plugin-office-addins/test/rules/load-object-before-read.test.ts
```diff
@@ -0,0 +1,136 @@
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
+import rule from '../../src/rules/load-object-before-read';
+
+const ruleTester = new ESLintUtils.RuleTester({
+  parser: '@typescript-eslint/parser',
+});
+
+ruleTester.run('load-object-before-read', rule, {
+  valid: [
```
millerds, line 9: Is it possible and valid to have:

  var myRange = context.workbook.worksheets.getItem(sheetName).getRange(rangeAddress);          
  myRange.load('address');
  myRange.address.someOtheProperty

or does the sync and load have to happen before the nested property can be accessed?

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: I didn't quite get it, what do you mean by "ect"? Was it a typo?

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: It is determining if the return will happen or not across different references, so it cannot go inside the if.
The get found will have the same value across different iterations.

packages/eslint-plugin-office-addins/test/rules/load-object-before-read.test.ts
igor-ribeiiro: The sync has to be called beforehand. 
This case should be handled by another rule: "Always call load or an object (followed by a sync) before reading it or one of its properties."

They almost detect the same thing, but the messages are different. I think the correct approach would be to merge both rules and show different messages?

millerds: I was focused on the nested property case.  I think the sync would be a separate rule.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
millerds: So maybe I'm a little confused about the reason for "getFound".  It is defined outside the forEach, but it is getting set each time you go in.  You are are returning right after you set it to true, but then returning if "!getFound" (which should be every case that isn't a get statement).  Reading it this time around makes be wonder how the rule is finding an error at all.  Seems like there is something I'm missing.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
millerds: That should have been "etc" . . . short hand for etcetera.  Basically I disagree with directly traveling up the tree so much in this way.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: Sure, I managed to change the code to not have duplicate checks (and also tackled the parent issue within the same code). Now, there is no duplicate checks on this

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,292 @@
...
+        && node.init.callee.type === TSESTree.AST_NODE_TYPES.MemberExpression
+        && node.init.callee.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+          if(callsGetAPIFunction(node.init.callee.property)) {
+            return true;
+          }
+      }
+      return false;
+    }
+
+    function isGetAssignmentExpression(node: TSESTree.AssignmentExpression): boolean {
```
millerds, line 209: This looks almost identical to isGetVariableDeclaration.  If you use the CallExpression node as the method argument then these could be collapsed into one.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,292 @@
...
+      "getLevelParagraphs",
+      "getBase64ImageSrc",
+      "getSelection",
+      "getByTypes",
+      "getByTag",
+      "getReferenceId", 
+      "getDocument",
+    ]);
+
+    function getPropertyThatHadToBeLoaded(node: TSESTree.Node | undefined): string | undefined {
```
millerds, line 180: Currently this method doesn't appear to have anything to do with a property from a get statement like it's name suggests.  I know that is what you are using it for, but there is no tie or check for the get method so the name doesn't really fit.

igor-ribeiiro: Ow, I guess I chose a confusing name
The `get` here really has nothing to do with the other gets.
The `get` here was more with the meaning: getSomeValue. I will change the function name because it got confusing.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: I merged both functions but I didn't quite figure how to merge the two ifs pairs easily.
I also made the statements inside the if much easier to read

packages/eslint-plugin-office-addins/test/rules/load-object-before-read.test.ts
igor-ribeiiro: @millerds after doing some research, I found that there are no navigation properties (which would lead to nested properties) coming from the get* functions.
So we don't need to do these checks, because there will never be nested properties.

igor-ribeiiro: Asked for clarification on this rule and there is a slash that needs to be used for those. I added test cases for that and the added the necessary code.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,293 @@
...
+        && node.callee.type === TSESTree.AST_NODE_TYPES.MemberExpression
+        && node.callee.property.type === TSESTree.AST_NODE_TYPES.Identifier) {
+          if(callsGetAPIFunction(node.callee.property)) {
+            return true;
+          }
+      }
+      return false;
+    }
+
+    function getLoadedPropertyName(node: TSESTree.MemberExpression): string {
```
millerds, line 212: Probably don't need to use "loaded" in the method name here as the body doesn't have any indication of loading anything.  It just retrieves the expression's first argument's name.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -0,0 +1,293 @@
...
+      "getLevelParagraphs",
+      "getBase64ImageSrc",
+      "getSelection",
+      "getByTypes",
+      "getByTag",
+      "getReferenceId", 
+      "getDocument",
+    ]);
+
+    function findPropertyThatHadToBeLoaded(node: TSESTree.Node | undefined): string {
```
millerds, line 180: Name still doesn't match body.  "HadToBeLoaded" is not evident here.

millerds (approved):

PR #466, opened 2021-06-01
New is null object rule

millerds:
Note that the current implementation does not distinguish weather or not a variable is set to the "*OrNullObject" result before it is used in a test . . . only that if a variable is set to the "*OrNullObject" at some point then there test of that object are using the isNullObject property.

It also contains a fix method.

### DIFF
--- a/packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
@@ -5,13 +5,17 @@ export = {
   meta: {
     type: <"problem" | "suggestion" | "layout">"suggestion",
     messages: {
-      noOfficeInitialize: "Office.onReady() is preferred over Office.initialize",
+      noOfficeInitialize:
+        "Office.onReady() is preferred over Office.initialize",
     },
     docs: {
       description: "Office.onReady() is more flexible than Office.initalize",
-      category: <"Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors">"Best Practices",
+      category: <
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Best Practices",
       recommended: <false | "error" | "warn">false,
-      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/initialize-add-in#initialize-with-officeonready",
+      url:
+        "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/initialize-add-in#initialize-with-officeonready",
     },
     schema: [],
   },
--- a/packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
@@ -1,0 +1,155 @@
+import { TSESTree, AST_NODE_TYPES } from "@typescript-eslint/typescript-estree";
+import {
+  Reference,
+  Variable,
+} from "@typescript-eslint/experimental-utils/dist/ts-eslint-scope";
+import { RuleFix, RuleFixer } from "@typescript-eslint/experimental-utils/dist/ts-eslint";
+
+export = {
+  name: "test-for-null-using-isNullObject",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"problem",
+    messages: {
+      useIsNullObject: "Test {{name}} for null using isNullObject.",
+    },
+    docs: {
+      description: "",
+      category: <
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Possible Errors",
+      recommended: <false | "error" | "warn">false,
+      url:
+        "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#ornullobject-methods-and-properties",
+    },
+    schema: [],
+    fixable: <"code" | "whitespace"> "code",
+  },
+  create: function (context: any) {
+    function isConditionalTestExpression(node: TSESTree.Identifier): boolean {
+      return (
+        node.parent != undefined &&
+        (node.parent.type === AST_NODE_TYPES.IfStatement ||
+          node.parent.type === AST_NODE_TYPES.WhileStatement ||
+          node.parent.type === AST_NODE_TYPES.DoWhileStatement ||
+          node.parent.type === AST_NODE_TYPES.ForStatement ||
+          node.parent.type === AST_NODE_TYPES.ConditionalExpression) &&
+        node === node.parent.test
+      );
+    }
+
+    function isInUnaryNullTest(node: TSESTree.Identifier): boolean {
+      return (
+        node.parent != undefined &&
+        node.parent.type === AST_NODE_TYPES.UnaryExpression &&
+        node.parent.operator === "!" &&
+        node.parent.argument === node
+      );
+    }
+
+    function isInBinaryNullTest(node: TSESTree.Identifier): boolean {
+      return (
+        node.parent != undefined &&
+        node.parent.type === AST_NODE_TYPES.BinaryExpression &&
+        ((node.parent.left === node &&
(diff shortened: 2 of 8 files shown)


packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
```diff
@@ -5,13 +5,17 @@ export = {
...
+      noOfficeInitialize:
+        "Office.onReady() is preferred over Office.initialize",
     },
     docs: {
       description: "Office.onReady() is more flexible than Office.initalize",
-      category: <"Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors">"Best Practices",
+      category: <
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Best Practices",
       recommended: <false | "error" | "warn">false,
```
igor-ribeiiro, line 16: Shouldn't this follow the same formatting as the cattegory?

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -0,0 +1,155 @@
...
+      description: "",
+      category: <
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Possible Errors",
+      recommended: <false | "error" | "warn">false,
+      url:
+        "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#ornullobject-methods-and-properties",
+    },
+    schema: [],
+    fixable: <"code" | "whitespace"> "code",
```
igor-ribeiiro, line 25: This too could follow the same category formatting pattern?

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -0,0 +1,155 @@
+import { TSESTree, AST_NODE_TYPES } from "@typescript-eslint/typescript-estree";
```
igor-ribeiiro, line 1: You said you would look at the [no-use-before-define](https://github.com/eslint/eslint/blob/85a2725b1fade5538e727102d9701ccb503e54d4/lib/rules/no-use-before-define.js) rule before submitting. Did you have a look in this yet?

packages/eslint-plugin-office-addins/test/rules/test-for-null-using-isNullObject.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (dataSheet.isNullObject) {
+          dataSheet = context.workbook.worksheets.add("Data");
+        }
+        dataSheet.position = 1;
+      });`,
+    },
+    {
+      code: `var dataSheet;
+      dataSheet = context.workbook.worksheets.getItemOrNullObject("Data");
+      return context.sync().then(function () {
```
igor-ribeiiro, line 36: Are you checking to context.sync in your code? I don't think it is necessary to your rule and it would be nice to have some tests without it.

I say that because the context.sync might be hidden in some function and I would just like to see some tests without it.

packages/eslint-plugin-office-addins/test/rules/test-for-null-using-isNullObject.test.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+        dataSheet.position = 1;
+      });`,
+      errors
+    },
+    {
+      code: `var dataSheet;
+      dataSheet = context.workbook.worksheets.getItemOrNullObject("Data");
+      return context.sync().then(function () {
+        if (true &&dataSheet) {
```
igor-ribeiiro, line 83: Space after &&

igor-ribeiiro (changes requested):

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -0,0 +1,155 @@
...
+      }
+      return false;
+    }
+
+    function findNullObjectNullTests(scope: any): void {
+      const variables = scope.variables;
+      const childScopes = scope.childScopes;
+      let i, l;
+
+      for (i = 0, l = variables.length; i < l; i++) {
```
igor-ribeiiro, line 104: Can you change the `l` variable? It doesn't even seem needed (It only has the lenght value) and makes the code more confusing

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -0,0 +1,155 @@
...
+        const variable: Variable = variables[i];
+        const references: Reference[] = variable.references;
+        let nullObjectCall: boolean = false;
+        let nullTests: Reference[] = [];
+
+        for (var ref = 0; ref < references.length; ref++) {
```
igor-ribeiiro, line 110: Use `let` instead of `var`

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        if (nullObjectCall === true && nullTests.length > 0) {
+          nullTests.forEach((reference) => {
+            context.report({
+              node: reference.identifier,
+              messageId: "useIsNullObject",
+              data: { name: reference.identifier.name },
+              fix: function(fixer: RuleFixer) {
+                var ruleFix: RuleFix;
```
igor-ribeiiro, line 129: Use `let` instead of `var`

igor-ribeiiro (changes requested):

packages/eslint-plugin-office-addins/src/rules/no-office-initialize.ts
millerds: Not sure what you mean . . . this is the category.  Running prettier is what created this formatting.

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
millerds: Handled by prettier.

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
millerds: I wasn't intending to rewrite this with that different pattern style.  Do you think I should?

packages/eslint-plugin-office-addins/test/rules/test-for-null-using-isNullObject.test.ts
millerds: The context sync is the technically correct way to use make the call (though there could be other forms I suppose), but will add a mix.

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
millerds: Copied from the style used by eslint.  Don't know if there is a technical cost to reading the member variable repeatedly, but I can certainly change it.

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
igor-ribeiiro: Nahh, I think it is fine :D

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
igor-ribeiiro: I think it shouldn't have any difference in performance, as typescript (or js) should optimize this call underneath.

igor-ribeiiro (approved):

igor-ribeiiro (dismissed):

igor-ribeiiro (approved):

PR #471, opened 2021-06-07
Lint packages

millerds:
This adds linting to each of the packages and then corrects things caught by the linter.  It also removes the tslint (deprecated) pieces that were there in favor of eslint with typescript-eslint


packages/custom-functions-metadata-plugin/package.json
```diff
@@ -18,10 +18,13 @@
...
   "scripts": {
-    "tsc": "tsc -p tsconfig.json",
-    "tsc-watch": "tsc -p tsconfig.json -w",
     "build": "rimraf lib && concurrently \"npm run tsc\"",
-    "test": "echo \"No tests.\""
+    "lint": "office-addin-lint check",
+    "lint:fix": "office-addin-lint fix",
+    "prettier": "office-addin-lint prettier",
+    "test": "echo \"No tests.\"",
+    "tsc": "tsc -p tsconfig.json",
```
akrantz, line 26: I don't think you should add "tsc" since there is already "build". You should rename "tsc-watch" to just "watch".

packages/custom-functions-metadata/.eslintrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "plugins": [
+    "office-addins"
+  ],
+  "extends": [
+    "plugin:office-addins/recommended"
+  ]
+}
```
akrantz, line 8: minor: add a newline since usually you want the last line to be empty. The diff shows a newline is missing at the end of this line.

packages/eslint-config-office-addins/package.json
```diff
@@ -4,9 +4,12 @@
...
   "main": "./lib/main.js",
   "scripts": {
-    "tsc": "tsc -p tsconfig.json",
     "build": "rimraf lib && concurrently \"npm run tsc\"",
-    "test": "echo \"No tests\""
+    "lint": "office-addin-lint check",
+    "lint:fix": "office-addin-lint fix",
+    "prettier": "office-addin-lint prettier",
+    "test": "echo \"No tests\"",
+    "tsc": "tsc -p tsconfig.json"
```
akrantz, line 12: Same comment about "tsc". Please go through all the packages and don't add tsc to them.

packages/office-addin-cli/package.json
```diff
@@ -4,12 +4,14 @@
...
-    "tsc-watch": "tsc -p tsconfig.json -w",
     "build": "rimraf lib && concurrently \"npm run tsc\"",
+    "cli": "node lib/cli.js",
+    "lint": "eslint src/**/*.{ts,tsx,js,jsx}",
+    "lint:fix": "eslint src/**/*.{ts,tsx,js,jsx} --fix",
     "start": "rimraf lib && concurrently \"npm run tsc-watch\"",
-    "test": "mocha -r ts-node/register test/**/*.ts"
+    "test": "mocha -r ts-node/register test/**/*.ts",
+    "tsc": "tsc -p tsconfig.json",
+    "tsc-watch": "tsc -p tsconfig.json -w"
```
akrantz, line 14: same comment here about tsc and tsc-watch...

packages/office-addin-debugging/package.json
```diff
@@ -4,13 +4,16 @@
...
     "build": "rimraf lib && concurrently \"npm run tsc\"",
+    "cli": "node lib/cli.js",
+    "lint": "office-addin-lint check",
+    "lint:fix": "office-addin-lint fix",
+    "node-debugger": "office-addin-node-debugger",
+    "prettier": "office-addin-lint prettier",
     "start": "rimraf lib && concurrently \"npm run tsc-watch\"",
     "test": "mocha -r ts-node/register test/**/*.ts",
-    "node-debugger": "office-addin-node-debugger"
+    "tsc": "tsc -p tsconfig.json",
```
akrantz, line 15: tsc / tsc-watch here too...

packages/office-addin-dev-certs/package.json
```diff
@@ -4,12 +4,15 @@
...
-    "tsc-watch": "tsc -p tsconfig.json -w",
     "build": "rimraf lib && concurrently \"npm run tsc\"",
+    "cli": "node lib/cli.js",
+    "lint": "office-addin-lint check",
+    "lint:fix": "office-addin-lint fix",
+    "prettier": "office-addin-lint prettier",
     "start": "rimraf lib && concurrently \"npm run tsc-watch\"",
-    "test": "mocha -r ts-node/register test/**/*.ts"
+    "test": "mocha -r ts-node/register test/**/*.ts",
+    "tsc": "tsc -p tsconfig.json",
```
akrantz, line 14: tsc / tsc-watch...

packages/office-addin-dev-settings/package.json
```diff
@@ -6,6 +6,9 @@
   "scripts": {
     "build": "rimraf lib && concurrently \"tsc -p tsconfig.json\"",
     "cli": "node lib/cli.js",
+    "lint": "office-addin-lint check",
+    "lint:fix": "office-addin-lint fix",
+    "prettier": "office-addin-lint prettier",
     "test": "mocha -r ts-node/register test/**/*.ts",
     "watch": "rimraf lib && concurrently \"tsc -p tsconfig.json -w\""
```
akrantz, line 13: Note that this has "build" and "watch" -- this is probably what you want across the packages...

packages/office-addin-lint/package.json
```diff
@@ -4,10 +4,13 @@
...
-    "cli": "node lib/cli.js",
-    "tsc": "tsc -p tsconfig.json",
     "build": "rimraf lib && concurrently \"npm run tsc\"",
-    "test": "mocha -r ts-node/register test/src/*.ts"
+    "cli": "node lib/cli.js",
+    "lint": "node lib/cli.js check",
+    "lint:fix": "node lib/cli.js fix",
+    "prettier": "node lib/cli.js prettier",
+    "test": "mocha -r ts-node/register test/src/*.ts",
+    "tsc": "tsc -p tsconfig.json"
```
akrantz, line 13: remove tsc, add watch?

packages/office-addin-manifest/src/manifestInfo.ts
```diff
@@ -42,30 +42,84 @@ function parseManifest(xml: Xml): ManifestInfo {
...
-    manifest.description = xmlMethods.getXmlElementAttributeValue(officeApp, "Description");
-    manifest.displayName = xmlMethods.getXmlElementAttributeValue(officeApp, "DisplayName");
-    manifest.highResolutionIconUrl = xmlMethods.getXmlElementAttributeValue(officeApp, "HighResolutionIconUrl");
-    manifest.hosts = xmlMethods.getXmlElementsAttributeValue(officeApp, "Hosts", "Host", "Name");
-    manifest.iconUrl = xmlMethods.getXmlElementAttributeValue(officeApp, "IconUrl");
-    manifest.officeAppType = xmlMethods.getXmlAttributeValue(officeApp, "xsi:type");
-    manifest.permissions = xmlMethods.getXmlElementValue(officeApp, "Permissions");
-    manifest.providerName = xmlMethods.getXmlElementValue(officeApp, "ProviderName");
-    manifest.supportUrl = xmlMethods.getXmlElementAttributeValue(officeApp, "SupportUrl");
+    manifest.allowSnapshot = xmlMethods.getXmlElementValue(
```
akrantz, line 51: Not really prettier. Is there a line length of 120 being used?

akrantz (changes requested):
looked over most of it. Assuming the code changes are only due to prettier and no actual code changes.

packages/custom-functions-metadata-plugin/package.json
millerds: I wasn't adding it . . . I just reorder the commands to be in alpha order.  That said I could go back and remove them if they are unnecessary.

packages/office-addin-manifest/src/manifestInfo.ts
millerds: I'm using office-addin-lint . . . so what ever settings existed are the ones that were used.

packages/custom-functions-metadata-plugin/package.json
millerds: I looked add if I understand correctly the build command is actually calling the tsc command.  I'm assuming this needs to be collapsed into the build command (which appears to be the case in the file you said looked good).

packages/office-addin-manifest/src/manifestInfo.ts
millerds: looks like our own prettier config package has established the 120 with setting (only thing it does).

millerds: Looks like I needed to add a reference to office-adding-prettier-config for it to take effect and us 120 as the width.  Will look into using that everywhere.

packages/custom-functions-metadata-plugin/package.json
akrantz: You shouldn't need to add `-p tsconfig.json` since typescript will use the tsconfig.json file by default. "npm run build" should already be sufficient for building.

millerds:
> looked over most of it. Assuming the code changes are only due to prettier and no actual code changes.

The vast majority ends up being prettier and global var.  I did however make some other changes based on eslint rules (like unused vars, unused imports, unused  methods, and unnecessary async in promises)

akrantz:
Can you point me to the unnecessary async in promises? I just want to double-check that. I

package.json
```diff
@@ -13,7 +13,10 @@
     "dist-tags": "npm run exec -- npm dist-tag ls",
     "install-packages": "npm run exec -- npm install & npm install",
     "exec": "lerna exec --concurrency 1 --no-bail --no-sort --stream",
+    "lint": "lerna run lint --parallel",
```
akrantz, line 16: Does --parallel give a good experience? For the "exec" script I set concurrency to 1 because otherwise the output was jumbled between different packages.

akrantz: It's fine to try it this way, it can always be easily changed.

millerds: The output is jumbled . . . but it's way faster.

millerds:
> Can you point me to the unnecessary async in promises? I just want to double-check that. I

Look in manifestInfo.ts for some examples.  You can also search the files change tab for "async function(resolve, reject)" or "async (resolve, reject)" to see cases of the change.

akrantz:
I only see one place in manifestInfo.ts in parseXmlAsync() which had `new Promise(async function` where the inner function didn't need to be async. That change is fine.

millerds:
> I only see one place in manifestInfo.ts in parseXmlAsync() which had `new Promise(async function` where the inner function didn't need to be async. That change is fine.

Looks like the page search isn't enough since some of the .ts file diffs aren't loaded.  Another example would be msgraph-helper.ts (in office-addin-sso) in getGraphData.  Other files with similar changes include server.ts  (in office-addin-sso), testHelpers.ts (in office-addin-test-helpers), and testServer.ts  (in office-addin-test-server).  In each case instead of returning a new promise with an async function in the constructor, the function code is run directly instead with Promise.resolve or Promise.reject used.

akrantz (changes requested):
I get errors when running "npm run lint" in office-addin-debugging. Also, I cannot enable automatic lint in VSCode with the eslint extension installed.

akrantz:
OK, after going through to run `npm install` in office-addin-lint and office-addin-debugging, it gets prettier dep installed correctly and works in VS Code.

akrantz (dismissed):

akrantz (approved):

PR #472, opened 2021-06-07
Update lerna to v4

millerds:
Clears up all but 2 vulnerabilities resulting from lerna dependancy

### DIFF
--- a/packages/eslint-config-office-addins/package.json
+++ b/packages/eslint-config-office-addins/package.json
@@ -16,6 +16,7 @@
   },
   "devDependencies": {
     "@types/node": "^12.12.34",
+    "@types/es6-promise": "^3.3.0",
     "concurrently": "^6.0.0",
     "rimraf": "^2.7.1",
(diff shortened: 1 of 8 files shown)


akrantz (approved):

packages/eslint-config-office-addins/package.json
```diff
@@ -16,6 +16,7 @@
   },
   "devDependencies": {
     "@types/node": "^12.12.34",
+    "@types/es6-promise": "^3.3.0",
```
akrantz, line 19: Is this related to switching lerna versions? Seems like this PR is doing more than one thing.

millerds: check your teams :-).  For some reason updating lerna versions exposed these missing pieces of a few package.json files causing the builds to break (somehow missed this when I build locally the first time).  That why the CI loop didn't pass the first time.

akrantz (approved):

PR #473, opened 2021-06-10
Use peer dependencies

millerds:
Update the eslint-plugin-office-addins package to only depend on thing it directly requires.  Adjust down level dependencies.


packages/eslint-config-office-addins/package.json
```diff
@@ -16,6 +16,10 @@
   "dependencies": {
     "eslint-plugin-office-addins": "^0.2.2"
   },
+  "peerDependencies": {
```
akrantz, line 19: why is this needed? I thought office-addin-lint as a dependency was enough.

akrantz: Or perhaps the peer dependency should be on office-addin-lint...

packages/office-addin-cli/package.json
```diff
@@ -37,6 +37,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "mocha": "^8.3.2",
     "office-addin-prettier-config": "^1.0.16",
+    "prettier": "^2.1.2",
```
akrantz, line 40: This dependency on prettier shouldn't be needed, right?  Shouldn't it be be office-addin-lint?

packages/office-addin-lint/package.json
```diff
@@ -31,6 +31,7 @@
     "eslint-config-prettier": "^8.1.0",
     "eslint-plugin-office-addins": "^0.2.2",
     "eslint-plugin-prettier": "^3.3.1",
+    "office-addin-cli": "^1.1.3",
```
akrantz, line 34: Correct, this should have always been a runtime dependency.

packages/eslint-config-office-addins/package.json
millerds: The config package needs to be deprecated since it's redundant.  All it's doing right now is point to the config information in the plugin package.  It's not using the linter package and so there is not dependency for it.

packages/office-addin-cli/package.json
millerds: office-addin-lint depends on office-addin-cli so a dependency the other way around would create a circular dependency.  This is a way to get the linting/prettier stuff in the project w/out the dependency on office-addin-lint.

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -111,11 +111,11 @@ export = {
         for (let ref = 0; ref < references.length; ref++) {
           const reference = references[ref];
 
-          if (isNullObjectNode(reference.identifier.parent)) {
+          if (isNullObjectNode(reference.identifier.parent as TSESTree.Node)) {
```
igor-ribeiiro, line 114: I think a better way than to explicit cast could be to test the type right?
Like: 
```
if (reference.identifier.parent
&& isNullObjectNode(reference.identifier.parent)) {
```

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
```diff
@@ -129,7 +129,7 @@ export = {
               fix: function (fixer: RuleFixer) {
                 var ruleFix: RuleFix;
                 if (
-                  isInBinaryNullTest(reference.identifier) &&
+                  isInBinaryNullTest(reference.identifier as TSESTree.Identifier) &&
```
igor-ribeiiro, line 132: Here you could test like:
```
if (reference.identifier.type === TSESTree.Identifier
&& isInBinaryNullTest(reference.identifier)
```

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
millerds: The problem is just a compilation one as a result of change dependency.  The comparison stated complaining because it was say type defined in one package was not the same as a nested package type . . . but the nested package was the same as the higher level.  This just forces them to look at the same package.

packages/eslint-plugin-office-addins/src/rules/test-for-null-using-isNullObject.ts
millerds: This is a technicality with dependencies and the cast ensures the nest package is cast into the upper level package, though they are the same package.

igor-ribeiiro: I know this code you made works on this case. But I think it might have some potential bugs like:
What if `reference.identifier` was not a `TSESTree.Identifier`? This explicit cast could make bad issues appear.

millerds: Typescript types are what helps us make sure the types are correct.  The reason for the cast is because after adjusting the 
dependencies the compiler complains about:

eslint-plugin-office-addins/node_modules/@typescript-eslint/types/dist/ast-spec Identifier 
. . . not being the same as . . . 
eslint-plugin-office-addins/node_modules/@typescript-eslint/typescript-estree/node_modules/@typescript-eslint/types/dist/ast-spec

Note that they are the same, it just that one seems to be resolving to the package nested in typescript-estree and the other is not.

igor-ribeiiro (approved):
Looks good to me. Would be great to also get approve from @akrantz

millerds:
Looks like I can simplify things a bit by adjusting the imports to a different typescript-eslint package.

igor-ribeiiro (approved):

PR #479, opened 2021-06-23
Eslint don't call empty load

igor-ribeiiro:
This is the rule: 
"
Calls of load() should specify only the properties needed.
If the specific properties aren't specified, all scalar properties are loaded.
"


packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,34 @@
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+
+export = {
+  name: "no-empty-load",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"problem",
+    messages: {
+      emptyLoad:
+        "Calling an empty load can slow down your add-in",
```
millerds, line 9: "calling an empty load" might be better explained by saying "Calling load without any argument".

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,34 @@
...
+    },
+    docs: {
+      description:
+        "Calling an empty load can cause load unneeded data and slow down your add-in",
```
millerds, line 13: "calling an empty load" might be better explained by saying "Calling load without any argument".

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,34 @@
...
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Best Practices",
+      recommended: <false | "error" | "warn">false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#calling-load-without-parameters-not-recommended",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    return {
+      "CallExpression[callee.property.name='load']:not([arguments>Literal])"(
```
millerds, line 24: Without type information we can't say that the load call is actually coming from an office object . . . but in other rules we are searching for object that use get methods in order to determine if they are the variables we are looking for.

igor-ribeiiro: Sure, I will try to make this check. Will take more time now as the code won't be just a query search.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,90 @@
...
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Best Practices",
+      recommended: <false | "error" | "warn">false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#calling-load-without-parameters-not-recommended",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    function isEmptyLoad(node: TSESTree.MemberExpression): boolean {
+      if (isLoadFunction(node)) {
```
millerds, line 29: Should be able to collapse this into a return statement with 3 "anded" (&&) tests

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,90 @@
...
+      return false;
+    }
+
+    function findEmptyLoad(scope: Scope) {
+      scope.variables.forEach((variable: Variable) => {
+        let getFound: boolean = false;
+        variable.references.forEach((reference: Reference) => {
+          const node: TSESTree.Node = reference.identifier;
+
+          if (
```
millerds, line 45: If you look at what I figured out for call-sync-before-read you could use the isWrite() method and writeExpr property of the reference to combine these two set's of if blocks into one simpler one.  The outer if would check for isWrite() and the inner block would generically check isGetFunction with writeExpr as the argument.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,90 @@
...
+      >"Best Practices",
+      recommended: <false | "error" | "warn">false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#calling-load-without-parameters-not-recommended",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    function isEmptyLoad(node: TSESTree.MemberExpression): boolean {
+      if (
+        isLoadFunction(node) &&
```
millerds, line 30: Do a direct return of the conditional expression result.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,90 @@
...
+            // If reference was not related to a previous get
+            return;
+          }
+
+          if (
+            node.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression &&
+            isEmptyLoad(node.parent)
+          ) {
+            context.report({
+              node: node,
```
millerds, line 75: Shouldn't this be the expression rather then just the identifier?

igor-ribeiiro: Sure, it does makes more sense that way

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,84 @@
...
+          if (
+            reference.isWrite() &&
+            reference.writeExpr &&
+            isGetFunction(reference.writeExpr) &&
+            reference.resolved
+          ) {
+            getFound = false; // In case of reassignment
+            if (
+              reference.writeExpr &&
+              reference.resolved &&
```
millerds, line 51: You don't need this since you aren't using it inside a method call like I did.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    function findEmptyLoad(scope: Scope) {
+      scope.variables.forEach((variable: Variable) => {
+        let getFound: boolean = false;
+        variable.references.forEach((reference: Reference) => {
+          const node: TSESTree.Node = reference.identifier;
+
+          if (
+            reference.isWrite() &&
```
millerds, line 43: Only need this on the outer loop since you only want to find out if the var is being written to before clearing the boolean.

millerds: You still have the other two check you don't need on the outer loop.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -0,0 +1,79 @@
...
+            reference.writeExpr &&
+            isGetFunction(reference.writeExpr)
+          ) {
+            getFound = false; // In case of reassignment
+            if (isGetFunction(reference.writeExpr)) {
```
millerds, line 48: You'll need "reference.writeExpr" as part of this conditional.

igor-ribeiiro: Added a test case for this condition and I think now it is fixed

millerds: As is it works . . . the only question is weather you need to check reference.writeExpr in the outer if or the inner if.  As far as I can tell it's needed for the isGetFunction call to work, but do you have another reason for checking it earlier?

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
igor-ribeiiro: I just put the check in the inner loop, it indeed makes more sense that way.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
igor-ribeiiro: I guess inner makes more sense as well, just updated it

millerds (approved):

PR #480, opened 2021-06-24
Remove requirement to be in run method for no-context-sync-in-loop rule

millerds:
It might flag more then desired, but I think it's better then potentially missing a bunch.


igor-ribeiiro (approved):
I do agree, we shouldn't need to check if it's inside a run.
Maybe in the future could be another rule for that.

PR #481, opened 2021-06-24
Call sync before read rule

millerds:
New linter rule to check for context.sync() before reading object that were retrieved via get methods.


packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -0,0 +1,130 @@
...
+        ) {
+          if (isLoadReference(reference.identifier)) {
+            apiReferences.push({ operation: "Load", reference: reference });
+          } else {
+            apiReferences.push({ operation: "Read", reference: reference });
+          }
+        }
+      });
+
+      for (var i = 0; i < scope.childScopes.length; i++) {
```
igor-ribeiiro, line 81: Could you use a ForEach here? I think it makes it more clear and consistent with the rest of this file

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -0,0 +1,130 @@
...
+        findReferences(scope.childScopes[i]);
+      }
+    }
+
+    function findReadBeforeSync(): void {
+      const needSync: Variable[] = [];
```
igor-ribeiiro, line 87: Does `needSync` need to be an array? I see you use the operation `includes` which I guess can take linear time.
Could you try using a set instead?

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+      });
+    }
+
+    return {
+      Program(
+        programNode: TSESTree.Node /* eslint-disable-line no-unused-vars */
+      ) {
+        findReferences(context.getScope());
+        apiReferences.sort((a, b) => {
```
igor-ribeiiro, line 121: Can you change `a`  `b` to `left` `right` just for more clarity?

packages/eslint-plugin-office-addins/test/rules/call-sync-before-read.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      Excel.run(function (context) {
+        var selectedRange = context.workbook.getSelectedRange();
+        selectedRange.load('address');
+        console.log('The selected range is: ' + selectedRange.address);
+        await context.sync();
+      });`,
+      errors: [{ messageId: "callSync", data: { name: "selectedRange" } }]
+    },
+    {
+      code: `
```
igor-ribeiiro, line 82: Could you make a test where `selectedRange` loads the `address`, there is a `context.sync`, but `selectedRange2` is the one that makes the read operation? This is to check if the code is handling different variables at the same time.

igor-ribeiiro (changes requested):

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
millerds: It didn't work when I was doing something slightly different, but looks like it's good now.

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
millerds: I had not considered this.  Looks like a good way to do it.

packages/eslint-plugin-office-addins/test/rules/call-sync-before-read.test.ts
millerds: This rule doesn't check the loads . . . just the reads of the original variable.  I think the load being called before being used is part of other rules and maybe this cases you mention here (needing to check the sync) is the P1.5 rule we have yet to start.

igor-ribeiiro (approved):

PR #484, opened 2021-06-28
Load sync read rule

igor-ribeiiro:
Creating the rule to check get->load->sync->read. 
The rule needs to have all these components.

I've also put more functions in the utils file and a few refactors on the other rules.


packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
```diff
@@ -0,0 +1,81 @@
+import { TSESTree } from "@typescript-eslint/experimental-utils";
+import { Variable } from "@typescript-eslint/experimental-utils/dist/ts-eslint-scope";
+import { findReferences, OfficeApiReference } from "../utils";
+
+export = {
+  name: "load-sync-read",
```
millerds, line 6: How about call-sync-after-load

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
```diff
@@ -0,0 +1,81 @@
...
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"suggestion",
+    messages: {
+      loadSyncRead:
+        "Call load on '{{name}}' for '{{loadValue}}' followed by context.sync() before reading the object or its properties",
```
millerds, line 11: I would phrase it "Call context.sync() after calling load on '{{name}}' property on the '{{loadValue}}' and before reading the property"

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    docs: {
+      description:
+        "Always call load on an object followed by a sync before reading it or one of its properties.",
```
millerds, line 15: I think we need to be clearer about calling the sync after the load call and not just the object with the property.

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
```diff
@@ -0,0 +1,81 @@
...
+      >"Best Practices",
+      recommended: <false | "error" | "warn">false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#load",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    let apiReferences: OfficeApiReference[] = [];
+
+    function findReadLoadSync(): void {
```
millerds, line 27: You're looking to loads without a sync before the read.  This name doesn't seem to indicate that.

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
```diff
@@ -0,0 +1,81 @@
...
+        "Always call load on an object followed by a sync before reading it or one of its properties.",
+      category: <
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Best Practices",
+      recommended: <false | "error" | "warn">false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model?view=powerpoint-js-1.1#load",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
```
millerds, line 24: This looks like it trying to track the sync for the object and some loads, but I don't think it needs to be that complex and I don't see how it's tracking the loaded properties.  You just want to track load calls without a sync before reading the property . . . you don't need to track the root object.

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
igor-ribeiiro: Added the context.sync() function to the message. Tell me if there are any suggestions

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
igor-ribeiiro: Rephased, just used "properties" instead of "property"

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
igor-ribeiiro: I do need to check if the object is a proxy (get*) object, that's why I'm tracking this.
I'm not really tracking the property that was loaded. Specific properties should be found by another rule:
`load-object-before-read`

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -0,0 +1,89 @@
...
+  OfficeApiReference,
+} from "../utils";
+
+export = {
+  name: "call-sync-after-load",
+  meta: {
+    type: <"problem" | "suggestion" | "layout">"suggestion",
+    messages: {
+      callSyncAfterLoad:
+        "Call context.sync() after calling load on '{{name}}' property on the '{{loadValue}}' and before reading properties",
```
millerds, line 15: End with "property" instead of "properties" (matches the rest of the sentence)

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
millerds: I think the descript is still unclear about what it is checking.

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
millerds: Specific property read after load is found by another rule . . . but not the sync after property load but before the read.  Saying that now makes me wonder if the two should be combined (read after load become read after sync after load).

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -0,0 +1,89 @@
...
+      const needSync: Set<Variable> = new Set<Variable>();
+      const needLoadAndSync: Set<Variable> = new Set<Variable>();
+
+      apiReferences.forEach((apiReference) => {
+        const operation = apiReference.operation;
+        const reference = apiReference.reference;
+        const variable = reference.resolved;
+
+        if (operation === "Write" && variable) {
+          needLoadAndSync.add(variable);
```
millerds, line 41: We don't always have to call load after a write . . . which is what this is making me think . . . you can get and object without loading a property of it.  I think what you mean to track is variables that could have load calls.  But the fact that they are in the apiReference already determined that . . . so I'm not sure why that is needed here.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        const propertyName: string | undefined = findPropertiesRead(
+          reference.identifier.parent
+        );
+
+        if (
+          operation === "Read" &&
+          variable &&
+          (needSync.has(variable) || needLoadAndSync.has(variable))
+        ) {
```
millerds, line 63: Tracking the sync of the variable is what the other rule does . . . this doesn't need to to that.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -0,0 +1,89 @@
...
+        const operation = apiReference.operation;
+        const reference = apiReference.reference;
+        const variable = reference.resolved;
+
+        if (operation === "Write" && variable) {
+          needLoadAndSync.add(variable);
+        }
+
+        if (operation === "Load" && variable) {
+          if (needLoadAndSync.has(variable)) {
```
millerds, line 45: You should be tracking the load not the variable that is get*().  Thinks like this won't get caught:

var property = worksheet.getItem("sheet");
property.load("values");
await context.sync();
console.log(property.values);
property.load("length");
console.log(property.length);

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -115,7 +53,7 @@ export = {
       Program(
         programNode: TSESTree.Node /* eslint-disable-line no-unused-vars */
       ) {
-        findReferences(context.getScope());
+        apiReferences = findReferences(context.getScope());
```
millerds, line 56: Make sure this works well.  I was trying to do something link this when I was developing this rule and it wasn't working (array was getting reset or emptied unexpectedly).

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
millerds: I'm also wondering if more rules should be combined.  Keeping them separate allows for more targeted rules to turn on and off and combing them would make for a complex rule (especially if we added a config to allow adjusting exactly what is checked), but we seem to be doing repeated work here and it can get confusing which rule is supposed to track what.  I'm not sure what the answer is for combine or not . . . but we should be targeted while keeping seperate.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: I'm trying to detect the following pattern:
get()* variables (need a write) -> load -> sync -> read

All these steps must exist for the rule to apply successfully. If we don't check for the first get()* variables, we can get false positives if people have a load funcion on their projects

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: I am now tracking the load and also added your test case

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
igor-ribeiiro: Not sure what exactly was the problem you were having, but this code structure is working fine now

packages/eslint-plugin-office-addins/src/rules/load-sync-read.ts
igor-ribeiiro: I guess this rule is still different than the others and can detect the unique error as following:
get -> sync -> load -> read
Which is a unique error that only appear to this rule.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
millerds: If the only reason for the get*() tracking is to verify the load call is legit, then that is already taken care of by  findReferences(context.getScope()) which narrows down the list of references to those that should be from the office api.  You don't need to then track it again when looking for the load -> sync -> read pattern.

igor-ribeiiro: The only purpose of the get*() is indeed to check if the load call is legit. You are right, I do not need the get* tracking and just removed it.

packages/eslint-plugin-office-addins/src/utils.ts
```diff
@@ -173,3 +178,89 @@ export function isLoadFunction(node: TSESTree.MemberExpression): boolean {
...
+  return (
+    node.parent &&
+    node.parent.type === TSESTree.AST_NODE_TYPES.MemberExpression &&
+    isLoadFunction(node.parent)
+  );
+}
+
+let proxyVariables: Set<Variable>;
+let apiReferences: OfficeApiReference[];
+export function findReferences(scope: Scope): OfficeApiReference[] {
```
millerds, line 207: Maye this should be named "findOfficeApiReferences" since it's being made a util function (no contest to interpret the name with).

packages/eslint-plugin-office-addins/src/utils.ts
```diff
@@ -173,3 +178,89 @@ export function isLoadFunction(node: TSESTree.MemberExpression): boolean {
...
+  proxyVariables = new Set<Variable>();
+  apiReferences = [];
+  findReferencesAuxiliar(scope);
+  return apiReferences;
+}
+
+function findReferencesAuxiliar(scope: Scope): void {
```
millerds, line 214: Why the separate methods?

igor-ribeiiro: The "normal" method is only needed to initialize the global variables. By doing this, we can ensure that the function will work across multiple calls.
Without this separation, if a user would call this function twice, the second call would have wrong results (given the global variables were not empty).

millerds: Yes . . . but why is a wrapper method needed just to initialize global variables . . . why can't it be in the same method?

igor-ribeiiro: Like, the first call on the function needs to be different than the rest, given the necessity of initializing the global variables every time.
The wrapper method is kinda way I`ve seen used before to solve this kind of problem. Do you think of another way? I could pass an additional variable to the functions (to detect the first time the function gets called), but I think this gets messier.

millerds: Ah yes . . . I forgot about the recursive call.  Makes sense now.

Note that I tried passing in the array when I was doing it . . . but that was part of the trouble I had (mentioned in another comment) where the array wasn't keeping all the values (something with the way javascript loops work).  The global variable was the way to keep everything straight.

Personally I don't like the use of "Aux" or "Auxiliary" in the method name.  I think findReferencesInScope or findOfficeApiReferencesInScope or something like that makes more sense.

igor-ribeiiro: Sure, I just did the refactoring

packages/eslint-plugin-office-addins/test/rules/call-sync-after-load.test.ts
```diff
@@ -0,0 +1,45 @@
...
+        var property = worksheet.getItem("sheet");
+        property.load("values");
+        await context.sync();
+        console.log(property.values);`
+    },
+    {
+      code: `
+        var fakeGet = worksheet.notAGetFunction("props");
+        await context.sync();
+        property.load("props");
```
millerds, line 21: Should also have a test case where fakeGet.load("props") is used.

igor-ribeiiro: Sure, that's the test I wanted to make, just had made a typo

millerds: Having both would be good (one test for load without any get and one tests for the fake get call)

millerds (approved):

PR #486, opened 2021-06-29
Run npm install with updated repo package versions

millerds:
Updated scripts package versions and the lock files referencing each other need to be updated to match.


package.json
```diff
@@ -9,9 +9,9 @@
     "audit": "npm audit & npm run exec -- npm audit",
     "audit:fix": "npm audit fix & npm run exec -- npm audit fix",
     "build": "lerna run build",
-    "clean-install-packages": "npm run exec -- npm ci & npm ci",
+    "clean-install-packages": "npm run exec -- npm ci",
```
akrantz, line 12: But this won't run lerna bootstrap after. Why are you removing the top-level `npm ci`?

package.json
```diff
@@ -9,9 +9,9 @@
...
     "dist-tags": "npm run exec -- npm dist-tag ls",
-    "install-packages": "npm run exec -- npm install & npm install",
+    "install-packages": "npm run exec -- npm install",
```
akrantz, line 14: Same goes for this...

akrantz (changes requested):
See question about script changes.

package.json
millerds: This was accidental (good catch).  I did this because running npm install at the root and running it in the package directories does conflicting things.  Running at the root seems to remove some of the package dependencies and if you go into the package directory and run install after running at the root it will always find things that needs to be installed,  I meant to keep this as a local change until I had a chance to figure out what was going on and make a separate change.

package.json
millerds: will revert.

package.json
akrantz: Lerna at least allows you to do normal npm commands (install, etc.) in the package directory (unlike Rush). You just need to do "lerna bootstrap" at the root afterwards to symlink the packages. I made this a post-install script, so it happens automatically after npm install at the root.

packages/custom-functions-metadata-plugin/package-lock.json
```diff
@@ -1458,371 +1420,23 @@
       "dev": true
     },
     "eslint-plugin-office-addins": {
-      "version": "0.2.2",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-office-addins/-/eslint-plugin-office-addins-0.2.2.tgz",
-      "integrity": "sha512-sinVHbsUUfOOiVRmcPeH+U1v7C0LBNFZKD8GczOR0nfHlFo81A0u5rNw8lMUJduDFhlJeyZmBH1n87rbID96LQ==",
+      "version": "0.3.0",
```
akrantz, line 1423: Just a comment that packages which are using version 0.x, if they are ready, it might be good to bump them up to 1.0.0. With semver, if the major version is zero (first number), then the minor version (second number) is treated such that it won't automatically pick up minor version number updates. When being developed, it's therefore better to start with 0.x, but once it is ready for production, it probably is better to move to 1.x.

akrantz (approved):

packages/custom-functions-metadata-plugin/package-lock.json
akrantz: No changes needed as part of this pr (in case it wasn't clear)...

PR #487, opened 2021-06-30
Docs fix

igor-ribeiiro:
For some reason, the description field was empty in the repository now.
I tried to find where this change was made, but it wasn't in the original PR nor in the PRs after that, no idea why this happened.

Anyway, I'm updating this again and I also am putting some .lock changes that were generated by running npm install command.

### DIFF
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
@@ -9,7 +9,7 @@ export = {
       callSync: "Call sync before trying to read '{{name}}'.",
     },
     docs: {
-      description: "",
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -9,7 +9,7 @@ export = {
       callSync: "Call sync before trying to read '{{name}}'.",
     },
     docs: {
-      description: "",
+      description: "Always call load on an object followed by a context.sync() before reading it or one of its properties.",
```
millerds, line 12: The load is only needed for one of it's properties and not the object itself.

millerds (approved):

PR #490, opened 2021-07-06
Fix sync after load rule bug

igor-ribeiiro:
Fixed the load after load rule bug where it was also firing when there was no sync (only load).
We don't want to fire in this case, as it would create a case where one code would fire twice with rule call-sync-before-read (any code missing sync but that had load).

### DIFF
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
@@ -68,11 +68,14 @@ export = {
           const propertyName: string = findPropertiesRead(
             reference.identifier.parent
           );
-          const variableProperty: VariableProperty = { variable: variable.name, property: propertyName };
+          const variableProperty: VariableProperty = {
+            variable: variable.name,
+            property: propertyName,
+          };
           if (
-            needSync.has(variableProperty) 
-            && wasLoaded.has(variableProperty) 
-            && hasSync
+            needSync.has(variableProperty) &&
+            wasLoaded.has(variableProperty) &&
+            hasSync
           ) {
             const node = reference.identifier;
             context.report({
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
@@ -43,6 +43,8 @@ export = {
 
     function findLoadBeforeSync(): void {
       const needSync: VariablePropertySet = new VariablePropertySet();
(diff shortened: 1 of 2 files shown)


millerds:
I don't think this change is correct.  The whole point of it is to check for a sync after a load call specifically.  The other rule is looking for a sync after a get.

igor-ribeiiro:
> I don't think this change is correct. The whole point of it is to check for a sync after a load call specifically. The other rule is looking for a sync after a get.

I know intuitively it makes sense to keep this rule as sync after load, but here we are already checking for a get* and read variables, and this makes both rules really similar.

This case for example:
```
        var property = worksheet.getItem("sheet");
        property.load("values");
        console.log(property.values);
```
Should only fire the missing sync rule right? Right now it would fire 2 errors on the different rules.

millerds:
> > I don't think this change is correct. The whole point of it is to check for a sync after a load call specifically. The other rule is looking for a sync after a get.
> 
> I know intuitively it makes sense to keep this rule as sync after load, but here we are already checking for a get* and read variables, and this makes both rules really similar.
> 
> This case for example:
> 
> ```
>         var property = worksheet.getItem("sheet");
>         property.load("values");
>         console.log(property.values);
> ```
> 
> Should only fire the missing sync rule right? Right now it would fire 2 errors on the different rules.

It should fire both . . . the sync is missing for both cases.  There are cases in general eslint where the same line of code will fire multiple similar but different rules.  You can have two errors on the same line of code and a single change can fix multiple rule violations.

If this really is a concern, then this would suggest more that we should consider combining the rules.

igor-ribeiiro:
> > > I don't think this change is correct. The whole point of it is to check for a sync after a load call specifically. The other rule is looking for a sync after a get.
> > 
> > 
> > I know intuitively it makes sense to keep this rule as sync after load, but here we are already checking for a get* and read variables, and this makes both rules really similar.
> > This case for example:
> > ```
> >         var property = worksheet.getItem("sheet");
> >         property.load("values");
> >         console.log(property.values);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should only fire the missing sync rule right? Right now it would fire 2 errors on the different rules.
> 
> It should fire both . . . the sync is missing for both cases. There are cases in general eslint where the same line of code will fire multiple similar but different rules. You can have two errors on the same line of code and a single change can fix multiple rule violations.
> 
> If this really is a concern, then this would suggest more that we should consider combining the rules.

It makes sense, I think it is fine to leave the rule as it was. I will close this PR for now, we can reopen in the future if needed.

PR #491, opened 2021-07-07
Navigation scalar eslint rule

igor-ribeiiro:
Creating the rule to check for load with navigation properties.

The 

- changes to `.eslintrc.json ` are needed to make lint not complain about enums: [Eslint Enum False Positive](https://github.com/typescript-eslint/typescript-eslint/issues/2621)
- The changes to `tsconfig.json` are needed to be able to import `json` from files: [Import JSON](https://www.codevscolor.com/import-json-typescript)
- `utils.ts` was simply changed to a different directory (utils) and split into `utils/utils.ts` and `utils/getFunctions.ts`. No changes were made on the new `utils.js` other than this refactor and splitting the get functions.

@Rick-Kirkham can you check the texts of the documentation in the rule `src/rules/no-navigational-load.ts `? This is related to rule 15.


Rick-Kirkham (changes requested):

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+      node.property.type === TSESTree.AST_NODE_TYPES.Identifier
+    ) {
+      propertyName += node.property.name + "/";
+    }
+    node = node.parent;
+  }
+  return propertyName.slice(0, -1);
+}
+
+export function getPropertyNameInLoad(node: TSESTree.Node | undefined): string {
```
millerds, line 90: This is actually just getting the string argument of any method call that happens directly off of the object.  It's not specific to "load".  It would also return an error for sub reference like <object>.<property>.forEach(...) which is valid code.

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (
+    node &&
+    node.type === TSESTree.AST_NODE_TYPES.MemberExpression &&
+    node.parent?.type === TSESTree.AST_NODE_TYPES.CallExpression &&
+    node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.Literal
+  ) {
+    return node.parent.arguments[0].value as string;
+  }
+  return "error in getPropertyNameInLoad";
```
millerds, line 99: This is confusing to return an error string vs. a property name string.  The caller couldn't tell which one it got.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
```diff
@@ -0,0 +1,88 @@
...
+              return;
+            }
+          }
+
+          if (!getFound) {
+            // If reference was not related to a previous get
+            return;
+          }
+
+          if (node.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression) {
```
millerds, line 65: This assumes that every member expression for an object is a load . . . which is not true.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
```diff
@@ -0,0 +1,88 @@
...
+      category: <
+        "Best Practices" | "Stylistic Issues" | "Variables" | "Possible Errors"
+      >"Best Practices",
+      recommended: <false | "error" | "warn">false,
+      url: "https://docs.microsoft.com/en-us/office/dev/add-ins/develop/application-specific-api-model#scalar-and-navigation-properties",
+    },
+    schema: [],
+  },
+  create: function (context: any) {
+    function trimPropertyName(propertyName: string): string {
```
millerds, line 31: I'm not sure what the value of this is.  Seems like you only getting part of the supposed property and throwing away potentially good information.

It also looks like you are handling the property in two different ways (with the other being findPropetiesRead())

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
```diff
@@ -0,0 +1,88 @@
...
+      return propertyName.split("/").pop() ?? "";
+    }
+
+    function isNotScalarLoad(
```
millerds, line 35: I prefer using a positive in method names and using a negation outside.  !isScalarLoad(...) is easier to read than isNotScalarLoad().

Also . . . what is a "ScalarLoad" seems like you're trying to combine different idea together that don't need to be.  I think it's clearer to test for isLoad and isScalar separately and then let a conditional check for both.

packages/eslint-plugin-office-addins/src/utils/propertiesType.ts
```diff
@@ -0,0 +1,30 @@
...
+  notProperty,
+}
+
+const navigationProperties: Set<string> = new Set<string>(
+  propertiesTypeJson.navigational
+);
+const scalarProperties: Set<string> = new Set<string>(
+  propertiesTypeJson.scalar
+);
+const undefinedProperties: Set<string> = new Set<string>(
```
millerds, line 16: What is an undefined property?

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
```diff
@@ -0,0 +1,65 @@
+import { ESLintUtils } from '@typescript-eslint/experimental-utils'
+import rule from '../../src/rules/no-navigational-load';
+
+const ruleTester = new ESLintUtils.RuleTester({
+	parser: '@typescript-eslint/parser',
+});
+
+ruleTester.run('no-navigational-load', rule, {
```
millerds, line 8: Need to include test with nested property references like range.format.font.load("color") from the spread sheet.  That one is a valid case, but something like range.format.load("font") would be invalid.  The spreadsheet also notes the use of range.load({format: { font: { color: true } } }) being a valid case.

millerds (changes requested):

packages/eslint-plugin-office-addins/src/utils/propertiesType.ts
igor-ribeiiro: Undefined property is when you can't determine if a property is navigational or scalar just by the property name.
For example, the property `format` can be both navigational or scalar, depending on who called it.

I felt like it would add a lot of extra complexity to handle those cases and it would not be worth it, as we only have 9 "undefined" properties, but more than 550 Scalar or navigational.

millerds: In that case might say "ambiguous" which suggests it's unclear instead of "undefined" which sounds like something that doesn't exist.

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
igor-ribeiiro: The code right is only checking for the last property and checking if the same of this property is navigational or scalar.
So like, it can detect that range. format.load("font") would be invalid if the font is characterized as a navigational property.

Then the case currently not supported is the range.load({format: { font: { color: true } } }), which I will work on

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
igor-ribeiiro: Yeah, this Trim is indeed throwing away useful information. It is together with the idea of the undefined properties, in which I only look to the last properties if it is navigational or scalar.

The findPropetiesRead actually builds the property name when the property is read. This trim propertyName works when the property is loaded, not read. I will change the name to address this.

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: I will refactor to not depend on the load.
What do you mean by this error for the sub references like forEach?

packages/eslint-plugin-office-addins/src/utils/propertiesType.ts
igor-ribeiiro: Makes sense, I do agree it would be a better name (ambiguous), because that's the case

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
millerds: I looked up what the '/' is used for in argument and apparently it's short had for some navigation properties (I didn't know that, but maybe you did).  So instead of range.format.font.load("color") someone could call range.load("format/font/color") . . . is that why you are checking just the last element?  If so that it makes a little more sense, though you might want to check the earlier part are actual navigational elements.

Also I'm wondering in multiple properties can't be asked for in a single load statement and what that might look like.  We should account for that.

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: Looks like my example got interpreted wrong.  I'm mean object.arrayProperty.forEach(...) or arrayObject.forEach(...).  Those things that might get flagged as an error in this method when really they are just not the right thing to be looking at.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
igor-ribeiiro: Yeah, I knew about this syntax using "/" and that's what I was accounting for.
I agree that checking if the middle properties are navigational is useful and I will be doing that.

I also don't know if the load can load multiple properties. I guess using the syntax of loading an object:
item.load({font: color, font: size} could be used to load multiple scalar properties, but I'm not sure.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -0,0 +1,57 @@
...
+        composedProperty +=
+          "/" + composeObjectExpressionPropertyIntoString(property.value);
+      }
+    }
+  });
+
+  return composedProperty;
+}
+
+export function getLoadArgument(node: TSESTree.Node | undefined): string {
```
millerds, line 40: This method still assumes load by name and one of the argument parsing, but doesn't check that it's actually a load statement.

It also assumes a single argument, but load can have more then one.

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
millerds: Should have a test case for that nested property

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
```diff
@@ -0,0 +1,90 @@
...
+          if (!getFound) {
+            // If reference was not related to a previous get
+            return;
+          }
+
+          if (
+            node.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression &&
+            isLoadFunction(node.parent)
+          ) {
+            const propertyName: string = getLoadArgument(
```
millerds, line 68: What about loads with multiple properties.

millerds (changes requested):

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
igor-ribeiiro: I created a similar one which is already checked in:
```
var range = context.workbook.getRange();
                range.load({borders: { fill: { color: true } } });
                if (myRange.borders.fill.color);
```
This one specifically (with format), would not pass as format is one of the 9 properties considered ambiguous.

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: Yeah, I didn't know `load` could have multiple arguments until very recently.
I guess we could try to change it to accept multiple arguments in another PR because some other rules also use this function and we would need to see how they do interact with it.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
igor-ribeiiro: As mentioned previously, I think we should create another task for that (I can take it after this PR).

packages/eslint-plugin-office-addins/src/utils/load.ts
millerds: The multiple args could be done in another PR, but name should reflect what it's doing . . . extracting args, not specifically load args.  The caller is making sure it's load args.

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
millerds: I'm talking about a range.borders.fill.load("color") . . . not the object format.

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: Now the rule is checking for load

millerds: The rule yes . . . but not the method.  This is a utility method and so there is not context of it being part of a load call.  Another rule could call this without checking for load on another method and the name wouldn't make sense.

millerds: Sorry, missed the check-in.  Makes sense now.

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
igor-ribeiiro: Actually didn't work before.
I added the code to detect these cases now in the utils/load and also added some test cases.

The nice thing of using utils is that this bug also existed for other rules (they never worked on navigational) and now all of the rules using load will also accept all formats of load.

millerds (approved):

PR #492, opened 2021-07-14
Use typescript@4.3 and ts-node@10.1

jozefizso:


packages/custom-functions-metadata/src/generate.ts
```diff
@@ -739,9 +739,9 @@ function findTag(node: ts.Node, tagName: string): ts.JSDocTag | undefined {
 /**
  * If a node contains the named tag, returns the tag comment, otherwise returns "".
  */
-function getTagComment(node: ts.Node, tagName: string) {
+function getTagComment(node: ts.Node, tagName: string): string {
   const tag = findTag(node, tagName);
-  return tag && tag.comment ? tag.comment : "";
```
akrantz, line 744: Could this be updated to `return tag?.comment?.toString() || ""` ?  Do you think it is better that way?

akrantz (approved):
Thanks for the PR! This looks good. I made one suggestion about a code change.

akrantz (changes requested):
Seems like it's better for me to mark this as request changes. Mention me if you make an update or if you don't want to, and I can follow up then.

jozefizso:
@akrantz I change the code as requested.

akrantz (approved):

PR #493, opened 2021-07-20
Fix read object check

millerds:
Rule looking for sync before read shouldn't flag cases where the property is being set instead of read.  Corrected to looking for top level object in property chain and check for assignment vs. actual read.


packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -22,6 +24,16 @@ export = {
   create: function (context: any) {
     let apiReferences: OfficeApiReference[] = [];
 
+    function checkPropertyRead(node: TSESTree.MemberExpression): boolean {
```
igor-ribeiiro, line 27: I don't think the name of this function is very informative on what it does.
`checkPropertyIsRead` would make it better

igor-ribeiiro (changes requested):

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
millerds: Done

igor-ribeiiro (approved):

PR #494, opened 2021-07-20
Run npm install and ci on each package

millerds:
Individual package package-lock.json files seem to be out of sync with package.json.  This should bring them back into sync.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -81,6 +81,12 @@
         }
       }
     },
+    "@hutson/parse-repository-url": {
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/@hutson/parse-repository-url/-/parse-repository-url-3.0.2.tgz",
+      "integrity": "sha512-H9XAx3hc0BQHY6l+IFSWHDySypcXsvsuLhgYLUGywmJ5pswRVQJUHpOsobnLYp2ZUaUlKiKDrgWWhosOwAEM8Q==",
+      "dev": true
+    },
     "@lerna/add": {
       "version": "4.0.0",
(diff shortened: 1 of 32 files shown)


igor-ribeiiro (dismissed):

akrantz:
I usually would title a PR like this as "update package-lock.json files".

akrantz:
@millerds I had no problems with npm install or running tests locally. I ran `npm audit fix` and pushed the commit to see what the updated packages do.

akrantz:
There are errors about "filename too long" when checking out on the Windows machine.

millerds:
> Usually this import would start with a lowercase letter...

The import is based on the name of the namespace I used.  I debated on using lower vs. upper case and saw that other objects imports were using upper (like ManifestInfo) so I used upper for the namespace.

akrantz:
That's because ManifestInfo is a specific type, but since this is using * and is package, then convention is to have it be lowercase. The package name is also lowercase, but uses camelcase because you cannot have hyphens in the name.

millerds:
> Do you really have to import this way for the sinon fakes to work?

Multiple places online indicated this problem with an updated version of typescript without any real solution.

The updated to typescript meant it wasn't generating code that would allow sinon to replace the method calls gotten by a "import * from ..." statement (done to me more inline with ES6).  By putting the methods in a namespace like this I can import using a "import {...} from ..." statement which get typescript to generate the coded needed.  I don't know of another workaround.

millerds:
> That's because ManifestInfo is a specific type, but since this is using * and is package, then convention is to have it be lowercase. The package name is also lowercase, but uses camelcase because you cannot have hyphens in the name.

Can't use the import * when needing to use stubs (see other comment).

igor-ribeiiro (approved):

PR #513, opened 2021-08-17
Prompt for running script if SSO configuration has already occurred

millerds:
If you run configure-sso a second time you break the original configuration, but you aren't told that until the end of the script after everything is done except for some manual pieces that are needed.  This will confirm that you want to go through with the changes before hand so things aren't accidentally broken.


packages/office-addin-sso/src/commands.ts
```diff
@@ -17,6 +18,19 @@ export async function configureSSO(manifestPath: string) {
   if (process.platform !== "win32" && process.platform !== "darwin") {
     console.log(chalk.yellow(`${process.platform} is not supported. Only Windows and Mac are supported`));
     return;
+  } else if (applicationDataConfigured(manifestPath)) {
+    console.log(
+      chalk.yellow("Project has already been at least partially configured.  Can't setup new configuration completely.")
```
akrantz, line 23: Please check with the content team about the phrasing here. It's missing articles like "the" and I think the tone could be improved.

igor-ribeiiro (changes requested):

packages/office-addin-sso/src/commands.ts
millerds: Anyone specific on the content team I should be asking?

packages/office-addin-sso/src/commands.ts
```diff
@@ -17,6 +18,19 @@ export async function configureSSO(manifestPath: string) {
...
+    );
+    const question = {
+      message: `Continue anyway?`,
+      name: "didUserConfirm",
+      type: "confirm",
+    };
+    const answer = await inquirer.prompt([question]);
+    if (!(answer as any).didUserConfirm) {
```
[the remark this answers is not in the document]
millerds: The example in the answers listed there shows that calling a method (and I presume a property) is not valid for "unknown" type . . . which is exactly what I'm trying to do here.  

This was actually copied from another location in our code.  I will double check if there is not a better way to do this.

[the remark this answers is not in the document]
igor-ribeiiro: I guess just changing `any` to `unknown` here should work, but I didn't test it.
Maybe the rest of the code still has a lot of `any`s, I just think unknown would be better as it is more restricted. 
But anyway, it is a very minor deal.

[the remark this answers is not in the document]
akrantz: It is better to use `unknown` rather than `any` but even better is to have actual type info. I believe the if statement listed here can be written without using unknown or any. What is the type of the answer variable?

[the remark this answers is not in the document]
millerds: The 'prompt' method returns 'any' type.  I figured out I don't need the cast and have a local change to remove it.

packages/office-addin-sso/src/commands.ts
igor-ribeiiro: Not anyone specific, but @millerds I will send to you in private some contacts of the contents team

packages/office-addin-sso/src/commands.ts
```diff
@@ -20,15 +20,15 @@ export async function configureSSO(manifestPath: string) {
     return;
   } else if (applicationDataConfigured(manifestPath)) {
     console.log(
-      chalk.yellow("Project has already been at least partially configured.  Can't setup new configuration completely.")
+      chalk.yellow("Project was already previously updated and will require manual steps when done.")
```
akrantz, line 23: What manual steps are needed? This isn't a very useful message for the user.

Consider something like. "Since the project has already been configured, please validate the configuration. See https://aka.ms/office-addin-sso for more information."  (You can create an aka.ms link to point to documentation. This allows it to be changed later if needed and not hardcoded.)

packages/office-addin-sso/src/commands.ts
millerds: Updated the text to be similar to the existing text.

packages/office-addin-sso/src/commands.ts
millerds: There is an existing warning at the end of the script that says what needs to be done.  That is the one I missed when trying to test things and I thought the template was broken (it wasn't).  I wanted to add this at the beginning so that we wouldn't break an existing setup unless explicitly told to do so.

millerds: I simplified the message and left the manual step info to the existing warning at the end of the script.

akrantz (approved):

millerds:
Mac failures is not in package that is affected by this change.

igor-ribeiiro:
> Mac failures is not in package that is affected by this change.

I guess you can just push the changes. Mac is breaking because of another issue and we have to look into it.

akrantz:
Please fix the mac issue first, so this can pass all of the checks before being merged.

millerds:
The mac test pass locally on my macbook pro.  I think it's a problem with the CI/CD machine (I'm currently trying to set up a new one).

PR #515, opened 2021-08-25
Adding telemetry to eslint rules for office addin

igor-ribeiiro:
Added telemetry code to check when a rule gets fired. This will help us monitor the frequency developers are hitting an issue.

Tested by running the `npm run lint` command on task pane add-in and checking that the telemetry is getting reported.

I needed to remove usage-data and cli dependency on eslint-office-addin because it was causing a circular dependency issue, which made the `install` command fail.


packages/office-addin-usage-data/.eslintrc.json
```diff
@@ -1,8 +0,0 @@
-{
-  "plugins": [
-    "office-addins"
-  ],
-  "extends": [
-    "plugin:office-addins/recommended"
-  ]
-}
```
millerds, line 8: What linting is left after removing the office plugin?  We should make sure we are only removing the plug-in and not the rest.

packages/office-addin-cli/package.json
```diff
@@ -33,7 +33,6 @@
     "@typescript-eslint/eslint-plugin": "^4.26.0",
     "concurrently": "^6.0.0",
     "eslint": "7.28.0",
-    "eslint-plugin-office-addins": "^1.0.0",
```
millerds, line 36: What's the dependency chain that causes office-addin-cli to need to remove the plug-in?

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -79,21 +80,26 @@ export = {
...
+          apiReferences.sort((left, right) => {
+            return (
+              left.reference.identifier.range[1] -
+              right.reference.identifier.range[1]
+            );
+          });
+          findLoadBeforeSync();
+        } catch(err: any) {
+          usageDataObject.reportException("call-sync-after-load", err);
+        }
```
millerds, line 102: So if I understand correctly we are logging success after being able to report one instance of a customer code problem, an error if the rule code failed, and nothing if the rule code ran but no error in the user customer code was found.  What questions about customer behavior do we hope to answer with these data points?

igor-ribeiiro: The main question we are trying to answer is:
"What are the most common mistakes customers make when writing their code with Office js API".
Checking the errors is more for us to catch any wrong implementation because we actually do not throw any errors, so it is more to check if our code might be faulty at some part.

packages/office-addin-cli/package.json
igor-ribeiiro: eslint-plugin-office-addins -> office-addin-usage-data -> office-addin-cli -> eslint-plugin-office-addins

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
millerds: So what happens with live linting in VSCode . . . does the same rule trigger send the same telemetry until the customer fixes it?  We would need to make sure we aren't getting repeated signal on the same error.

packages/office-addin-cli/package.json
igor-ribeiiro: This is what gets logged when I run `npm run build`:
eslint-plugin-office-addins -> office-addin-usage-data -> office-addin-cli -> eslint-plugin-office-addins

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: Yeah, this would lead to having more data overall, but I couldn't think of a way to make this not happen.

My idea was that customers either ignore or fix the error (then we wouldn't get the telemetry anymore). If they don't, then telemetry will keep getting sent.
I thought this wouldn't be that bad, given that the absolute number of times the rule was hit doesn't matter much. What really matters is the count between each rule.

packages/office-addin-usage-data/.eslintrc.json
igor-ribeiiro: Sure, I added the lint functionalities back again. Tested them with some examples and ran `npm run lint` on them to check that they are working.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
millerds: That would actually sku to someone who isn't fixing the error vs. someone who does.  It could be that something comes up as more common if a handful of people never fix the error vs. and error that people hit and fix.  Do we end up catering to people who aren't fixing flagged error vs. those who are finding the error reporting useful and fixing them.

I'm also wondering about the volume itself.  If the telemetry is getting logged during live linting we could be getting way too much signal in just one customer session.

It also just occurred to me that I'm not sure what happens to the telemetry if someone puts the comment in to ignore the rule (does the rule still continue to fire, silently hid the error, and telemetry get logged).

packages/office-addin-cli/.eslintrc.json
```diff
@@ -1,8 +1,23 @@
...
+      "prettier/prettier": ["error", { "endOfLine": "auto" }],
+      "no-eval": "error",
+      "no-delete-var": "warn",
+      "no-octal": "warn",
+      "no-inner-declarations": "warn",
+      "no-unused-vars": "off",
+      "@typescript-eslint/no-unused-vars": "error"
+    }
+  }
+
```
millerds, line 23: Where did this set come from?

igor-ribeiiro: I took it from our `eslint-plugin-office-addins` module.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: I just tested with live linting and what happens is that the linter runs around every 10 seconds.
This means that we would get a lot, but really of a lot of noise data in a way that I don't quite think would be worth data.

I think now I will look into the telemetry of Office-Addin-Lint because this we can control we it gets called.

igor-ribeiiro:
I will be closing this PR for now, as I don't think there is a way to have telemetry on the rules because of the live linting frequency. Will investigate what we do on office-addin-lint.

PR #518, opened 2021-09-02
Add quotes around linter command args

millerds:
This will fix the problem with the lint command when projects have a space in the name or path and for the linter running on Mac.

### DIFF
--- a/packages/office-addin-lint/src/lint.ts
+++ b/packages/office-addin-lint/src/lint.ts
@@ -15,6 +15,7 @@ const eslintConfigPath = path.resolve(__dirname, "../config/.eslintrc.json");
 const eslintTestConfigPath = path.resolve(__dirname, "../config/.eslintrc.test.json");
 
 function execCommand(command: string) {
+  console.log(`> ${command}`);
   const execSync = require("child_process").execSync;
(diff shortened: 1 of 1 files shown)


packages/office-addin-lint/src/lint.ts
```diff
@@ -15,6 +15,7 @@ const eslintConfigPath = path.resolve(__dirname, "../config/.eslintrc.json");
 const eslintTestConfigPath = path.resolve(__dirname, "../config/.eslintrc.test.json");
 
 function execCommand(command: string) {
+  console.log(`> ${command}`);
```
igor-ribeiiro, line 18: What is this about?

millerds: This just adds the actual command that was run as part of the output.  It's useful to see the paths that are part of the command . . . but maybe not necessary.

akrantz: I think you should take this out.

millerds:
The mac test pass locally on my macbook pro.  I think it's a problem with the CI/CD machine (I'm currently trying to set up a new one).

igor-ribeiiro (approved):

PR #519, opened 2021-09-09
Adding a check to see if load is empty

igor-ribeiiro:
Fixing the exception that fired when load is empty. The error was being fired on other rules and not on the `emptyLoad` rule, that's why our tests didn't catch the error.

### DIFF
--- a/packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
@@ -5,7 +5,7 @@ import {
   Variable,
 } from "@typescript-eslint/experimental-utils/dist/ts-eslint-scope";
 import { isGetFunction } from "../utils/getFunction";
-import { isLoadFunction } from "../utils/load";
+import { getLoadArgument, isLoadFunction } from "../utils/load";
 
 export = {
   name: "no-empty-load",
@@ -27,11 +27,7 @@ export = {
   },
   create: function (context: any) {
     function isEmptyLoad(node: TSESTree.MemberExpression): boolean {
-      return (
-        isLoadFunction(node) &&
-        node.parent?.type == TSESTree.AST_NODE_TYPES.CallExpression &&
-        node.parent.arguments.length === 0
-      );
+      return isLoadFunction(node) && getLoadArgument(node) === "";
     }
 
     function findEmptyLoad(scope: Scope) {
--- a/packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
@@ -27,7 +27,7 @@ export = {
   },
   create: function (context: any) {
     function isEmptyLoad(node: TSESTree.MemberExpression): boolean {
-      return isLoadFunction(node) && getLoadArgument(node) === "";
+      return isLoadFunction(node) && !getLoadArgument(node);
     }
 
     function findEmptyLoad(scope: Scope) {
--- a/packages/eslint-plugin-office-addins/src/utils/load.ts
+++ b/packages/eslint-plugin-office-addins/src/utils/load.ts
@@ -52,17 +52,15 @@ export function getLoadArgument(
   ) {
     if (node.parent.arguments.length === 0) {
       return undefined;
-    } else {
-      if (node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.Literal) {
-        return node.parent.arguments[0].value as string;
-      } else if (
-        node.parent.arguments[0].type ===
-        TSESTree.AST_NODE_TYPES.ObjectExpression
-      ) {
-        return composeObjectExpressionPropertyIntoString(
-          node.parent.arguments[0]
-        );
-      }
+    }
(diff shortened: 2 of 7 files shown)


packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -27,11 +27,7 @@ export = {
   },
   create: function (context: any) {
     function isEmptyLoad(node: TSESTree.MemberExpression): boolean {
-      return (
-        isLoadFunction(node) &&
-        node.parent?.type == TSESTree.AST_NODE_TYPES.CallExpression &&
-        node.parent.arguments.length === 0
-      );
+      return isLoadFunction(node) && getLoadArgument(node) === "";
     }
```
millerds, line 31: Why are we using getLoadArgument instead the other two test cases (which seem more to the point of the rule)?  Wouldn't it be simpler to just add the 'getLoadArgument(node) === ""' and an '||' with the length test (in parenthesis)?

millerds (changes requested):

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
igor-ribeiiro: It was like that before, I just think it is better to reutilize the functions we already have, given that getLoadArgument is doing this now.

igor-ribeiiro:
While changing the code to use undefined for empty load arguments, I managed to also fix the false positive when the navigational property rule was reporting on empty loads. I also added tests for it.

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
millerds: It was doing the check for call expression before and we still had it here . . . though we weren't checking for the empty string before.

I guess if we are updating the meaning of "empty load" to include empty string then this could make sense.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,21 +41,28 @@ function composeObjectExpressionPropertyIntoString(
...
-    } else if (
-      node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.ObjectExpression
-    ) {
-      return composeObjectExpressionPropertyIntoString(
-        node.parent.arguments[0]
-      );
+    if (node.parent.arguments.length === 0) {
+      return undefined;
+    } else {
+      if (node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.Literal) {
```
millerds, line 56: We should be able to flatten the nested conditional so it's less complicated to follow.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,21 +41,28 @@ function composeObjectExpressionPropertyIntoString(
...
+        return node.parent.arguments[0].value as string;
+      } else if (
+        node.parent.arguments[0].type ===
+        TSESTree.AST_NODE_TYPES.ObjectExpression
+      ) {
+        return composeObjectExpressionPropertyIntoString(
+          node.parent.arguments[0]
```
millerds, line 63: I think I mentioned this before, but getting only the first property loose all checks of the other properties.  Do we have a separate bug for this?  Is there a plan to fix it?

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
igor-ribeiiro: Updating to include empty strings I guess makes sense, as it is the same error and same problem (of performance)

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: This is to see only the first argument of the load, which should be a string. So any string inside load will get caught by this `[0]`, even if the string represents multiples properties.

Example: `load("address, font");` still is only 1 string and all of it will be inside `arguments[0]`.

millerds: I think I was getting this confused with the method that takes the string and pulls out the different properties . . . I guess this is the reverse of that where we take an object and make a string (note your example of load("address, font") would get caught by the previous if statement since it's already a literal string).

The thought and question still applies though really in a different location.  We are using this method to get the argument and use it places like call-sync-after-load to identify properties, but we are treating like a single property.  It's not going to be addressed here, but do we have a follow up for that.

igor-ribeiiro: Sure makes sense. Yeah, I will later address the bug of us not supporting commas `load("x, y")` and I will look into previous rules to make sure they work for all types of loads.

millerds: Is there a task or bug tracking this work that needs to be done?

igor-ribeiiro: I was planning on doing it on this task:
load with multiple properties in the argument needs to be able to parse each property

I will add your comments in the description for us to not forget about supporting these other loads for all rules.

millerds (approved):

PR #520, opened 2021-09-10
Updating typescript eslint version

igor-ribeiiro:
Updating packages versions, because beforehand, the following warning would appear when running `npm run lint` command.  The warning goes away after updating those versions.

```
=============
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.3.0

YOUR TYPESCRIPT VERSION: 4.3.5

Please only submit bug reports when using the officially supported version.

=============
```

### DIFF
--- a/packages/eslint-plugin-office-addins/package-lock.json
+++ b/packages/eslint-plugin-office-addins/package-lock.json
@@ -5,35 +5,54 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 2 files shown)


millerds:
Did you verify the linting still worked as expected on our templates?  I don't know how well the unit tests cover linting.

igor-ribeiiro:
> Did you verify the linting still worked as expected on our templates? I don't know how well the unit tests cover linting.

Yeah, no problems on then. I don't think there is much of a reason linter would fail on the templates and not in Scripts, but anyway, I tested it :D

millerds (approved):

PR #521, opened 2021-09-14
Not reporting errors on ambiguous navigational properties

igor-ribeiiro:
The navigational eslint rule was having a lot of false positives regarding ambiguous properties (that can be navigational or scalar).
I'm putting it so that the rule won't fire a case of ambiguous rules to reduce false positives.


igor-ribeiiro:
@millerds the errors are on SSO Mac, which is unrelated to this change

millerds:
> @millerds the errors are on SSO Mac, which is unrelated to this change

Intune required an OS update to connect to corporate resources.  Passing now.

millerds (approved):

PR #522, opened 2021-09-16
Creating package to help unit testing with Office js

igor-ribeiiro:
Created the package `office-addin-unit-test` to help mock and unit test with Office.js
Samples can be seem on the test file.


packages/office-addin-unit-test/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "office-addin-unit-test",
```
millerds, line 2: I would suggest office-addin-mock or something like that.  'unit-test' is ambiguous about weather this is a test suit or the tool that we intend it to be.

packages/office-addin-unit-test/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "1.0.0",
```
millerds, line 3: It's not version one until is ready to publish for prime time (and I'm guessing it's not).  should probably do 0.0.1

packages/office-addin-unit-test/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "Provides unit testing utilities to help mock Office-js APIs",
```
millerds, line 4: Provides mocking support for office.js rather then utilities to help mock.

packages/office-addin-unit-test/src/defaults.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
millerds, line 1: defaults.ts?  The file names seems a bit vague.

And why does the usageData object need it's own file?

packages/office-addin-unit-test/src/officeJSMock.ts
```diff
@@ -0,0 +1,150 @@
+import { usageDataObject } from "./defaults";
+
+export class OfficeJSMock {
+  constructor(object?: OfficeObject) {
```
millerds, line 4: We should probably do a better job and the formal method headers . . . I think that aids in intellisense.

packages/office-addin-unit-test/src/officeJSMock.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  // Adds a function to OfficeJSMock
+  addMockFunction(methodName: string, functionality?: Function) {
+    this[methodName] = functionality ? functionality : function () {};
+  }
+
+  // Adds a OfficeJSMock to OfficeJSMock
+  addMockObject(objectName: string) {
```
millerds, line 19: I'm confuse at the intention of this method.  We're creating a new instance of ourself and allowing it to be done multiple times with different names?

packages/office-addin-unit-test/src/officeJSMock.ts
```diff
@@ -0,0 +1,150 @@
...
+      this[propertyName] = this.properties.get(propertyName)?.value;
+    }
+  }
+
+  private loadCalled() {
+    this.loaded = true;
+    this.value = `Error, context.sync() was not called`;
+  }
+
+  private loadMultipleProperties(property: string) {
```
millerds, line 66: 'loadMultipleProperties' but the argument is just one 'property'?

packages/office-addin-unit-test/src/officeJSMock.ts
```diff
@@ -0,0 +1,150 @@
...
+  // Adds a OfficeJSMock to OfficeJSMock
+  addMockObject(objectName: string) {
+    const officeJSMock = new OfficeJSMock();
+    officeJSMock.isMockObject = true;
+    this.properties.set(objectName, officeJSMock);
+    this[objectName] = this.properties.get(objectName);
+  }
+
+  // load of Office.js API
+  load(property: string) {
```
millerds, line 27: What does "load of Office.js API" mean?  Took me a while to figure out this was the replacement for the load call from the original office.js API.

Also don't see a need to have a nested private function instead of having it all directly here.

packages/office-addin-unit-test/src/officeJSMock.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  private assignValue(propertyName: string) {
+    if (this.properties.get(propertyName)?.isMockObject) {
+      this[propertyName] = this.properties.get(propertyName);
+    } else {
+      this[propertyName] = this.properties.get(propertyName)?.value;
+    }
+  }
+
+  private loadCalled() {
```
millerds, line 61: How does this work with multiple load calls?

packages/office-addin-unit-test/src/officeJSMock.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.value = `Error, context.sync() was not called`;
+  }
+
+  private loadMultipleProperties(property: string) {
+    property
+      .replace(/\s/g, "")
+      .split(",")
+      .forEach((individualProperties: string) => {
+        this.loadNavigational(individualProperties);
+      });
```
millerds, line 72: Is it properties or navigational?  I don't see the connection.  

In general it looks like your using the term property and properties in a mixed way and I'm uncertain at a given point in the code what it's referring to.

packages/office-addin-unit-test/test/officeJSMock.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    "getColor": function() {
+      return this.color;
+    },
+    "font": {
+      "size": 12
+    }
+  },
+}
+
+describe("OfficeJSMock", function() {
```
millerds, line 20: Could we get a better description.

packages/office-addin-unit-test/test/officeJSMock.test.ts
```diff
@@ -0,0 +1,96 @@
...
+      assert.strictEqual(officeJSMock.range.getColor(), "green");
+
+      officeJSMock.range.setMock("color", "yellow");
+      officeJSMock.range.load("color");
+      officeJSMock.sync();
+      assert.strictEqual(officeJSMock.range.getColor(), "yellow");
+    });
+  });
+
+  describe("Load", function() {
```
millerds, line 64: "Load" is a little vague

millerds (changes requested):

packages/office-addin-unit-test/src/defaults.ts
igor-ribeiiro: Yeah, I guess `default` is a bad name for `usageData` in general. All of our packages scripts currently use `default`, so I guess we keep it for now and flag this to be changed in another task.

packages/office-addin-unit-test/src/officeJSMock.ts
igor-ribeiiro: I don't think I quite understood what you mean here.
Do mean creating the .d.ts file?

packages/office-addin-unit-test/src/officeJSMock.ts
igor-ribeiiro: This method is to add a new object to our Office-js Mock.
Example:
```
range = new OfficeJSMock();
range.addMockObject("font");
```
Would create the following format:
```
{
    range: {
        font: {}
    }
}
```
Where both range and font are OfficeJSMocks.
I will change the commentary to explain better what the user needs to know about this function

packages/office-addin-unit-test/src/officeJSMock.ts
igor-ribeiiro: The name is wrong, but the function is accepting multiple properties. Fixing it

packages/office-addin-unit-test/src/officeJSMock.ts
igor-ribeiiro: Yeah, this is the replacement of the original load from office-js API. I will work on making it more clear in the comments.

About the nested private function, I think I still like having it explicitly there to make it easier to understand what the piece of code is doing (handling multiple properties). Given that `load` has to be the name of this function, having the private one with a more descriptive name seems more clear to me.

packages/office-addin-unit-test/src/officeJSMock.ts
igor-ribeiiro: It didn't. Added tests to check for multiple loads calls now and it is now working.

packages/office-addin-unit-test/src/officeJSMock.ts
igor-ribeiiro: Yeah, I do agree, it is kinda hard to name them all in a meaningful way. I changed some names to see if it gets a little easier to understand.

The problem is that we have the following structure:
- Complete properties, separated each by commas (note that completely property can be navigational or not)
- Complete property: navigational, which contains several properties and one scalar, or just a scalar.

packages/office-addin-unit-test/test/officeJSMock.test.ts
igor-ribeiiro: Are you thinking of anything? It kinda says the name of the object being tested and I felt it was fine.

packages/office-addin-unit-test/src/officeJSMock.ts
millerds: Maybe we aren't using these ourselves.  I remember seeing them somewhere and thought we should use them.  Something like this above each exported functions:
/**
 * Adds two numbers.
 * @param {number} num1 The first number to add.
 * @param {number} num2 The second number to add.
 * @return {number} The result of adding num1 and num2.
 */

packages/office-addin-unit-test/src/officeJSMock.ts
millerds: The load from Office.js does the same handling of multiple properties so it matches that.

I just dislike having to dig through many layers of code to find the meat of an operation.

packages/office-addin-unit-test/test/officeJSMock.test.ts
millerds: It's just the name of the object and doesn't give any context about the testing being done.

packages/office-addin-unit-test/src/officeJSMock.ts
millerds: The name "OfficeJSMock" sound like the top level mock of the whole thing rather than each layer within the whole API.  That is what is confusing me.

I also think now that we don't need the term "Object" in the name . . . just 'addMock' or something like that.

igor-ribeiiro: Yeah, "OfficeJSMock" is acting as both the top layer as each layer of objects. Every object is an OfficeJSMock. Do you think of any other name I could change it to to be more meaningful? I am thinking of OfficeJSMockObject if you think it would be better.

packages/office-addin-unit-test/test/officeJSMock.test.ts
igor-ribeiiro: Added a more descriptive name

packages/office-addin-unit-test/src/officeJSMock.ts
millerds: That name sill leaves me thinking it's the top level only,  Something like OfficeJSObjectMock sounds more like we're mocking one of the objects instead of just the top layer . . . thought I'm still not excited about that name.  Right now I'm also questioning the need for "JS" in the name . . . and maybe that is what is triggering the idea of the top level object.  Maybe something like "OfficeObjectMock" or "OfficeMockObject"?  These sound better to me, but I'm still not entirely sure.  Maybe @LouMM or @akrantz might have some thoughts here too.

igor-ribeiiro: I like the idea of removing JS from the name. This is all javascript, so it is already implicit that it is related to JS.
I changed the name to `OfficeMockObject` for now

packages/office-addin-mock/test/officeJSMock.test.ts
```diff
@@ -0,0 +1,144 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
```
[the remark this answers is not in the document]
igor-ribeiiro: I guess this comment was done before I changed the name of the test file. Could you please update the line this is referring to?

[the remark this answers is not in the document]
millerds: It's referring to the suit in general.  Maybe I missed it, but all the tests I was looking at seem to be reading values from the mock that was setup at the beginning without try to update in the middle of the test (think like our templates outputting code that set's the color of the cell).

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -0,0 +1,178 @@
...
+    this.value = `Error, property was not loaded`;
+    this.valueBeforeLoaded = value;
+    this.loaded = false;
+  }
+
+  private properties: Map<string, OfficeMockObject>;
+  private loaded: boolean;
+  private value: unknown;
+  private valueBeforeLoaded: unknown;
+  private isMockObject: boolean | undefined;
```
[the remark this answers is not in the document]
igor-ribeiiro: This variable refers to whether the `officeMockObject.property` will return a value or the object per say.
The question is that everything is an officeMockObject. Even you have a `range.address`, which address is a string, address will be an officeMockObject.

So it needs somehow to know that when calling `range.address`, it should return a string value and not the object.

I guess then a name like: "isTreatedLikeMockObject" is more detailed.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -0,0 +1,178 @@
...
+  /* eslint-disable-next-line */
+  [key: string]: any;
+}
+
+class OfficeObject {
```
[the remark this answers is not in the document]
igor-ribeiiro: It represents the Object to be used when populating Office JS with data. It was a suggested way to remove typescript errors, because the following error appears if I just use a normal object (instead of this class):
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'Object'.
No index signature with a parameter of type 'string' was found on type 'Object'.```

Other way to make code compile is to use a bunch of `as any`. So instead of calling: `objectData[property]`, it would have to call: `(objectData as any)[property]`.

The use of the class is then to make the code cleaner. I will change its name to match more its use.

File: packages/office-addin-mock/test/officeJSMock.test.ts (hunk 13 above)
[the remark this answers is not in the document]
Comment (igor-ribeiiro) [reply packages/office-addin-mock/test/officeJSMock.test.ts]: You are correct, the test suite was missing tests related to writing and setting of values. Updated it to include these now.

File: packages/office-addin-mock/src/officeMockObject.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (millerds) [reply packages/office-addin-mock/src/officeMockObject.ts]: I would simplify to something link "isObject", or "isObjectType", or "isProperty" (inverse the logic here), or something like that.  We don't need to say Mock as that is the context we are operating under.

File: packages/office-addin-mock/src/officeMockObject.ts (hunk 15 above)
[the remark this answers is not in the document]
Comment (millerds) [reply packages/office-addin-mock/src/officeMockObject.ts]: If it's public, add a description to explain what it is.

File: packages/office-addin-mock/test/officeMockObject.test.ts [hunk 16]
```diff
@@ -0,0 +1,153 @@
...
+      officeMock.range.load("color");
+      assert.strictEqual(officeMock.range.color, "blue");
+    });
+    it("Functions added", async function() {
+      const officeMock = new OfficeMockObject(testObject);
+      officeMock.range.load("color");
+      officeMock.sync();
+      assert.strictEqual(officeMock.range.getColor(), "blue");
+      officeMock.range.color = "green";
+      assert.strictEqual(officeMock.range.getColor(), "green");
```
millerds, line 95: I haven't tried it myself in the real Office.js code, but if I set a value does a sync need to happen for that to be committed?  I'm not sure that the correct code should be.

packages/office-addin-mock/src/officeMockObject.ts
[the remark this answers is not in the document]
igor-ribeiiro: It is not public, but I guess a description can't hurt.

packages/office-addin-mock/test/officeMockObject.test.ts
igor-ribeiiro: You don't need to sync when setting the values. I just tested on the taskpane project.
So the current implementation is fine and working.

packages/office-addin-mock/src/officeMockObject.ts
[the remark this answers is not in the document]
igor-ribeiiro: I changed to "isTreatedLikeObject" to have a more meaningful name.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -168,11 +167,12 @@ export class OfficeMockObject {
   private loaded: boolean;
   private value: unknown;
   private valueBeforeLoaded: unknown;
-  private isTreatedLikeMockObject: boolean | undefined;
+  private isTreatedLikeObject: boolean | undefined;
```
millerds, line 170: I think it can still be simpler like isObject (don't need the 'TreatedLike')

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -0,0 +1,179 @@
...
+    officeMockObject.isTreatedLikeObject = true;
+    this.properties.set(objectName, officeMockObject);
+    this[objectName] = this.properties.get(objectName);
+  }
+
+  /**
+   * Mock replacement of the load method in the Office.js API
+   * @param propertyArgument Argument of the load call. Will load any properties in the argument
+   */
+  load(propertyArgument: string | string[]) {
```
millerds, line 41: Is it possible for the load call to have multiple arguments?  How would we account for a variable list of arguments?

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -0,0 +1,179 @@
...
+      const officeMockObject = new OfficeMockObject();
+      officeMockObject.isTreatedLikeObject = false;
+      this.properties.set(propertyName, officeMockObject);
+    }
+    this.properties.get(propertyName)?.resetValue(value);
+    this[propertyName] = this.properties.get(propertyName)?.value;
+  }
+
+  /**
+   * Mock replacement for the sync of Office.js API
```
millerds, line 67: Mock replacement for the sync method in the Office.js API

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.properties.forEach((property: OfficeMockObject, key: string) => {
+      property.sync();
+      this.assignValue(key);
+    });
+    if (this.loaded) {
+      this.value = this.valueBeforeLoaded;
+    }
+  }
+
+  private assignValue(propertyName: string) {
```
millerds, line 79: what's the value getting assigned?  I think the method name is confusing.

packages/office-addin-mock/src/officeMockObject.ts
igor-ribeiiro: It can have just a string or an array of strings. I am accounting for that.
If it is an array, it will do a `forEach` on each of the properties. If is just a string, we simply do a normal load on it.

packages/office-addin-mock/src/officeMockObject.ts
igor-ribeiiro: I do agree. This method is to "assign" the property to be called using this.propertyName.
Changed the name to `makePropertyCallable`

packages/office-addin-mock/src/officeMockObject.ts
millerds: The linter rule code that parses the load method is looking for multiple method arguments instead of just one . . . so I'm not sure which one is accurate.

igor-ribeiiro: What do mean about multiple method arguments? They should be the same

millerds: In the linter rule's load.ts "parseLoadArguments(node: TSESTree.MemberExpression)" method the line "node.parent.arguments.forEach" looks for for more then just one argument.  So the linter rule is looking for multiple arguments but this mock code is not.

igor-ribeiiro: Probably you are mistaken in some outdated code PR version. This is how the code is right now:
```
    if (typeof propertyArgument === "string") {
      this.loadMultipleProperties(propertyArgument);
    } else {
      propertyArgument.forEach((property: string) => {
        this.loadMultipleProperties(property);
      });
```
There is a forEach here.

millerds: That foreach is looking at at each property in an array of string that came from a single method argument.  In the linter rule you are looking at multiple method arguments (and after looking more closely at the linter rule I don't think we're looking for a string array as an argument)

igor-ribeiiro: Both cases should be looking at receiving one string that represents multiples properties: `load("address, font");` or multiple arguments: `load("address", "font");`. There are test cases for both of them, so they are covered. 

In the linter rules:
```
		{
			code: `
			  range.load("font/fill/color, address");
			  await context.sync();
			  console.log(range.font.fill.color);
			  console.log(range.address);`
		},
		{
			code: `
			  range.load("font/fill/color", "address");
			  await context.sync();
			  console.log(range.font.fill.color);
			  console.log(range.address);`
		},
```

In the mocking:
```
    it("Multiple properties load", async function() {
      const officeMock = new OfficeMockObject(testObject);
      officeMock.load("range/color, range/font/size");
      officeMock.sync();
      assert.strictEqual(officeMock.range.getColor(), "blue");
      assert.strictEqual(officeMock.range.font.size, 12);
    });
    it("Comma separated load", async function() {
      const officeMock = new OfficeMockObject(testObject);
      officeMock.load(["range/color", "range/font/type", "range/font/size"]);
      officeMock.sync();
      assert.strictEqual(officeMock.range.getColor(), "blue");
      assert.strictEqual(officeMock.range.font.type, "arial");
      assert.strictEqual(officeMock.range.font.size, 12);
    });
```

millerds: Your test's for the linter and mocking are different and pass because they are checking for what they are coded for.
'load(["font/fill/color", "address"]);' is not the same as 'load("font/fill/color", "address");'

Add this test to your mock:
    it("Multi argument load", async function() {
      const officeMock = new OfficeMockObject(testObject);
      officeMock.load("range/color", "range/font/type", "range/font/size");
      officeMock.sync();
      assert.strictEqual(officeMock.range.getColor(), "blue");
      assert.strictEqual(officeMock.range.font.type, "arial");
      assert.strictEqual(officeMock.range.font.size, 12);
    });

And this test to the linter:
		{
			code: `
			  range.load(["font/fill/color", "address"]);
			  await context.sync();
			  console.log(range.font.fill.color);
			  console.log(range.address);`
		},

I suspect they will both fail.

igor-ribeiiro: Thanks for noticing that. The eslint one is wrong, we can't have multiple load arguments like that. I will work on that on the eslint thread. The mocking is the correct way to load, so we can proceed with this PR

millerds (approved):

PR #523, opened 2021-09-22
Fix linter errors

millerds:
Should allow the new pipeline to pass


igor-ribeiiro (approved):

.azure-devops/full-pipeline.yml
```diff
@@ -7,8 +7,12 @@ jobs:
     pool:
       name: OE-OfficeClientApps
     steps:
+      - task: Npm@1
+        displayName: "Package level npm install"
+        inputs:
+          command: custom
+          customCommand: "run exec -- npm install"
       - template: ./steps/npm-install.yaml
-      - template: ./steps/lint.yml
```
igor-ribeiiro, line 11: Given that we are not referencing the lint.yml file anymore, do we still need it?

millerds: I was plan on trying to fix it later on.  I suppose I could remove the file to keep things clean in the mean time.

igor-ribeiiro (approved):

PR #524, opened 2021-09-22
Making Eslint package support loading multiple properties

igor-ribeiiro:
Adding support for linter to work with loads with commas.


packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -43,25 +43,47 @@ function composeObjectExpressionPropertyIntoString(
...
     ) {
-      return composeObjectExpressionPropertyIntoString(
-        node.parent.arguments[0]
-      );
+      return undefined;
     }
+
+    let properties: string[] = [];
+    node.parent.arguments.forEach(
+      (propertiesSeparatedByComma: TSESTree.CallExpressionArgument) => {
```
millerds, line 63: 'propertiesSeparatedByComma' is not accurate give it may not be a literal and if it is it may not be separated by commas.  I would use a simple 'argument' given that is what you are examining.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -43,25 +43,47 @@ function composeObjectExpressionPropertyIntoString(
...
-      return undefined;
-    }
-    if (node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.Literal) {
-      return node.parent.arguments[0].value as string;
-    } else if (
-      node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.ObjectExpression
+    if (
+      node.parent.arguments.length === 0 ||
+      (node.parent.arguments[0].type === TSESTree.AST_NODE_TYPES.Literal &&
+        (node.parent.arguments[0].value as string) === "")
```
millerds, line 56: This seems to assume that the first value is the only one that can be empty string.  Later on you are looking at all arguments instead of just the first and anyone of them could be an empty string (or even 'undefined' for that matter).

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -43,25 +43,47 @@ function composeObjectExpressionPropertyIntoString(
 
 export function getLoadArgument(
```
millerds, line 44: The name of this method is a off now that you are looking at all arguments and each argument could end up returning multiple fields that are going to be loaded.  Should use a name that reflects this like 'parseLoadArguments' or 'getLoadPropeties or 'getPropertyList' or something like that.

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
```diff
@@ -68,6 +68,22 @@ ruleTester.run('no-navigational-load', rule, {
...
 		},
+		{
+			code: `
+			  var range = worksheet.getSelectedRange();
+			  range.load("font/fill/color, address");
+			  await context.sync();
+			  console.log(range.font.fill.color);
+			  console.log(range.address);`
+		},
+		{
```
millerds, line 79: Am I missing something or is this just a repeat?

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
```diff
@@ -108,5 +124,23 @@ ruleTester.run('no-navigational-load', rule, {
...
+			code: `
+			  var range = worksheet.getSelectedRange();
+			  range.load("address, font/fill");
+			  await context.sync();
+			  console.log(range.font.fill.color);
+			  console.log(range.address);`,
+			  errors: [{ messageId: "navigationalLoad", data: { loadValue: "font/fill" } }]
+		},
+		{
```
millerds, line 136: What's the difference here?

millerds (changes requested):

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
igor-ribeiiro: There was no difference. They were duplicated. Deleted one of them.

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: I added these empty case strings in the empty load rule, so we can also flag them and warn the client.

millerds: I'm not sure how that applies.  You are checking here for no arguments and "", but only for the first argument ('node.parent.arguments[0].value').  Then the next thing you do is look at all the arguments of the function ('node.parent.arguments.forEach(...)'.  I'm suggesting that the check for "" (or even an explicit 'undefined') be done inside the loop.

I realize that it doesn't make sense from what the code should look like, but if that's what were looking at we should be thorough.

packages/eslint-plugin-office-addins/test/rules/no-navigational-load.test.ts
millerds: Now there is a very subtle but important difference. :-)

packages/eslint-plugin-office-addins/src/rules/no-empty-load.ts
```diff
@@ -27,7 +27,23 @@ export = {
   },
   create: function (context: any) {
     function isEmptyLoad(node: TSESTree.MemberExpression): boolean {
-      return isLoadFunction(node) && !getLoadArgument(node);
+      // return isLoadFunction(node) && !parseLoadArguments(node);
```
millerds, line 30: Unneeded comment

packages/eslint-plugin-office-addins/src/utils/load.ts
millerds: I see it now . . . it is better to look for the empty property in the rule instead of here since this should just parse the arguments.  If that is the case then you don't need to check for "" here at all.

igor-ribeiiro: I do agree. Moved all the checks to the empty load case.

millerds: Are you checking for arguments.length somewhere?  I think you might have a problem calling aruguments.forEach(...) when there are no arguments.

igor-ribeiiro: It is working even without checking for the length. In the `no-empty-load.test.ts` there are multiple examples with empty load and no problem there.

millerds (approved):

PR #526, opened 2021-09-23
Eslint methods being treated like properties

igor-ribeiiro:
Fixing the ESLint bug where methods would be treated like properties and demand `sync` or `load` to be called on them.
This on its own fixes the `delete` problem, the `getCell`, and all the cases using methods.


packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -30,6 +34,8 @@ export = {
       switch (topExpression.parent?.type) {
         case TSESTree.AST_NODE_TYPES.AssignmentExpression:
           return topExpression.parent.right === topExpression;
+        case TSESTree.AST_NODE_TYPES.CallExpression:
+          return isProperty(node);
```
millerds, line 38: The method is for checking "isRead" . . . now we are checking something else.  It doesn't belong here.

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
```diff
@@ -30,6 +34,8 @@ export = {
       switch (topExpression.parent?.type) {
         case TSESTree.AST_NODE_TYPES.AssignmentExpression:
           return topExpression.parent.right === topExpression;
+        case TSESTree.AST_NODE_TYPES.CallExpression:
```
millerds, line 37: I thought CallExpression itself indicated a method . . . why do we also need to check again?

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -72,12 +72,28 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
   scope.childScopes.forEach(findOfficeApiReferencesInScope);
 }
 
+export function isProperty(node: TSESTree.Node): boolean {
```
millerds, line 75: This doesn't make sense . . . you aren't testing if the node it a property, you are checking if there is a property in the arguments.

millerds (changes requested):

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
igor-ribeiiro: We didn't check for callExpression anywhere else... This is the first check for it because we want to know I the thing after the `.` is a property or a method.
If you are wondering why do we need to check with the `isProperty(node)`, it is because there are two types of methods:
`console.log(range.address)` will have a node type = callExpression as with `range.delete()`. So we need to somehow differentiate those two methods, as both have CallExpression on them.

The way I found was by checking with the `isProperty` function, which now I realize it should be renamed

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: Agree, the name of the function is wrong.

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
igor-ribeiiro: With the other changes, did this get solved? I didn't get this comment very much

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
millerds: Why does it matter if a method call has an argument or not?

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
millerds: Nope . . .

The switch statement is in a method call 'checkPropertyIsRead' and you are adding a case that has nothing to do with weather a property is being read or written to.

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: What is the purpose of checking if we are making a method call with a property argument (why does it matter)?

igor-ribeiiro: I guess the problem lies mostly in the function name. I made a few adjustments to make it check for calling a method and not a property, which makes more sense.

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
igor-ribeiiro: It refers to the case that in fact, it is a property that is there.
The method needs to check if the "thing" being read is a property or not and also check if it is being read after confirming it is a property. I still like the way it is, do you think it deserves some function name change?

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: Just fixed it as we discussed @millerds . The PR is ready to be reviewed again

packages/eslint-plugin-office-addins/src/rules/call-sync-before-read.ts
millerds: Took this offline to get a clear picture of our different points of view.

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -72,12 +78,23 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
   scope.childScopes.forEach(findOfficeApiReferencesInScope);
 }
 
+export function isCallingMethod(node: TSESTree.MemberExpression): boolean {
```
millerds, line 81: I would name it 'isMethod' or 'isMethodCall'

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -61,8 +61,14 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
       reference.resolved &&
       proxyVariables.has(reference.resolved)
     ) {
-      if (isLoadReference(reference.identifier)) {
+      const node: TSESTree.Node = reference.identifier;
+      if (isLoadReference(node)) {
         apiReferences.push({ operation: "Load", reference: reference });
+      } else if (
+        node.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression &&
+        isCallingMethod(findTopLevelExpression(node.parent))
```
millerds, line 69: I would do the findTopLevelExpression call inside the 'isCallingMethod' method and just pass in the node.

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -61,8 +61,14 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
       reference.resolved &&
       proxyVariables.has(reference.resolved)
     ) {
-      if (isLoadReference(reference.identifier)) {
+      const node: TSESTree.Node = reference.identifier;
+      if (isLoadReference(node)) {
         apiReferences.push({ operation: "Load", reference: reference });
+      } else if (
+        node.parent?.type === TSESTree.AST_NODE_TYPES.MemberExpression &&
```
millerds, line 68: Why are we looking at the parent and not the reference itself?

millerds (changes requested):

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: I like it `isCallingMethod` especially now after I put the `topLevelExpression` call inside the function, given that the argument of the functions is the object.
For example, in `range.getCell(0,0)`, the function receives `range` and it is not a method, it is just calling a method.

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -31,7 +31,7 @@ export function findTopLevelExpression(
 }
 
 export type OfficeApiReference = {
-  operation: "Get" | "Load" | "Sync" | "Read";
+  operation: "Get" | "Load" | "Method" | "Read" | "Sync";
```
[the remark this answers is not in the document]
igor-ribeiiro: Sure, I added some documentation into it

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -61,8 +61,14 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
       reference.resolved &&
       proxyVariables.has(reference.resolved)
     ) {
-      if (isLoadReference(reference.identifier)) {
+      const node: TSESTree.Node = reference.identifier;
```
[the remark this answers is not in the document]
igor-ribeiiro: I moved it. I think node is fine as pretty much every other rule refers to it as node.

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: It is because of the tree itself. This is how we do it normally when calling the `findTopExpression`. I just continued using what we use.

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: Is the the method here is determining if the argument is a method call (i.e. is 'getCell' a method => true while is 'range' a method => false) or is it determining if the argument is the caller of the method (i.e. is 'range' calling a method => true while is 'getCell' calling a method => false).  The name should match what the returning T/F is indicating . . . and the logic can be inverted to match the name.

I'm thinking we are passing in the reference and finding out if it is a method call or not.

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -31,7 +31,14 @@ export function findTopLevelExpression(
 }
 
 export type OfficeApiReference = {
-  operation: "Get" | "Load" | "Sync" | "Read";
+  /**
+   * Get: An OfficeJs object, which is created when calling a `get` type function
+   * Load: A reference to `object.load()` type call
+   * Method: The reference is calling a method. Ex: `object.methodCall()`
```
millerds, line 37: So we are labeling 'methodCall' as method and not 'object' correct?  This sounds like we're labeling 'object' as calling a method.

igor-ribeiiro: We are labeling the object as calling a method right now.

packages/eslint-plugin-office-addins/src/utils/utils.ts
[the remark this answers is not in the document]
millerds: Using a simple name work if the context makes it obvious.  I was a little unsure about this context, but I guess it works.

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: OK . . I'm forgetting my tree.  I thought the identifier was a CallExpression, but it's a type in and of itself (pointing to the actual property in question) and the parent is the CallExpression that is the 'property path' leading up to this one.  But by moving the 'findTobLevelExpression' into the method we can leave that method to worry about the CallExpression type.  Here we can just pass in the node to the method asking if this node is a method call.

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: I don't think that is correct.  We are examining in references and categorizing them, but this is suggesting not categorizing the reference itself, it's categorizing the parent of the reference.  We want to indicate that a reference is a method (not that it's parent is calling a method) and then act on that reference later when looking for things that need load and sync calls.

igor-ribeiiro: I dont know for sure how we could do this the way you are suggesting. The references, for example, in the code below:
```
range.getCell(0,0);
```
are only `range` (var range) and `range`(range.getCell). So we don't have a way to categorize `getCell`. We need to categorize the `range` in the `range.getCell` call and that is what is being done.

millerds: Why can't we categorize 'getCell'?  Isn't it a reference in our list of references and aren't we are currently categorizing it as a property without a load and sync (thus it is getting falsely flagged by the existing rules)?

igor-ribeiiro: It is not considered a reference, thus it is not categorized.

millerds: Then how is it getting picked up by "scope.references.forEach((reference) => ..." to test in the first place?  The method is examining every reference in the scope (including things like 'let 1 = 1; let x = 1' which have nothing to do with office.js) and deciding if it is something we should pay attention to later (adding it to the apiReferences collection) or not.  When adding it to the collection we further classify.

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -72,12 +82,28 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
   scope.childScopes.forEach(findOfficeApiReferencesInScope);
 }
 
+function isCallingMethod(node: TSESTree.Node | undefined): boolean {
```
millerds, line 85: Should do like the 'isLoadReference' method and pass in the identifier type and worry about the undefined in the next if statement

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: Your passing in the argument 'getCell' (which is why you are trying to walk up the tree to find range) so the argument isn't the caller, it's the method itself and should return false, but you are finding the parent 'range' node, identifying it as the caller and returning true . . . so the argument doesn't match the name.

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: I'm saying the reference picked up by the function is `range` and not the `getCell` call. So we can only categorize `range`, we don't get a reference for the method call name, only for the object that called it.

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: I'm not passing `getCell`, the argument is `range`. I figured that the call to `findTopLevelExpression` is not needed and that is why you are probably getting confused with.
I updated the code to not have it.

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: Changed it to not have to worry about identifier and also not needs to send `node.parent`. You now just pass `node` which makes it easier

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: OK . . . I had to go back and spend sometime remembering the code and tree.  I was getting my tree backwards thinking it was walking backwards from the getCell call, but actually the first segments in a series of properties is the inner most node in the tree rather then other outermost (which is actually where the method call name ends up).  I think you are correct that the 'range' node is the node being tested, but that is also why you **do** need the findTopLevelExpression call.

In reviewing the code I see that "isLoadFunction" and "isLoadReference" are more along the line of what I'm think this should look like (in name and type usage) . . . if fact, I'm thinking that the difference between "isLoadReference" & "isLoadFunction" and what I think should be "isMethodReference" & "isMethodFunction" (or "isMethodCallReference") is really just the test for the identifier node "name" property.  I would think about following the established pattern with some reuse between the methods.

packages/eslint-plugin-office-addins/src/utils/utils.ts
millerds: That seems to be the way we've been coding (see isLoadFunction) . . . we've opted for more strongly typed function arguments.  It does look like that pushes us to determine type sooner rather then later and possibly in more places, but I can't remember specifically why we ended up going that way (I do vaguely remember trying in out in each way).  We should be uniform so you could change the other calls to do it this way or you could change this to match the others.

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: Sure, I changed the designs to be consistent

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -72,12 +82,31 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
   scope.childScopes.forEach(findOfficeApiReferencesInScope);
 }
 
+function isMethodFunction(node: TSESTree.MemberExpression): boolean {
+  if (node.parent?.type === AST_NODE_TYPES.CallExpression) {
+    const callExpression: TSESTree.Node = node.parent;
```
millerds, line 87: I don't think you need a local variable for just one use in a single if statement and if you don't do that it can be a single conditional statement instead of two.  And if you are just returning a boolean that matches the result of if statement then you should just return the result of the conditional (more to the point).

If you really want to do it that way you should use the CallExpression type since that is what you just tested for

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -72,12 +82,31 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
   scope.childScopes.forEach(findOfficeApiReferencesInScope);
 }
 
+function isMethodFunction(node: TSESTree.MemberExpression): boolean {
```
millerds, line 85: I know I put this in my own comment, but reading it here make me realize "MethodFunction" is redundant :-p . . . it should just be "isFunction" or "isMethod"

Also . . . I think you still need the 'findTopLevelExpression' . . . otherwise this won't work for thing like Ref.prop.prop.method()

packages/eslint-plugin-office-addins/src/utils/utils.ts
```diff
@@ -72,12 +82,31 @@ function findOfficeApiReferencesInScope(scope: Scope): void {
...
+  if (node.parent?.type === AST_NODE_TYPES.CallExpression) {
+    const callExpression: TSESTree.Node = node.parent;
+    if (callExpression.callee === node) {
+      return true;
+    }
+  }
+  return false;
+}
+
+export function isMethodReference(node: TSESTree.Identifier) {
```
millerds, line 95: Did you consider code reuse between these and the load versions of them.  Seems like checking for load should be a special case of checking for a method.

millerds (changes requested):

packages/eslint-plugin-office-addins/src/utils/utils.ts
igor-ribeiiro: I guess it is better to keep them separated for now. The load is considered right now different than the method. 
We would have to change a lot of structures to be able to make load be a special case.

millerds: I'm not sure I follow changing structures . . . I'm just thinking about the 'isLoadFunction' calling this one instead of the duplicate if statement.  It's not a big deal.

millerds (approved):

PR #527, opened 2021-09-27
Revert version changes from bad publish

igor-ribeiiro:
Reverting two latest publish which both made the packages stay in a broken state.


akrantz (changes requested):
don't do this.

millerds:
> don't do this.

Care to explain?  It looks like changes were make to the master branch during a lerna commands that didn't work (Igor was trying to push a new package into the repo) and now the dependencies are broken within the repo (packages are trying to reference a version of another package that doesn't exist in the npm registry) and npm install doesn't work in some cases.  Seems like reverting the broken changes is the right way to go.

akrantz (commented):
I want some time to look at things...

PR #529, opened 2021-09-28
Fixing multiple argument load properties

igor-ribeiiro:
Fixing the bug where `load` was expecting multiple arguments. Instead, it should only expect a single argument, which can be a string with the commas inside or an array of strings.


packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,28 +41,46 @@ function composeObjectExpressionPropertyIntoString(
...
+
 export function parseLoadArguments(node: TSESTree.MemberExpression): string[] {
   node = findTopLevelExpression(node);
 
   if (
     isLoadFunction(node) &&
     node.parent?.type === TSESTree.AST_NODE_TYPES.CallExpression
   ) {
+    const argument = node.parent.arguments[0];
+    if (!argument) return [];
```
millerds, line 64: I believe it's a best practice to always use curly braces for if statement even if it's just one line.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,28 +41,46 @@ function composeObjectExpressionPropertyIntoString(
...
-          (argument.value as string)
-            .replace(/\s/g, "")
-            .split(",")
-            .forEach((property: string) => {
-              properties.push(property);
-            });
-        } else if (argument.type === TSESTree.AST_NODE_TYPES.ObjectExpression) {
-          properties.push(composeObjectExpressionPropertyIntoString(argument));
+    if (argument.type === AST_NODE_TYPES.ArrayExpression) {
+      argument.elements.forEach((args) => {
```
millerds, line 68: 'args' suggest a method call (and more then one value) rather then a single element in an array.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,28 +41,46 @@ function composeObjectExpressionPropertyIntoString(
...
+        if (args.type === TSESTree.AST_NODE_TYPES.Literal) {
+          properties = properties.concat(
+            parseLoadStringArgument(args.value as string)
+          );
         }
```
millerds, line 73: Can the array have the object expression in it (is it valid)?

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,28 +41,46 @@ function composeObjectExpressionPropertyIntoString(
...
+      });
+    } else {
+      if (argument.type === TSESTree.AST_NODE_TYPES.Literal) {
+        properties = properties.concat(
+          parseLoadStringArgument(argument.value as string)
+        );
```
millerds, line 79: This is handling the same thing as an element in an array . . . should share logic if you can.

packages/eslint-plugin-office-addins/src/utils/load.ts
```diff
@@ -41,28 +41,46 @@ function composeObjectExpressionPropertyIntoString(
   return composedProperty;
 }
 
+function parseLoadStringArgument(argument: string): string[] {
+  let properties: string[] = [];
+  argument
+    .replace(/\s/g, "")
+    .split(",")
```
millerds, line 48: Is this the only delimiter allowed?

millerds (changes requested):

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: Renamed to `arg` to reference just one. Do you feel it is enough?

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: It is not valid. The use of object is a separate case of load. I feel we still need more tests with it, I had created a task about supporting better this case of load: https://office.visualstudio.com/OC/_workitems/edit/5422476

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: This is just one line of code, it is just that the linter makes it 3 lines.
I don't quite think we need to share the logic, it just appears complex because the commands themselves are lengthy.

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: Yes, only commas

packages/eslint-plugin-office-addins/src/utils/load.ts
millerds: Personally I would prefer 'element' since it relates to the collection which is called 'elements' and 'arg' still sounds like a method/function parameter, but I suppose 'arg' can be acceptable.

packages/eslint-plugin-office-addins/src/utils/load.ts
millerds: So 'load([{font { size: 12} }])' isn't valid in an Office.js load call?  Seems like if it works as a direct argument in the load it could work as an element in the array . . . but I'm not the one that implemented it in Office.js.

packages/eslint-plugin-office-addins/src/utils/load.ts
millerds: I was referring to the whole if statement not just the method call.  

I was thinking that each element in the array has the same possibilities as an individual non-array load argument and therefore you could process an element the same as a non-array load argument.  But if that isn't the case (see the comment about the object as an array element) then it does become less compelling

Note that the one method is only because you moved that chunk of code to a method so you could use it in both locations.  It if made sense to move the if statements there too if would help to reduce the nesting of if, else if, and forEach code that gets a little tricky to follow.  But that depends on the other comment about Objects being valid in the array.

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: I like element! Nice suggestion, I will change for that

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: It is not valid. These are the parameters Load can receive (official docs):
https://docs.microsoft.com/en-us/javascript/api/excel/excel.chart?view=excel-js-preview#load_options_

packages/eslint-plugin-office-addins/src/utils/load.ts
igor-ribeiiro: Objects are not valid, so I guess we don't have much to change about this.

packages/eslint-plugin-office-addins/src/utils/load.ts
millerds: OK.

Seems like it might be nice to flag that that case in a rule (new one? existing one?) since it doesn't work, but seems like it should.  Will log a task for this.

Note: I'm noticing a lack of object style argument tests in rules that are parsing the load argument (like load-object-before-read).

millerds: Added a comment to that task to include more tests as well.

created: https://office.visualstudio.com/OC/_workitems/edit/5461502 for the object rule

millerds (approved):

PR #530, opened 2021-09-28
Add Mac and Linux pipeline jobs to new pipeline

millerds:


.azure-devops/full-pipeline.yml
```diff
@@ -12,3 +12,21 @@ jobs:
       - template: ./steps/devcerts.yml
       - template: ./steps/edgewebview.yml
       - template: ./steps/test.yml
+
+  - job: Mac
+    pool:
+      name: OPX Mac Dogfood Pool
```
igor-ribeiiro, line 18: Why are the names between the Windows, Mac, and Linux pipelines so different between themselves?

.azure-devops/full-pipeline.yml
```diff
@@ -12,3 +12,21 @@ jobs:
...
+    - template: ./steps/cleanup.yaml
+
+  - job: Linux
+    pool:
+      name: Azure Pipelines
+      vmImage: ubuntu-latest
+    steps:
+    - template: ./steps/npm-install.yaml
+    - template: ./steps/build.yaml
+    - template: ./steps/run-tests.yaml
```
igor-ribeiiro, line 32: The Mac and Linux jobs are using a different test template. The Windows is using `./steps/test.yml`. Why is that different?

.azure-devops/full-pipeline.yml
```diff
@@ -12,3 +12,21 @@ jobs:
...
+    - template: ./steps/cleanup.yaml
+
+  - job: Linux
+    pool:
+      name: Azure Pipelines
+      vmImage: ubuntu-latest
```
igor-ribeiiro, line 28: The Windows job doesn't need this `vmImage` argument as well?

.azure-devops/full-pipeline.yml
```diff
@@ -12,3 +12,21 @@ jobs:
...
       - template: ./steps/test.yml
+
+  - job: Mac
+    pool:
+      name: OPX Mac Dogfood Pool
+    steps:
+    - template: ./steps/npm-install.yaml
+    - template: ./steps/build.yaml
+    - template: ./steps/run-tests.yaml
+    - template: ./steps/cleanup.yaml
```
igor-ribeiiro, line 23: Is this a mac exclusive task?

igor-ribeiiro (changes requested):

.azure-devops/full-pipeline.yml
millerds: Those are the names of the agent pools they get the image from.  The windows case is 1ES and was named by jeff.  The mac case was named by Courtney for our custom pools, and the Linux case is a vanilla azure image.

.azure-devops/full-pipeline.yml
millerds: Good catch.  'run-tests' was the old way and uses npm to trigger, but 'test' uses a command so the logs are live instead of only after the command is finished.  Should be using 'test'.

.azure-devops/full-pipeline.yml
millerds: Nope . . . there is only one image in that pool.

.azure-devops/full-pipeline.yml
millerds: Sort of.  It's needed on the mac because we're reusing the hardware repeatedly.  The other two cases are single use vms that get tossed when the job is done.

igor-ribeiiro (approved):

PR #532, opened 2021-09-29
Adding support to load(*) to eslint rules

igor-ribeiiro:
Making eslint support the syntax of using load("*")

### DIFF
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
@@ -72,12 +72,9 @@ export = {
             reference.identifier.parent
           );
 
-          if (needSync.has({ variable: variable.name, property: "*" })) {
-            propertyName = "*";
-          }
-
           if (
-            needSync.has({ variable: variable.name, property: propertyName })
+            needSync.has({ variable: variable.name, property: propertyName }) ||
+            needSync.has({ variable: variable.name, property: "*" })
           ) {
             const node = reference.identifier;
             context.report({
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
@@ -68,9 +68,13 @@ export = {
         }
 
         if (operation === "Read" && variable) {
-          const propertyName: string = findPropertiesRead(
+          let propertyName: string = findPropertiesRead(
             reference.identifier.parent
           );
+
+          if (needSync.has({ variable: variable.name, property: "*" })) {
+            propertyName = "*";
+          }
 
           if (
             needSync.has({ variable: variable.name, property: propertyName })
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
@@ -13,7 +13,7 @@ export = {
     type: <"problem" | "suggestion" | "layout">"suggestion",
     messages: {
       callSyncAfterLoad:
-        "Call context.sync() after calling load on '{{name}}' for property '{{loadValue}}' and before reading the property.",
+        "Call context.sync() after calling load on '{{name}}' for '{{loadValue}}' and before reading the property.",
     },
     docs: {
       description:
--- a/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
@@ -13,7 +13,7 @@ export = {
     type: <"problem" | "suggestion" | "layout">"suggestion",
     messages: {
       callSyncAfterLoad:
-        "Call context.sync() after calling load on '{{name}}' for '{{loadValue}}' and before reading the property.",
+        "Call context.sync() after calling load on '{{name}}' for the property '{{loadValue}}' and before reading the property.",
     },
     docs: {
       description:
@@ -68,7 +68,7 @@ export = {
         }
 
         if (operation === "Read" && variable) {
-          let propertyName: string = findPropertiesRead(
+          const propertyName: string = findPropertiesRead(
             reference.identifier.parent
           );
 
--- a/packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
+++ b/packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
@@ -37,11 +37,24 @@ export = {
       return false;
     }
 
+    function hasBeenLoaded(
+      node: TSESTree.Node,
+      loadLocation: Map<string, number>,
+      propertyName: string
+    ): boolean {
+      if (
+        loadLocation.has(propertyName) && // If reference came after load, return
+        node.range[1] > (loadLocation.get(propertyName) ?? 0)
+      ) {
+        return true;
+      }
+      return false;
+    }
+
     function findLoadBeforeRead(scope: Scope) {
       scope.variables.forEach((variable: Variable) => {
         let loadLocation: Map<string, number> = new Map<string, number>();
         let getFound: boolean = false;
-        let loadStar: boolean = false;
 
         variable.references.forEach((reference: Reference) => {
           const node: TSESTree.Node = reference.identifier;
@@ -78,11 +91,7 @@ export = {
               // In case it is a load function
               const propertyNames: string[] = parseLoadArguments(node.parent);
               propertyNames.forEach((propertyName: string) => {
-                if (propertyName === "*") {
-                  loadStar = true;
-                } else {
-                  loadLocation.set(propertyName, node.range[1]);
-                }
+                loadLocation.set(propertyName, node.range[1]);
               });
               return;
             }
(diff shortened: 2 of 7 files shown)


packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -68,10 +68,14 @@ export = {
...
 
         if (operation === "Read" && variable) {
-          const propertyName: string = findPropertiesRead(
+          let propertyName: string = findPropertiesRead(
             reference.identifier.parent
           );
 
+          if (needSync.has({ variable: variable.name, property: "*" })) {
+            propertyName = "*";
+          }
```
millerds, line 77: I don't think we need to update the propertyName value like this . . . we just need do the test for variable.name + "*" in the next if statement (using '||').

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -98,6 +103,10 @@ export = {
             return;
           }
 
+          if (loadStar) {
```
millerds, line 106: I don't think we need the separate 'loadStar' tracking.  If I'm reading the pre-changed code right then the "*" property should be in the loadLocation map if it's there.  We should be looking for it as part of the previous if check which would also be checking the location and not just that there was a '*' used in a load somewhere.

I'm also not liking this expanding series of isolated if statements each just doing a return.  We should look at consolidating them is possible and possibly moving the 'isInsideWriteStatement' check earlier since it doesn't need the property and we can skip the extra processing.

We're getting an series of regular "if" statements here

millerds (changes requested):

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
```diff
@@ -43,6 +43,9 @@ export = {
         }
       }
 
+      if (lastProperty === "*") {
+        return true;
+      }
```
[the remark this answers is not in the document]
igor-ribeiiro: It should be after the loop because I want to catch cases like:
`range.load("font/size/*")` so we still need to validate that `font` and `size` are navigational properties.

packages/eslint-plugin-office-addins/test/rules/call-sync-after-load.test.ts
```diff
@@ -75,5 +82,12 @@ ruleTester.run('call-sync-after-load', rule, {
 			  console.log(range.font.fill.color);`,
```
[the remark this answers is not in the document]
igor-ribeiiro: There is something wrong in the way vscode vs github see indentation, because I see the code normally on my editor:
image

Something to look out for, I will see if github detects any other indentation error as well.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: That what I did before, the problem was that later we report the error based on `propertyName`. So if we just check for the existence of `*` the error message would be with a wrong `propertyName`.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
igor-ribeiiro: I agree that the chain of ifs is getting big and we shouldn't add more to it.
I added a new `hasBeenLoaded` to be reutilized and now the `loadStar` variable is not needed any longer.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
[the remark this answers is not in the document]
millerds: Have you checked that '*' at the end of the path like that is valid (by trying the code out)?

packages/eslint-plugin-office-addins/test/rules/call-sync-after-load.test.ts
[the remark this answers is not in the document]
millerds: It's likely a difference it tabs vs. spaces . . . that is usually what triggers a miss match since tab get interpreted differently depending on what is opening the file..

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
millerds: I don't think you want to print out the '*'.  That would make the message "Call context.sync() after calling load on 'obj' for property '*' and before reading the property." . . . but there is no property '*', that just happens to be what the load call argument it.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -37,6 +37,20 @@ export = {
...
+      propertyName: string
+    ): boolean {
+      if (
+        loadLocation.has(propertyName) && // If reference came after load, return
+        node.range[1] > (loadLocation.get(propertyName) ?? 0)
+      ) {
+        return true;
+      }
+      return false;
+    }
```
millerds, line 52: Can just return the result of the if statement directly . . . no need for the expanded if statement.

packages/eslint-plugin-office-addins/src/rules/load-object-before-read.ts
```diff
@@ -92,8 +106,8 @@ export = {
           }
 
           if (
```
millerds, line 108: Should include the '!propertyName' test in this same 'if' statement.  Can also move the next 'if' statement to right after setting the 'node' value to avoid the extra code processing when we can.

millerds (changes requested):

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: The message would be as you said and I think it is more correct than without it. For example, in the following code:
```
range.load("*");
console.log(range.address);
```
Would we rather the message to be: 
1. "Call context.sync() after calling load on range for property * and before reading the property."
2. "Call context.sync() after calling load on range for property address and before reading the property."

I prefer number 1, given that's what the load is. We can remove the "property" in the message to also give room to mention star, which is not a property.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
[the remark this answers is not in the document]
igor-ribeiiro: I just tried and it is valid.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
millerds: I think 2 is better because it identifies 'address' as the property that needs to be sunk where as * is ambiguous.

And actually after reading it now I have a few new thoughts:
- We could leave the name of the property out of the message so it just reads "Call context.sync() after calling load and before reading the property."
- The wording of the sentence in general it a little confusing . . . the phrase "calling load on range for property address" could be miss understood.
- We should probably get PM and or doc writers to get the right wording for our message . . . they are better at it then we are :-).

igor-ribeiiro: Sure, talking to someone is probably the right approach. I will email Ricky about this example I provided here to see what he suggests.

packages/eslint-plugin-office-addins/src/rules/no-navigational-load.ts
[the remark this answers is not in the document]
millerds: Good to know.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: @millerds I just fixed the message!
The PR is ready to be reviewed again!

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -68,12 +68,13 @@ export = {
         }
 
         if (operation === "Read" && variable) {
-          const propertyName: string = findPropertiesRead(
+          let propertyName: string = findPropertiesRead(
```
millerds, line 71: This can be put back to const.

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
```diff
@@ -13,7 +13,7 @@ export = {
     type: <"problem" | "suggestion" | "layout">"suggestion",
     messages: {
       callSyncAfterLoad:
-        "Call context.sync() after calling load on '{{name}}' for property '{{loadValue}}' and before reading the property.",
+        "Call context.sync() after calling load on '{{name}}' for '{{loadValue}}' and before reading the property.",
```
millerds, line 16: Ricky said use: 

"Call context.sync() after calling load on range for the property ‘address’ and before reading the property."

which translates to:

"Call context.sync() after calling load on '{{name}}' for the property '{{loadValue}}’ and before reading the property."

millerds (changes requested):

packages/eslint-plugin-office-addins/src/rules/call-sync-after-load.ts
igor-ribeiiro: Sure, thanks for seeing that.

millerds (approved):

PR #533, opened 2021-09-30
Update package version to match npm registry

millerds:
Looks like when we only needed to revert the version update for some of the packages and not all of them.  This updates the versions in the repo to match what is published in npm.  The next lerna publish should put things back in complete sync.


igor-ribeiiro (approved):

akrantz:
Before you publish again with Lerna, it might require a commit in each package. Use `lerna changed` to see what changes are pending before publish. Reach out directly if there are questions.

PR #538, opened 2021-10-19
Supporting load star to mocking

igor-ribeiiro:
Added support to load star in the mocking package.

Also fixed a bug where `loadScalar` was not recursively loading properties. It was just loading up to two layers of properties.

### DIFF
--- a/packages/office-addin-mock/src/officeMockObject.ts
+++ b/packages/office-addin-mock/src/officeMockObject.ts
@@ -83,12 +83,9 @@ export class OfficeMockObject {
   }
 
   private loadAllProperties() {
-    this.properties.forEach((_, propertyName: string) => {
-      this.properties.get(propertyName)?.loadCalled();
+    this.properties.forEach((property, propertyName: string) => {
+      property.loadCalled();
       this.makePropertyCallable(propertyName);
-      this.properties.forEach((property: OfficeMockObject) => {
-        property.loadCalled();
-      });
     });
   }
 


packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -82,6 +82,12 @@ export class OfficeMockObject {
     }
   }
 
+  private loadAllProperties() {
+    Array.from(this.properties.keys()).forEach((property: string) => {
```
millerds, line 86: Do we need the 'Array.from' call?  Doesn't the "keys()' call already return an array?

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -125,11 +135,11 @@ export class OfficeMockObject {
...
 
-      this.properties
-        .get(scalarPropertyName)
-        ?.properties.forEach((property: OfficeMockObject) => {
-          property.loadCalled();
-        });
+      Array.from(
+        this.properties.get(scalarPropertyName)?.properties.keys() ?? []
+      ).forEach((property: string) => {
+        this.properties.get(scalarPropertyName)?.loadScalar(property);
```
millerds, line 141: It's unclear to me why this change is needed?  It looks like you're replacing a 'loadCalled' method call with a 'loadScalar' call . . . which looks like infinite recursion.

igor-ribeiiro: It is not infinite because we only call the function recursively when it does have a value. We only call loadScalar on each of the keys, so when the object doesn't have any property, the recurse will end.

LoadScalar does call the `loadCalled` method and it also recurses that is the main difference. The recurse is necessary because of cases where someone might load a navigational property (it is a bad practice, but possible), which causes all properties to be loaded recursively.
I just added a test case to test this scenario.

packages/office-addin-mock/src/officeMockObject.ts
millerds: OK . . . had to remind myself of the objects at play here.  What about 'this.properties.foreach' as properties is a Map object that does have a foreach directly on it?  I'm not sure how much is matters or if one is way is more performant then others, I just thing it's easier to read.

igor-ribeiiro: I just managed to iterate over the map using `forEach((value, key))` now. It should be more clear now.

packages/office-addin-mock/src/officeMockObject.ts
igor-ribeiiro: I now did some clean-ups and I think the changes are more readable (just a few line changes). Should be easier to see what's different now. The difference is pretty much `loadCalled` vs recursing with `loadScalar`.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -82,6 +82,14 @@ export class OfficeMockObject {
     }
   }
 
+  private loadAllProperties() {
+    this.properties.forEach((_, propertyName: string) => {
+      this.properties.get(propertyName)?.loadCalled();
+      this.makePropertyCallable(propertyName);
+      this.properties.get(propertyName)?.loadAllProperties();
```
millerds, line 89: Doesn't this create the same recursion that load scalar was doing that you removed?  We only want to go one layer down and it seems like the loop itself accomplishes that by itself doesn't it?

igor-ribeiiro: For some reason I though `load("*");` would recursively load properties, but it doesn't (just tested again).
Reverted it so it only loads a single layers

millerds (approved):

PR #557, opened 2021-11-15
Renaming functions

igor-ribeiiro:
Merging `addMockFunction` and `setMock` into `setProperty` for more name clarity.

### DIFF
--- a/packages/office-addin-mock/src/officeMockObject.ts
+++ b/packages/office-addin-mock/src/officeMockObject.ts
@@ -18,6 +18,7 @@ export class OfficeMockObject {
    * Adds a function to OfficeMockObject. The function can be accessed by simply calling `this.methodName`
    * @param methodName Method name of the function to be added
    * @param methodBody Function to be added to the object. A blank function will be added if no argument is provided (Optional)
+   * @deprecated Use setProperty instead
    */
   addMockFunction(methodName: string, methodBody?: Function) {
     this[methodName] = methodBody ? methodBody : function () {};
@@ -58,6 +59,7 @@ export class OfficeMockObject {
    * Adds a property of any type to OfficeMockObject
    * @param propertyName Property name to the property to be added
    * @param value Value this added property will have
+   * @deprecated Use setProperty instead
    */
   setMock(propertyName: string, value: unknown) {
     if (!this.properties.has(propertyName)) {
@@ -67,6 +69,25 @@ export class OfficeMockObject {
     }
     this.properties.get(propertyName)?.resetValue(value);
     this[propertyName] = this.properties.get(propertyName)?.value;
+  }
+
+  /**
+   * Adds a property of any type to OfficeMockObject
+   * @param propertyName Property name to the property to be added
+   * @param value Value this added property will have
+   */
+  setProperty(propertyName: string, value: any) {
+    if (typeof value === "function") {
+      this[propertyName] = value;
+    } else {
+      if (!this.properties.has(propertyName)) {
+        const officeMockObject = new OfficeMockObject();
+        officeMockObject.isObject = false;
+        this.properties.set(propertyName, officeMockObject);
+      }
+      this.properties.get(propertyName)?.resetValue(value);
+      this[propertyName] = this.properties.get(propertyName)?.value;    
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -18,6 +18,7 @@ export class OfficeMockObject {
    * Adds a function to OfficeMockObject. The function can be accessed by simply calling `this.methodName`
    * @param methodName Method name of the function to be added
    * @param methodBody Function to be added to the object. A blank function will be added if no argument is provided (Optional)
+   * @deprecated Use setProperty instead
```
millerds, line 21: Use setProperty instead of addMockFunction?

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -69,6 +71,25 @@ export class OfficeMockObject {
     this[propertyName] = this.properties.get(propertyName)?.value;
   }
 
+  /**
+   * Adds a property of any type to OfficeMockObject
+   * @param propertyName Property name to the property to be added
+   * @param value Value this added property will have
+   */
+  setProperty(propertyName: string, value: any) {
```
millerds, line 79: Before we got and just change this we need to confirm with others that this is the right name and that it should work for both properties and functions.

We don't want to get into further trouble by changing the name so easily and frequently.

millerds (changes requested):

packages/office-addin-mock/src/officeMockObject.ts
igor-ribeiiro: Yes, the `setProperty` is simply now a check that if it is a function it behaves as the `addMockFunction` behaved and if it is not, it will behave as the `setMock` did.

packages/office-addin-mock/src/officeMockObject.ts
igor-ribeiiro: Sure. Do you have some names in mind for us to confirm?

millerds: I don't have specific names in mind, I just think the audience on approval of the new names and the timeline for releasing them needs to be more then just the two of us and while others are invited to look at these reviews, frequently they don't have the time :-).  You should get buy off in a broader audience email.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -25,7 +25,7 @@ export class OfficeMockObject {
   }
 
   /**
-   * Adds a mock object to OfficeMockObject. The object can be accessed by simply calling `this.objectName`
+   * addMock(name) will add a property named “name”, with a new OfficeMockObject as its value, to the object
    * @param objectName Object name of the object to be added
    */
   addMock(objectName: string) {
```
millerds, line 31: What happens if this is called twice with the same name?

igor-ribeiiro: The same as the new "set" function, it will simply override what is there.
Could we change to setMock as well? We would then have more consistent names

millerds: If we're overwriting the value then 'set' makes more sense to me . . . but are we then generating confusion between what the difference between 'set' and 'setMock' is?  For that matter why 'addMock' and 'set'.  I don't think is easily clear why developers would use one or the other.

I suppose I should be bringing this up in the email thread instead.

igor-ribeiiro: I suppose we can continue this PR now?

wandyezj: `setMock` is preferable per above explanation.

millerds: I believe the current thinking is to have 'set' which allows for setting the property or function of an Office object and 'addMock' which doesn't overwrite and adds a new mock office object.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -202,6 +207,25 @@ export class OfficeMockObject {
     this.loaded = false;
   }
 
+  /**
```
wandyezj, line 210: Delete set?

igor-ribeiiro: I like having this function internally, as it makes it easier to code in general, as every time you want to add something to the object, you can just call this function.
I could just change the name of this function to "setValue". Would that suffice?

igor-ribeiiro:
@millerds @wandyezj can I get your reviews again?

millerds (approved):

PR #559, opened 2021-11-18
Replace deprecated packages

millerds:
'npm install' was reporting that some @types packages were deprecated and un needed.  In order to compile a replacement packages was needed in most cases.

### DIFF
--- a/packages/custom-functions-metadata/package.json
+++ b/packages/custom-functions-metadata/package.json
@@ -33,7 +33,6 @@
   },
   "devDependencies": {
     "@types/assert": "^1.4.6",
-    "@types/es6-promise": "^3.3.0",
     "@types/mocha": "^7.0.2",
     "@types/node": "^14.17.2",
     "@types/node-fetch": "^2.5.10",
--- a/packages/custom-functions-metadata/package.json
+++ b/packages/custom-functions-metadata/package.json
@@ -38,6 +38,7 @@
     "@types/node-fetch": "^2.5.10",
     "@types/xregexp": "^3.0.30",
     "concurrently": "^6.2.2",
+    "es6-promise": "^4.2.8",
     "mocha": "^9.1.1",
     "office-addin-lint": "^1.4.3",
     "office-addin-prettier-config": "^1.1.1",
--- a/packages/custom-functions-metadata/package.json
+++ b/packages/custom-functions-metadata/package.json
@@ -38,7 +38,6 @@
     "@types/node-fetch": "^2.5.10",
     "@types/xregexp": "^3.0.30",
     "concurrently": "^6.2.2",
-    "es6-promise": "^4.2.8",
     "mocha": "^9.1.1",
     "office-addin-lint": "^1.4.3",
     "office-addin-prettier-config": "^1.1.1",
--- a/packages/office-addin-test-helpers/package.json
+++ b/packages/office-addin-test-helpers/package.json
@@ -23,8 +23,7 @@
     },
     "devDependencies": {
         "@types/es6-collections": "^0.5.31",
-        "@types/es6-promise": "0.0.32",
-        "@types/express": "^4.17.3",
+            "@types/express": "^4.17.3",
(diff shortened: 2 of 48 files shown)


packages/custom-functions-metadata/package.json
```diff
@@ -33,12 +33,12 @@
   },
   "devDependencies": {
     "@types/assert": "^1.4.6",
-    "@types/es6-promise": "^3.3.0",
     "@types/mocha": "^7.0.2",
     "@types/node": "^14.17.2",
     "@types/node-fetch": "^2.5.10",
     "@types/xregexp": "^3.0.30",
     "concurrently": "^6.2.2",
+    "es6-promise": "^4.2.8",
```
akrantz, line 41: We don't want to add es6-promise as we did work previously to remove it. We shouldn't need the @types either.

akrantz (changes requested):
Please don't add es6-promise back!

packages/custom-functions-metadata/package.json
akrantz: The promise polyfill should be coming from babel now.

packages/office-addin-test-helpers/package.json
```diff
@@ -23,8 +23,7 @@
     },
     "devDependencies": {
         "@types/es6-collections": "^0.5.31",
-        "@types/es6-promise": "0.0.32",
-        "@types/express": "^4.17.3",
+            "@types/express": "^4.17.3",
```
akrantz, line 26: nit: indentation.

packages/custom-functions-metadata/package.json
millerds: There's a compiler error when I don't have any reference to es6-promise (I'm guessing a compiler option needs to be changed, but I'm not sure of the side effects):

> eslint-plugin-office-addins@1.1.2 build
> rimraf lib && concurrently "tsc -p tsconfig.json"

[0] error TS2688: Cannot find type definition file for 'es6-promise'.
[0]   The file is in the program because:
[0]     Entry point of type library 'es6-promise' specified in compilerOptions
[0] tsc -p tsconfig.json exited with code 2

millerds: I tried removing 'es6-promise' from the tsconfig.json types array which allowed me to remove the package.  The build and tests work after this so I think it should be OK.

packages/office-addin-test-helpers/package.json
```diff
@@ -18,12 +18,10 @@
         "Office Add-in"
     ],
     "dependencies": {
-        "es6-promise": "^4.2.8",
-        "isomorphic-fetch": "^3.0.0"
+            "isomorphic-fetch": "^3.0.0"
```
igor-ribeiiro, line 21: Indentation here.

packages/office-addin-test-server/package.json
```diff
@@ -24,16 +24,14 @@
     "dependencies": {
         "commander": "^6.2.0",
         "cors": "^2.8.5",
-        "es6-promise": "^4.2.8",
-        "express": "^4.17.1",
+            "express": "^4.17.1",
```
igor-ribeiiro, line 27: Also here

packages/office-addin-test-server/package.json
```diff
@@ -24,16 +24,14 @@
...
         "office-addin-cli": "^1.3.2",
         "office-addin-dev-certs": "^1.7.3"
     },
     "devDependencies": {
         "@types/cors": "^2.8.6",
         "@types/es6-collections": "^0.5.31",
-        "@types/es6-promise": "0.0.32",
-        "@types/express": "^4.17.3",
+            "@types/express": "^4.17.3",
```
igor-ribeiiro, line 34: Also here indentation

igor-ribeiiro (changes requested):

akrantz (dismissed):
Looks good. There are the indentation fixes that Igor has noted.

igor-ribeiiro (approved):

Issue #500, opened 2021-08-10
Cannot sideload an Office document without an existing webextension xml part


akrantz:
# Expected behavior

The --document parameter can be used to specify an existing document to sideload. For sideload to web, this is the url of the document. For sideload to desktop, this is the file path to an existing document.

For sideload to desktop, if the specific document doesn't contain a webextension xml part, one should be added so it contains a reference to the add-in, thereby causing the add-in to be loaded when the document is opened.

NOTE: When sideload to web was implemented and the --document parameter added, we intentionally did work to support sideload to desktop for an existing document with a webextension part, but did not do the work to support it for arbitrary documents. I'm creating this issue to track this as some customers have asked and we haven't been clear about this.

# Current behavior

Sideload to desktop for an existing document without a webextension part will give an error.

## Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. "npm start manifest.xml --document <path to Office document without webextension part>"

igor-ribeiiro:
This feature request should be fixed when https://github.com/OfficeDev/Office-Addin-Scripts/pull/564 goes live.

PR #564, opened 2021-11-24
Sideload desktop existing file

igor-ribeiiro:
Making code accepts existing documents for sideloading on desktop.
Fixes in part: https://github.com/OfficeDev/Office-Addin-Scripts/issues/500 and https://github.com/OfficeDev/Office-Addin-Scripts/pull/564

This will make the CLI for `office-addin-debugging` work for documents created using our tooling and referenced later.
This won't fix the problem of making the tool work with any document, because editing them would cause the client to detect a corruption issue.

### DIFF
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -54,23 +54,28 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
     throw new ExpectedError("The manifest contains an unsupported OfficeApp xsi:type.");
   }
 
-  const templatePath = document && document !== "" ? path.resolve(document) : getTemplatePath(app, addInType);
+  const documentWasProvided = document && document !== "";
+  const templatePath = documentWasProvided ? path.resolve(document) : getTemplatePath(app, addInType);
 
   if (!templatePath) {
     throw new ExpectedError(`Sideload is not supported for apptype: ${addInType}.`);
   }
 
   const templateBuffer = await readFileAsync(templatePath);
-  const zip = await jszip.loadAsync(templateBuffer);
-  const webExtensionPath = getWebExtensionPath(app, addInType);
-
-  if (!webExtensionPath) {
-    throw new ExpectedError("Don't know the webextension path.");
-  }
+  const zip: jszip = await jszip.loadAsync(templateBuffer);
 
   const appName = getOfficeAppName(app);
   const extension = path.extname(templatePath);
-  const pathToWrite = makePathUnique(path.join(os.tmpdir(), `${appName} add-in ${manifest.id}${extension}`), true);
+  const pathToWrite: string = makePathUnique(path.join(os.tmpdir(), `${appName} add-in ${manifest.id}${extension}`), true);
+
+  if (documentWasProvided) {
+    return await writeZipFile(zip, pathToWrite);
+  }
+
+  const webExtensionPath = getWebExtensionPath(app, addInType);
+  if (!webExtensionPath) {
+    throw new ExpectedError("Don't know the webextension path.");
+  }
 
   // replace the placeholder id and version
   const zipFile = zip.file(webExtensionPath);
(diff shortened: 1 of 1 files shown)


packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -54,24 +54,32 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
...
+  const zip: jszip = await jszip.loadAsync(templateBuffer);
+
+  const appName = getOfficeAppName(app);
+  const extension = path.extname(templatePath);
+  const pathToWrite: string = makePathUnique(
+    path.join(os.tmpdir(), `${appName} add-in ${manifest.id}${extension}`),
+    true
+  );
 
+  if (documentWasProvided) {
```
millerds, line 74: Seems like a simpler change would be to just put an if(!documentWasProvided) around the section that manipulates the file instead of having to break things apart.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -82,16 +90,7 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
...
-  await new Promise((resolve, reject) => {
-    zip
-      .generateNodeStream({ type: "nodebuffer", streamFiles: true })
-      .pipe(fs.createWriteStream(pathToWrite))
-      .on("error", reject)
-      .on("finish", resolve);
-  });
-
-  return pathToWrite;
+  return await writeZipFile(zip, pathToWrite);
```
millerds, line 93: Does this still open the host application with the provided file?

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -54,24 +54,32 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
...
+  const pathToWrite: string = makePathUnique(
+    path.join(os.tmpdir(), `${appName} add-in ${manifest.id}${extension}`),
+    true
+  );
 
+  if (documentWasProvided) {
+    return await writeZipFile(zip, pathToWrite);
+  }
+
+  const webExtensionPath = getWebExtensionPath(app, addInType);
```
millerds, line 78: Does this work for launching in the web as well as the desktop client?

millerds (changes requested):

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: Yes, the provided file is opened

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: Yes, tested on web and desktop (with existing and non-existing files)

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -380,3 +377,16 @@ export async function sideloadAddIn(
...
+  // Write the file
+  await new Promise((resolve, reject) => {
+    zip
+      .generateNodeStream({ type: "nodebuffer", streamFiles: true })
+      .pipe(fs.createWriteStream(pathToWrite))
+      .on("error", reject)
+      .on("finish", resolve);
+  });
+
+  return pathToWrite;
```
millerds, line 391: Why do we need to return the same value that is passed in?  Seems like it's extra.  If we return anything I would think it would be a boolean indicating success of the file write . . . but we weren't doing that before so it would just be a bonus if we did it now.

millerds (changes requested):

packages/office-addin-dev-settings/src/sideload.ts
igor-ribeiiro: I just removed the function. It was not needed any longer after some of the previous refactorings

Mahmoud3193:
Hi every one?
I'm facing a problem that I think is related to this topic
I'm trying to debug an excel add-in and make it run on a local document 
The file opens without my add-in be side loaded
So can any one helps.

PR #565, opened 2021-11-29
Fix expired certs on mac

millerds:
The wrong dev certs are being retrieved on mac machines when there are expired certs and a new one added to the keychain.  This is an effect of the shell commands used for mac always picking the oldest one as well as not deleting all the expired ones.  This changes adds some script files (similar to the powershell cases for Win32) to use for those case to more accurate get the right cert and delete the old ones.


packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -31,13 +32,12 @@ export function isCaCertificateInstalled(): boolean {
...
-      return true;
-    } else if (output.length !== 0) {
-      return true; // powershell command return empty string if the certificate not-found/expired
+    // script files return empty string if the certificate not found or expired
+    if (output.length !== 0) {
+      return true; 
     }
   } catch (error) {
-    // Mac security command throws error if the certifcate is not-found/expired
+    // Some commands throw errora if the certifcate is not found or expired
```
igor-ribeiiro, line 40: NIT: throws*

packages/office-addin-dev-certs/scripts/uninstall.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+hashes=$(security find-certificate -c "$1" -a -Z | grep SHA-1 | awk '{ print $NF }')
+for hash in $hashes; do
+    security delete-certificate -Z $hash
+done
```
igor-ribeiiro, line 5: NIT: endline on the end of the file

packages/office-addin-dev-certs/scripts/verify.sh
```diff
@@ -0,0 +1,14 @@
...
+        cert=$line
+    else
+        cert="$cert"$'\n'"$line"
+        if [[ "$line" == *"--END"* ]]; then
+            if [ 0 -lt $(echo "$cert" | openssl x509 -checkend 86400 | grep -c "will not expire") ]; then
+                echo "$cert"
+            fi
+        fi
+    fi
+done <<< "$certs"
```
igor-ribeiiro, line 14: NIT: endline

packages/office-addin-dev-certs/src/verify.ts
millerds: Actually, "commands throw" is correct . . . but the "errora" should be "errors"

igor-ribeiiro (approved):

packages/office-addin-dev-certs/scripts/uninstall.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+hashes=$(security find-certificate -c "$1" -a -Z | grep SHA-1 | awk '{ print $NF }')
```
TCourtneyOwen, line 2: Will this change also work on Windows? I recall the reason we never completed the original change for this issue is that the solution I wrote up wouldn't work cross-platform?

Another open issue I recall is that the original solution would not remove ALL expired certs, but rather only the first one it encountered.  Will this remove all expired certs?

millerds: This is for bash script and uses the mac "security" command and I don't know if that works for other platforms or not (seems like it should for linux, but I don't have a test environment for that).  Right now windows uses powershell script files.

It does remove all the certificates (expired or otherwise) . . . that's why the for loop below.  Note I think the powershell for windows deletes them all as well.

The discussion I saw in you change (https://github.com/OfficeDev/Office-Addin-Scripts/pull/245/files) centered around the deletion being a side effect of checking if installed and the duplication of code.  I avoided those things by beefing up the uninstall and verify commands for mac to be more accurate instead.

PR #573, opened 2021-12-10
Export manifest package

millerds:
This adds a command to the Office-Addin-Manifest packages that allow the packaging of a manifest and the assets into a zip file.  This will be used in other improvements later on.


packages/office-addin-manifest/src/cli.ts
```diff
@@ -22,6 +22,13 @@ commander
 
 commander.command("validate <manifest-path>").action(commands.validate);
 
+commander
+  .command("export")
+  .option("-m, --manifest <manfest-path>", "Specify the location of the manifest file.  Default is './manifest.json'")
+  .option("-a, --assets <assets-folder-path>", "Specify the location of the assets folder.  Default is './assets'")
+  .option("-o, --output <output-path>", "Specify where to save the package.  Default is './package/manifest.zip'")
```
igor-ribeiiro, line 29: Can we put just `manifest.zip` as default?
I feel having the file in the same directory is more intuitive.

packages/office-addin-manifest/src/commands.ts
```diff
@@ -170,3 +171,17 @@ export async function validate(
     logErrorMessage(err);
   }
 }
+
+export async function exportManifest(command: commander.Command) {
+  try {
+    const outputPath: string | undefined = command.output ?? "./package/manifest.zip";
```
igor-ribeiiro, line 177: If we change the default path for the zip, we need to change here to

packages/office-addin-manifest/src/commands.ts
```diff
@@ -170,3 +171,17 @@ export async function validate(
...
+    const manifestPath: string | undefined = command.manifest ?? "./manifest.json";
+    const assetsPath: string | undefined = command.assets ?? "./assets";
```
igor-ribeiiro, line 179: Also, why do we need to put this `|undefined`? I think the type is always `string` right?

packages/office-addin-manifest/.vscode/launch.json
```diff
@@ -38,6 +38,21 @@
...
     },
+    {
+      "name": "Command Line: export",
+      "type": "node",
+      "request": "launch",
+      "program": "${workspaceFolder}/lib/cli.js",
+      "args": [
+        "export",
+        "-o",
+        "${workspaceFolder}\\test\\out.zip",
```
igor-ribeiiro, line 49: Not really in this line, but should we also add this `out.zip` to this PR?
This is more so git doesn't complain every time this command runs. We can also add this file to the `.gitignore`.

packages/office-addin-manifest/src/export.ts
```diff
@@ -0,0 +1,36 @@
+import * as fs from "fs";
+import * as fsExtra from "fs-extra";
+import * as jszip from "jszip";
+import * as path from "path";
+
+export function exportMetadataPackage(
+  output: string = "package/manifest.zip",
```
igor-ribeiiro, line 7: Also here

packages/office-addin-manifest/src/export.ts
```diff
@@ -0,0 +1,36 @@
...
+    zip.file("manifest.json", manifestData);
+  } else {
+    throw new Error(`The file '${manifest}' does not exist`);
+  }
+
+  if (fs.existsSync(assets)) {
+    const files: string[] = fs.readdirSync(assets);
+    zip.folder(assets);
+    files.forEach((element) => {
+      const filePath = assets + "/" + element;
```
igor-ribeiiro, line 24: Can we use a library to make this pathing work? There are probably some already made.
I remember for something like `path.join()`.
I say that because using a library would make this work for other Operation Systems as well.

packages/office-addin-manifest/src/export.ts
```diff
@@ -0,0 +1,36 @@
...
+      zip.file(filePath, fileData);
+    });
+  } else {
+    throw new Error("Need folder of assets referenced by manifest file");
+  }
+
+  fsExtra.ensureDirSync(path.dirname(output));
+  zip.generateNodeStream({ type: "nodebuffer", streamFiles: true }).pipe(fs.createWriteStream(output));
+
+  return output;
```
igor-ribeiiro, line 35: I don't think we need to return anything. 
Can we just make this function `void`?

packages/office-addin-manifest/test/test.ts
```diff
@@ -623,27 +624,56 @@ describe("Unit Tests", function() {
       });
     });
   });
+  describe("export.ts", function() {
+    describe("exportMetadataPackage()", function() {
+      it("export manifest", async function() {
+        this.timeout(6000);
+        const testFolder = path.resolve("./testExecution");
```
igor-ribeiiro, line 631: does this dir contain something? Sometimes it is hard to push dir to git.
If you need to push an empty dir, using `.gitkeep` can solve it.

packages/office-addin-manifest/test/test.ts
```diff
@@ -623,27 +624,56 @@ describe("Unit Tests", function() {
...
+        const manifestPath = path.normalize("test/manifests/manifest.json");
+        const assetsPath = path.normalize("test/assets");
+        const outputPath = path.normalize(`${testFolder}/testPackage.zip`);
+        const outputFile = await exportMetadataPackage(outputPath, manifestPath, assetsPath);
```
igor-ribeiiro, line 635: I see now why this function returns, but I think just testing the file exists is fine.

packages/office-addin-manifest/test/test.ts
```diff
@@ -623,27 +624,56 @@ describe("Unit Tests", function() {
...
+
+        assert.strictEqual(outputFile, outputPath, "Output path should match the argument");
+        assert.strictEqual(fs.existsSync(outputFile), true, "Output file should exist");
+        
+        // Cleanup
+        fs.rmSync(testFolder, { recursive: true });
```
igor-ribeiiro, line 641: Can we use the `deleteFolder` function?

packages/office-addin-manifest/test/test.ts
```diff
@@ -623,27 +624,56 @@ describe("Unit Tests", function() {
...
+      it("invalid manifest path", async function() {
+        this.timeout(6000);
+        let result: string = "";
+        const invalidManifestPath = path.normalize(`${manifestTestFolder}/foo/manifest.json`);
+        try {
+          await exportMetadataPackage(invalidManifestPath);
+        } catch (err: any) {
+          result = err.message;
+        }
+        assert.strictEqual(result.indexOf("The file 'manifest.json' does not exist") >= 0, true);
```
igor-ribeiiro, line 652: Can we just test if the error exists? 
Maybe we will change the message in the future and we don't want for the tests to fail.

igor-ribeiiro (changes requested):

packages/office-addin-manifest/src/cli.ts
```diff
@@ -22,6 +22,13 @@ commander
 
 commander.command("validate <manifest-path>").action(commands.validate);
 
+commander
+  .command("export")
+  .option("-m, --manifest <manfest-path>", "Specify the location of the manifest file.  Default is './manifest.json'")
```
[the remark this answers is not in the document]
millerds: The difference is that it becomes a required parameter without a default.  I noticed that on the other commands and wasn't sure why they didn't use a default instead of requiring it on the command line.  

If I was going to require and manual manifest argument what does that mean for the assets folder argument?

packages/office-addin-manifest/src/cli.ts
millerds: I wasn't sure where the best place for the output should be by default.  Some thoughts include

- root as you suggested
- temp folder
- folder in the project we could mark to be ignored (so it doesn't show up as a new file for the project).

Haven't found something I'm particularly drawn to.

packages/office-addin-manifest/src/commands.ts
millerds: Good point.  Seemed to be the pattern, but unnecessary here.

packages/office-addin-manifest/.vscode/launch.json
millerds: Why would we add the out file?  This is providing an argument for the "output" argument of the command so it can be debugged . . . meaning it's the file that gets generated for debugging.

packages/office-addin-manifest/src/export.ts
millerds: I didn't think the notation with '+' depended on the OS . . . but using the library makes better sense anyway.

packages/office-addin-manifest/src/export.ts
millerds: I started with a void, but thought returning the actual path used could be useful and I made use of it in the test case.

packages/office-addin-manifest/test/test.ts
millerds: Was trying that first, but for some reason using a method call creates some sort of ordering problem with the calls such that the delete ends up trying to delete the folder with the file still in it (and I didn't want to change the delete function to use this simpler call because I wasn't sure the side effect on the other usage since it's using specific operations within the function).

packages/office-addin-manifest/test/test.ts
millerds: If we return different errors we wouldn't be figuring out if the right one was the one received.

packages/office-addin-manifest/.vscode/launch.json
igor-ribeiiro: You are right, we don't need to add this file to git.
But we should add it to `.gitignore` right?

packages/office-addin-manifest/src/export.ts
igor-ribeiiro: I'm now completely sure I won't work on other OS, it is just sometimes the OS uses `//` or other ways to create paths, which is different than `/`.

packages/office-addin-manifest/src/cli.ts
igor-ribeiiro: Well, I'm drawn to being in the same directory the command was called, so `./manifest.zip`.
We can also put it to be ignored as well, even if we don't know the file:
Something like `*manifest.zip` should work.

millerds: Or maybe I put it in the same location the manifest file came from instead of specifying a specific default location.  This would be just ./manifest.zip for the default argument.

I'd like to use the .gitignore . . . however that only works for the default case.  If anything else is specified then it won't do any good.

packages/office-addin-manifest/.vscode/launch.json
millerds: Not sure.  I put it there to get it to work, but I'm not sure I like the idea of the left over artifact from debugging.  Maybe it belongs in the temp directory instead.

packages/office-addin-manifest/src/export.ts
millerds: But the default case is likely to be the most common one.  It might be interesting to return the whole path or a path object instead.

packages/office-addin-manifest/test/test.ts
millerds: This is just a directory created for testing.  You'll see the other test copies test file to a folder like this that it modifies in the process of the test.  It's just a simple way using temporary disk files that get cleaned up at the end of the test.

packages/office-addin-manifest/src/cli.ts
igor-ribeiiro: Yeah, `.gitignore` would only solve for the default case, but I think it is better than nothing.

packages/office-addin-manifest/.vscode/launch.json
igor-ribeiiro: Yeah, I agree we should not have leftovers pushed into git. Putting into temp folder is fine as well

packages/office-addin-manifest/src/export.ts
igor-ribeiiro: As I don't think this is a big deal, we can leave it returning the path like it is.

packages/office-addin-manifest/src/cli.ts
millerds: Actually, .gitignore would go in the template project not the script package here

igor-ribeiiro (approved):

PR #574, opened 2021-12-10
Manifest json

igor-ribeiiro:
Changing the code related to the XML, which parses and extracts elements from the XML file.
An abstract class was created and the idea is this will support other types of manifest very easily.

The code itself would only change the class name and all the rest of the calls could be the same.


millerds:
I think this needs a bit more design/explanation.  I get that you want to abstract away the parsing of different file types from the internal representation of the manifest . . . but the set of methods that were used as the basis of the abstract class (should this be an interface instead?) were designed for xml and even with 'xml' removed from their names I'm not sure the set of methods also applies to json manifests which don't have attributes (and I'm not sure 'elements' applies very well either).  

Probably worth discussion a bit before implementation.

igor-ribeiiro:
Closing as we will be tracking only the branch now

PR #578, opened 2022-01-04
Messaging updates

millerds:
Tweaking a couple of customer facing messages


igor-ribeiiro (approved):

igor-ribeiiro (dismissed):
Approved. But why creating a new error is needed?

millerds:
Assuming you are talking about the "object.create" it's because it ends up being a by reference case other wise and when the message and stack are modified to remove sensitive information for telemetry it also modifies the output of the exception.

The other cases are clarifying the messages being logged with more useful information why trying to diagnose errors (one was pointed out by a customer).

BTW . . . can you re-approve after my last update?

igor-ribeiiro (approved):

PR #581, opened 2022-01-12
Update validation url

millerds:
The OMEX teams has asked that we change from a gate to a url param for indicating our manifest validation comes from a dev environment.  I have also added a command argument that allows the validation to be done for a production environment.

### DIFF
--- a/packages/office-addin-manifest/test/test.ts
+++ b/packages/office-addin-manifest/test/test.ts
@@ -552,6 +552,16 @@ describe("Unit Tests", function() {
         assert.strictEqual(validation.report!.warnings!.length, 0);
         assert.strictEqual(validation.report!.addInDetails!.supportedProducts!.length > 0, true);
       });
+      it("valid manifest prod (fail)", async function() {
+        this.timeout(6000);
+        const validation = await validateManifest("test/manifests/TaskPane.manifest.xml", true);
+        assert.strictEqual(validation.isValid, false);
+        assert.strictEqual(validation.status, 200);
+        assert.strictEqual(validation.report!.errors!.length, 2);
+        assert.strictEqual(validation.report!.notes!.length > 0, true);
+        assert.strictEqual(validation.report!.warnings!.length, 0);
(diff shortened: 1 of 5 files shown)


millerds:
Note I was considering change the output or to have the status and errors at the end (after the long info section) rather than the beginning as that is what is first seen . . . but though I should get some feedback on that first.

akrantz (dismissed):
This looks OK to me. I don't recall what the output looks like, so I cannot comment on whether it would be OK to change the output. Feel free to email me a sample.

packages/office-addin-manifest/test/test.ts
```diff
@@ -552,6 +552,16 @@ describe("Unit Tests", function() {
         assert.strictEqual(validation.report!.warnings!.length, 0);
         assert.strictEqual(validation.report!.addInDetails!.supportedProducts!.length > 0, true);
       });
+      it("valid manifest prod (fail)", async function() {
+        this.timeout(6000);
+        const validation = await validateManifest("test/manifests/TaskPane.manifest.xml", true);
+        assert.strictEqual(validation.isValid, false);
+        assert.strictEqual(validation.status, 200);
+        assert.strictEqual(validation.report!.errors!.length, 2);
```
igor-ribeiiro, line 560: I don't think it is that nice for the tests to be so specific like that. The error length might change in the future.
Could we change this to: `assert.strictEqual(validation.report!.errors!.length > 0, true);`

igor-ribeiiro (changes requested):
Just some small updates to the test

packages/office-addin-manifest/test/test.ts
millerds: We do want to be specific and check for the number of errors errors (it'ls the length of an array being checked) because we need to know if the validation changes and our template manifest needs to be updated (from what I hear we've been broken before).

igor-ribeiiro: Makes sense that we are testing if the validation changes

igor-ribeiiro (approved):

PR #583, opened 2022-01-14
Removing sideffects from maskFilePaths telemetry function

igor-ribeiiro:
The function `maskFilePaths ` was having a side effect of modifying the value passed to it.
This would lead to unwanted behavior in the function `reportExpectedException` for example, in which calling it would have side effects on the parameter.

### DIFF
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -284,10 +284,10 @@ export class OfficeAddinUsageData {
    */
   public maskFilePaths(err: Error): Error {
     try {
-      const regexRemoveUserFilePaths = /[/\\](.*[/\\]+)*(.+\.)+/gim;
+      const regexRemoveUserFilePaths = /(\w:)*[/\\](.*[/\\]+)*(.+\.)+[a-zA-Z]+/gim;
 
-      err.message = err.message.replace(regexRemoveUserFilePaths, "\\.");
-      err.stack = err.stack.replace(regexRemoveUserFilePaths, "\\.");
+      err.message = err.message.replace(regexRemoveUserFilePaths, '""');
+      err.stack = err.stack.replace(regexRemoveUserFilePaths, '""');
 
       return err;
     } catch (err) {
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -352,7 +352,7 @@ export class OfficeAddinUsageData {
   public reportExpectedException(method: string, err: Error | string, data: object = {}) {
     let error = err instanceof Error ? Object.create(err) : new Error(`${this.options.projectName} error: ${err}`);
     error.name = this.getEventName();
-    error = this.maskFilePaths(error);
+    this.maskFilePaths(error);
     const errorMessage = error instanceof Error ? error.message : error;
 
     this.sendUsageDataEvent({
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -282,13 +282,14 @@ export class OfficeAddinUsageData {
    */
   public maskFilePaths(err: Error): Error {
     try {
+      const error: Error = err;
       const regexRemoveUserFilePaths = /[\/\\](.*)[\/\\]/gim; /* eslint-disable-line no-useless-escape */
       const regexRemoveAbsoluteUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gim;
       const regexRemoveFirstFilePathFromStack = /\\(.*)\./i;
-      err.message = err.message.replace(regexRemoveUserFilePaths, "\\");
-      err.stack = err.stack.replace(regexRemoveFirstFilePathFromStack, "\\.");
-      err.stack = err.stack.replace(regexRemoveAbsoluteUserFilePathsFromStack, "");
-      return err;
+      error.message = error.message.replace(regexRemoveUserFilePaths, "\\");
+      error.stack = error.stack.replace(regexRemoveFirstFilePathFromStack, "\\.");
+      error.stack = error.stack.replace(regexRemoveAbsoluteUserFilePathsFromStack, "");
+      return error;
     } catch (err) {
       this.reportError("maskFilePaths", err);
       throw new Error(err);
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -184,7 +184,7 @@ export class OfficeAddinUsageData {
     if (this.getUsageDataLevel() === UsageDataLevel.on) {
       let error = Object.create(err);
 
-      err.name = this.options.isForTesting ? `${errorName}-test` : errorName;
+      error.name = this.options.isForTesting ? `${errorName}-test` : errorName;
       this.usageDataClient.trackException({
         exception: this.maskFilePaths(error),
       });
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -285,9 +285,10 @@ export class OfficeAddinUsageData {
   public maskFilePaths(err: Error): Error {
     try {
       const regexRemoveUserFilePaths = /(\w:)*[/\\](.*[/\\]+)*(.+\.)+[a-zA-Z]+/gim;
+      const maskToken: string = "<filepath>";
 
-      err.message = err.message.replace(regexRemoveUserFilePaths, '""');
-      err.stack = err.stack.replace(regexRemoveUserFilePaths, '""');
+      err.message = err.message.replace(regexRemoveUserFilePaths, maskToken);
+      err.stack = err.stack.replace(regexRemoveUserFilePaths, maskToken);
 
       return err;
     } catch (err) {
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -352,7 +352,7 @@ export class OfficeAddinUsageData {
   public reportExpectedException(method: string, err: Error | string, data: object = {}) {
     let error = err instanceof Error ? err : new Error(`${this.options.projectName} error: ${err}`);
     error.name = this.getEventName();
-    this.maskFilePaths(error);
+    error = this.maskFilePaths(error);
     const errorMessage = error instanceof Error ? error.message : error;
 
     this.sendUsageDataEvent({
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -284,10 +284,10 @@ export class OfficeAddinUsageData {
    */
   public maskFilePaths(err: Error): Error {
     try {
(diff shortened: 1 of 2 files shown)


packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -282,13 +282,14 @@ export class OfficeAddinUsageData {
    */
   public maskFilePaths(err: Error): Error {
     try {
+      const error: Error = err;
```
millerds, line 285: This doesn't work.  Note the check-in I made recently to reportException to clone the object . . . that's because statement like this end up just be a new reference to the same data.

millerds (changes requested):
Do we really want this?  The method says it's going to change the path and I think that's OK.  Methods like "reportException" and "reportExpectedException" don't suggest they are changing anything.  I guess in the end it boils down to where the responsibility lies in duplicating error . . . in the methods reporting the error (which are the cases that call the masking because that is where the data is exposed) or in the utility method for masking (which could be used anywhere).

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: We definitely don't want the side effects. I was having a problem with it when reporting an exception and testing the error value in the manifest (because reporting the exception changed the error).

You are right that simply assigning the error to a new variable doesn't solve the problem.
This lead me to that the tests here are broken in which the side effects are making one test depend on another. Working on it.

igor-ribeiiro: @millerds you are right that didn't fix it. Those side effects were making the tests pass wrongfully (one test was depending on others). Did some work to fix these errors as well.
I also moved the side effect handling to the report error functions as I also think it makes sense for the mask function to change the error.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -282,12 +284,11 @@ export class OfficeAddinUsageData {
...
-      const regexRemoveAbsoluteUserFilePathsFromStack = /\w:\\(?:[^\\\s]+\\)+/gim;
-      const regexRemoveFirstFilePathFromStack = /\\(.*)\./i;
-      err.message = err.message.replace(regexRemoveUserFilePaths, "\\");
-      err.stack = err.stack.replace(regexRemoveFirstFilePathFromStack, "\\.");
-      err.stack = err.stack.replace(regexRemoveAbsoluteUserFilePathsFromStack, "");
+      const regexRemoveUserFilePaths = /\w:[\/\\](?:[^\\\s]+[\/\\])+/gmi;
+
+      err.message = err.message.replace(regexRemoveUserFilePaths, "");
+      err.stack = err.stack.replace(regexRemoveUserFilePaths, "");
+
```
millerds, line 291: Why all these changes?  We shouldn't change anything here.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -349,9 +350,9 @@ export class OfficeAddinUsageData {
    * @param data Data object(s) sent to Application Insights
    */
   public reportExpectedException(method: string, err: Error | string, data: object = {}) {
-    let error = err instanceof Error ? err : new Error(`${this.options.projectName} error: ${err}`);
+    let error = err instanceof Error ? Object.create(err) : new Error(`${this.options.projectName} error: ${err}`);
     error.name = this.getEventName();
-    this.maskFilePaths(error);
+    error = this.maskFilePaths(error);
     const errorMessage = error instanceof Error ? error.message : error;
```
millerds, line 356: If you've cloned it and we aren't cloning it inside maskFilePaths then you don't need to reassign it.

packages/office-addin-usage-data/test/test.ts
```diff
@@ -162,23 +161,29 @@ describe("Test office-addin-usage data-package", function() {
...
+      compareErrorWithBackslash.message = "this error contains a file path:excel file .xlsx";
+      compareErrorWithBackslash.stack = "this error contains a file path:excel file .xlsx";;
+      const errWithBackslash = new Error(`this error contains a file path:C:\\Users\\admin\\AppData\\Local\Temp\\excel file .xlsx`);
+      
       addInUsageData.maskFilePaths(errWithBackslash);
-      assert.equal(compareErrorWithBackslash.message, errWithBackslash.message);
-      assert.equal(errWithBackslash.stack.includes(compareErrorWithBackslash.stack), true);
+
+      assert.strictEqual(compareErrorWithBackslash.message, errWithBackslash.message);
+      assert.strictEqual(errWithBackslash.stack.includes(compareErrorWithBackslash.stack), true);
```
millerds, line 186: Why all these test changes?  The clone change shouldn't have required test changes.

millerds (changes requested):

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: The code was wrong. It simply didn't mask the files sometimes.
The tests were passing because they were relying on the order they were executed.

When I made the mask function not have side effects for example, the tests started not passing.

packages/office-addin-usage-data/test/test.ts
igor-ribeiiro: Yeah, it shouldn't, but the `mask` function was bugged and wasn't actually masking the files consistently.
The syntax of having:
`regexRemoveAbsoluteUserFilePathsFromStack` and  `regexRemoveFirstFilePathFromStack` were probably a way to make the tests pass but were not really correct.

packages/office-addin-usage-data/src/usageData.ts
millerds: But now we're leaving it to have the "side effects" so this doesn't need to change (the less you change the less chance you have of unintentionally breaking something).  Aside from that I can't tell if the new expression covers all the cases the 3 previous expressions did without doing too much.

packages/office-addin-usage-data/test/test.ts
millerds: Can you explain why you think it was made that way to make the tests pass and why it wasn't correct and why this new way is correct?

millerds (changes requested):

packages/office-addin-usage-data/test/test.ts
igor-ribeiiro: I didn't dive deep to understand why it wasn't correct, it simply didn't get the regex and parse user files. 

The process I made to figure out that there was a problem with the function: I removed side effects on the functions that call it and on the tests -> this made the tests start to fail. So my guess is that the function was only working when the tests suites were the way they were. 

Also, I think the way the code looks now is more "right". We want to find user paths on the code and remove them. Why can't we just have one regex for that? My guess is that the code before was overly complicated to account for the side effects in the tests.

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: My bad for not communicating this. I had a plan for these functions to not have side effects, but then you talked about it being fine (which I agree with for the `maskFilePaths`). So what I did was to remove the side effects from the `sendUsageDataException` and the `reportExpectedException`  which had them (and called `maskFilePaths`), and I also removed the side effects from the `test.ts`, which had multiple tests using the same variable (the `err` global variable).   

So side effects from the public functions and tests were removed, but the side effect from `maskFilePaths` was kept.  

About the fact that you cannot tell if the expression covers all the cases that were previously covered. The tests are still passing, the same tests we had.

akrantz: Please ensure this meets compliance requirements.

igor-ribeiiro: @akrantz It should fix the error from before (which could not parse some paths) and that's it. No other behaviors are expected.

igor-ribeiiro:
Added some more tests and checked manually some tests errors to ensure we were not keeping user paths.
Right now the regex is getting all the path information up to `.`, which means it will only keep the file extension.

packages/office-addin-usage-data/test/test.ts
```diff
@@ -11,8 +11,7 @@ import * as officeAddinUsageData from "../src/usageData";
 import * as jsonData from "../src/usageDataSettings";
 
 let addInUsageData: officeAddinUsageData.OfficeAddinUsageData;
-const err = new Error(`this error contains a file path:C:/${os.homedir()}/AppData/Roaming/npm/node_modules//alanced-match/index.js`);
-const errWithBackslash = new Error(`this error contains a file path:C:\\Users\\admin\\AppData\\Local\Temp\\excel file .xlsx`);
+const err = new Error(`this error contains a file path:C:/Users/admin/AppData/Roaming/npm/node_modules/alanced-match/index.js`);
```
millerds, line 14: I wonder if "os.homedir()" was used intentionally to look for things relating to MacOS and Unix?

packages/office-addin-usage-data/src/usageData.ts
millerds: I'm not certain we should be removing all file names.  You made that change before to address the first file in the stack specifically and I'm thinking that's because the rest of the stack mostly includes information from libraries that would be useful in figuring out what is going on.  If we're going to change these regexs at all we should be very targeted to what's needed for compliance.  We want to remove information considered user data, but not so much information that the remaining data becomes useless (what's the point in keeping the drive letter and the extension).

millerds (changes requested):

packages/office-addin-usage-data/test/test.ts
igor-ribeiiro: I may have been. I will simply add it back since there is no harm in doing so

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: Maintaining the drive letter and extensions is very useless I agree. It just made it easier to write the regexes, as it is very hard for these two to contain users' information.

We can't just remove the first file from the stack, because other files from it might contain user's info.
I think there's still point in keeping the message and logging it, even without the files on it.

If we know the message, we can later do a search on the code and find where that message is. We them can see what the error is about if a lot of users are having the problem, so it is not useless.

millerds: The message is certainly useful and I'm not questioning having that.  

I was thinking that call stacks are useful and are limited when we don't know the files from the libraries we provide that are part of the stack.  One could argue that the file names are going to be public when the package is created anyway (which is the reason we have file names from libraries that are not theirs and not ours) since it's all ts/js.

However, I suppose to be on the safe side we shouldn't have the file paths at all (even the drive letter and extension).  If we're going to change this now we should change all file paths to just be a token (like "<FilePath>") indicating it was a file path in the message or the stack.

igor-ribeiiro: Yeah, it is more to be on the safe side. I just added the `""` string for us to know this was filtered from the original message.
I also made the regex remove the drive letter and extension as well

millerds: The blank "" doesn't tell us what was removed.  It could be that we mask other things per requirements and then we can't tell which is which . . . thus the suggestion of some sort of token.

millerds (approved):

PR #584, opened 2022-01-14
Adding scaffold code to support .json manifests

igor-ribeiiro:
Making `office-addin-manifest` be able to distinguish `.json` and `.xml` manifests and treat each of them with different functions.


packages/office-addin-manifest/src/manifestInfo.ts
```diff
@@ -66,110 +65,24 @@ export namespace OfficeAddinManifest {
...
-
-      manifest.defaultSettings = defaultSettings;
-    }
-  }
-
-  return manifest;
-}
-
-async function modifyManifestXml(manifestPath: string, guid?: string, displayName?: string): Promise<Xml> {
+async function readFromManifestFile(manifestPath: string): Promise<string> {
```
millerds, line 79: Seems like this should be part of the handlers.  It looks like previously this was done as part of the readXmlFromManifestFile and now you've split reading from parsing, but is there any reason you would read w/out parsing or need to parse something that was already read?  You are right in that we don't want to read a file multiple times, but I'm not sure it's the case that we would unless explicitly called to do (or at least we could make it so).

packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { ManifestHandler } from "./manifestHandler";
+import { ManifestHandlerJson } from "./manifestHandlerJson";
+import { ManifestHandlerXml } from "./manifestHandlerXml";
+
+function isJsonFile(path: string) {
+  return path.endsWith(".json");
+}
+
+export function getManifestHandler(manifestPath: string): ManifestHandler {
```
millerds, line 12: Maybe this should be part of ManifestInfo?  Is it used anywhere else besides that?

packages/office-addin-manifest/test/test.ts
```diff
@@ -472,7 +477,37 @@ describe("Unit Tests", function() {
...
+      it("should read the manifest json info", async function() {
+        let result;
+        try {
+          await OfficeAddinManifest.readManifestFile("test/manifests/manifest.json");
+        } catch (err: any) {
+          result = err.message;
+        }
+        assert.strictEqual(result, "Manifest cannot be parsed in .json files");
+      });
+      it ("should an invalid json format", async function() {
```
millerds, line 490: Not sure what the description means

packages/office-addin-manifest/test/test.ts
```diff
@@ -472,7 +477,37 @@ describe("Unit Tests", function() {
...
+        let result;
+        try {
+          await OfficeAddinManifest.readManifestFile("test/manifests/invalid/invalid-manifest.json");
+        } catch (err: any) {
+          result = err.message;
+        }
+        assert.strictEqual(result, "Unable to read data for manifest file: test/manifests/invalid/invalid-manifest.json. \nSyntaxError: Unexpected token ] in JSON at position 4114");
+      });
+      it ("inexistent manifest", async function() {
```
millerds, line 499: Should be "nonexistent"

packages/office-addin-manifest/test/test.ts
```diff
@@ -678,6 +743,40 @@ describe("Unit Tests", function() {
...
+        const manifestHandler: ManifestHandler = getManifestHandler("test/manifests/TaskPane.Excel.manifest.xml");
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerXml, true);
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerJson, false);
+      });
+      it("Detects an inexistent JSON manifest", async function() {
+        const manifestHandler: ManifestHandler = getManifestHandler("test/foo/manifest.json");
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerJson, true);
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerXml, false);
+      });
+      it("Detects an inexistent XML manifest", async function() {
```
millerds, line 763: should be "a nonexistent"

packages/office-addin-manifest/test/test.ts
```diff
@@ -678,6 +743,40 @@ describe("Unit Tests", function() {
...
+        const manifestHandler: ManifestHandler = getManifestHandler("test/foo/TaskPane.Excel.manifest.xml");
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerXml, true);
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerJson, false);
+      });
+      it("Detects an invalid JSON manifest", async function() {
+        const manifestHandler: ManifestHandler = getManifestHandler("test/manifests/invalid-manifest.json");
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerJson, true);
+        assert.strictEqual(manifestHandler instanceof ManifestHandlerXml, false);
+      });
+      it("Detects an inexistent XML manifest", async function() {
```
millerds, line 773: Should be "a nonexistant"

millerds (changes requested):

packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (isJsonFile(manifestPath)) {
```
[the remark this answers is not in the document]
igor-ribeiiro: I don't want to add the XML check now because I don't want to change the current behavior of the code.
Right now, any file goes to the XML handler and I want to keep it that way to avoid messing too much with production code for now.

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { ManifestInfo } from "../manifestInfo";
```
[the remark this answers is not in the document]
igor-ribeiiro: Yeah, I will fix this apparently circular problem by just creating another file.

packages/office-addin-manifest/src/manifestInfo.ts
igor-ribeiiro: This separation was made because now we need to check whether a file is JSON or not before assigning a handler.
To verify that, we need to read the file. That's why this `readFromManifestFile` was extracted from the handler: to be used to check the file extension.
I think it would be nice for the file to only be read once, which is currently the case with this implementation.

packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts
igor-ribeiiro: The main reason is to abstract the `ManifestInfo` file from importing the specific manifest types.
By having this in a different file, the `ManifestInfo` doesn't have to know which handler it is dealing with (not importing them)

packages/office-addin-manifest/test/test.ts
igor-ribeiiro: Typo

packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts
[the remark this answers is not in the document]
millerds: So what happens currently when the wrong file is sent?  How do you plan on making sure this is done later if not now?

packages/office-addin-manifest/src/manifestInfo.ts
millerds: You don't need to read the file to check the extension . . . you are already doing that and picking the handler before reading.

Was the code before your change reading it more than once?  How does this prevent it from being read more than once?

igor-ribeiiro: Nah, you are right. I was thinking of a previous implementation I was making.
I added the file reading part to the `ManifestHandler` class, so the code is cleaner.

packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts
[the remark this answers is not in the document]
igor-ribeiiro: If a wrong file is sent, the same thing that happened previously will continue happening:
The following error will be thrown:
```reject(new Error(`Unable to parse the manifest file: ${manifestPath}. \n${parseError}`));```
I don't think we need to care about it too much. The `ManifestHandlerXML` already handles all kind of files (by throwing an error).

**Comment by igor-ribeiiro** [comment]
Code is ready to be reviewed again! @millerds

File: packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts (hunk 7 above)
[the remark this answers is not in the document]
Comment (akrantz) [reply packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts]: It's always good to think about having good, informative, and clear errors. Nothing is more frustrating that the user having to guess what is wrong. If you can tell the difference between a parse error or another condition, it's good to have a more specific message. If the parse error can be clear about the problem, and show that a non-XML file cannot be parsed propertly, then that can be sufficient.

File: packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts [hunk 9]
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { ManifestHandler } from "./manifestHandler";
+import { ManifestHandlerJson } from "./manifestHandlerJson";
+import { ManifestHandlerXml } from "./manifestHandlerXml";
+
+export async function getManifestHandler(manifestPath: string): Promise<ManifestHandler> {
```
millerds, line 8: Couldn't this be part of manifestOperations (the only places it's used . . . beside testing, which only really needs to happen because it's being exported which it wouldn't need to be).

packages/office-addin-manifest/src/manifestHandler/getManifestHandler.ts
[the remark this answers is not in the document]
igor-ribeiiro: I agree @akrantz. I just added a more clarifying message in case an unsupported extension is given to this method.

millerds (changes requested):

packages/office-addin-manifest/src/manifestOperations.ts
```diff
@@ -0,0 +1,63 @@
...
+      const manifestHandler: ManifestHandler = await getManifestHandler(manifestPath);
+      const manifest: ManifestInfo = await manifestHandler.parseManifest();
+      return manifest;
+    } else {
+      throw new Error(`Please provide the path to the manifest file.`);
+    }
+  }
+}
+
+export async function getManifestHandler(manifestPath: string): Promise<ManifestHandler> {
```
millerds, line 49: I think this doesn't need to be exported.  It's only used inside this class in the product and the testing can verify by way of the exported functions rather then directly.

millerds (approved):

PR #585, opened 2022-01-18
Improve typescript linting

millerds:
There is a library of typescript specific rules and while we included the library as a plugin to our linter cli, but didn't enable the rules in cli config.  This change adds a rule to the cli config, enabled the cli to use a project linter config file if it's there (allowing customer to override or extend the set of rules used), and then updates the other packages to follow the linting rules.


packages/eslint-plugin-excel-custom-functions/src/rules/utils.ts
```diff
@@ -8,9 +8,9 @@ import {
 import * as ts from "typescript";
 
 export enum OfficeCalls {
-  WRITE = "WRITE", // eslint-disable-line no-unused-vars
-  READ = "READ", // eslint-disable-line no-unused-vars
-  NOTOFFICE = "NOTOFFICE", // eslint-disable-line no-unused-vars
+  WRITE = "WRITE",
```
akrantz, line 11: I think there was a bug in typescript-eslint package where no-unused-vars error message were reported, but I think it is now fixed. I am wondering if it is really desirable to have no-unused-vars set to off in the default config.

millerds: There is the base eslint no-unused-vars and the @typescrpt-eslint/no-unused-vars.  The typescript version handles things like enums that the base rule does not.  We had rule updates in eslintrc files that disabled the base version in favor of the typescript version, but it wasn't in the linter cli.  This PR adds that to the exported configs so these disable lines are no longer needed since the typescript version covers it (verified when making the change).

packages/office-addin-lint/src/lint.ts
```diff
@@ -24,8 +25,10 @@ function normalizeFilePath(filePath: string): string {
 }
 
 function getEsLintBaseCommand(useTestConfig: boolean = false): string {
-  const configFilePath = useTestConfig ? eslintTestConfigPath : eslintConfigPath;
-  const eslintBaseCommand: string = `node "${eslintFilePath}" -c "${configFilePath}" --resolve-plugins-relative-to "${__dirname}"`;
+  const projLintConfig = path.resolve(process.cwd(), ".eslintrc.json");
+  const libLintConfig = useTestConfig ? eslintTestConfigPath : eslintConfigPath;
+  const useConfig = fs.existsSync(projLintConfig) ? "" : `-c ${libLintConfig} `;
```
igor-ribeiiro, line 30: If `useTestConfig` is true, don't we want to use the `eslintTestConfigPath`? It is weird to pass a parameter `useTestConfig` to the function that may not be used at all depending on if a file exists.
Maybe this file checking can be done in another place? Or this boolean handling somewhere else?

igor-ribeiiro (changes requested):

igor-ribeiiro (approved):

PR #586, opened 2022-01-21
[Minor] Moving logError from cli to usage-data

igor-ribeiiro:
Moving `logErrorMessage` from "office-addin-cli" to "office-addin-usage-data".


igor-ribeiiro:
Added a commit to also fix packages vulnerabilities.
Ran `npm run audit:fix`

millerds:
> Added a commit to also fix packages vulnerabilities. Ran `npm run audit:fix`

Should do the fix in a separate PR (I'll work on that) and publish a patch version of the packages.  This kind of change (moving calls from different packages) will require a major version updated (breaking change) and we don't want to hold up vulnerability fixes for that (i.e. major version changes in customer projects are more involved than minor version changes).

millerds:
> > Added a commit to also fix packages vulnerabilities. Ran `npm run audit:fix`
> 
> Should do the fix in a separate PR (I'll work on that) and publish a patch version of the packages. This kind of change (moving calls from different packages) will require a major version updated (breaking change) and we don't want to hold up vulnerability fixes for that (i.e. major version changes in customer projects are more involved than minor version changes).

Actually . . . it looks like we need a major version update for some other changes (linting, mock, etc.) anyway.  But the vulnerabilities should still be fixed in a different PR to keep this one simple and to the point.

igor-ribeiiro:
Just reverted the upgrade commit. PR is ready to be reviewed now

packages/custom-functions-metadata/src/cli.ts
```diff
@@ -4,7 +4,7 @@
 // Licensed under the MIT license.
 
 import * as commander from "commander";
-import { logErrorMessage } from "office-addin-cli";
+import { logErrorMessage } from "office-addin-usage-data";
```
akrantz, line 7: Why are you doing this? It shouldn't be part of office-addin-usage-data, as it has nothing to do with telemetry. It's also a breaking change for all of the package.

akrantz (changes requested):
I'd like to understand the motivation for this PR.

igor-ribeiiro:
> I'd like to understand the motivation for this PR.

We would like to be adding new commands to `office-addin-cli`. It would work as a general package that could house some new generic commands.
To do this, it is necessary for `office-addin-cli` to import `office-addin-usage-data` in order to have telemetry.

The problem is that `office-addin-usage-data` imports `office-addin-cli` because of the method `logError`, which is a very generic method. This causes a circular dependency that Lerna doesn't like.

This PR aims to move this `logError` function from `office-addin-cli` to `office-addin-usage-data`, in order to also avoid future circular dependency problems.

packages/custom-functions-metadata/src/cli.ts
igor-ribeiiro: Agree that `usage-data` is probably not the best name for the package, but I think it makes more sense to have the `logError` command than `office-addin-cli`.
We were thinking of changing its name to `logging` or something like that later on.

Yeah, it would be a breaking change from `office-addin-cli`, but why would it break other packages?

millerds: Dependencies . . . packages that depend on cli to get logError will need to depend on usage-data instead (which our probably do anyway . . . but not necessarily customers).  That is what this will be a major version update for the packages.

igor-ribeiiro: But this package depends on `cli`, we can just change the dependencies of it internally. There would be no external consequences that clients would see, we are just changing some dependencies of the package.

It would need a version bump, but I don't get why it would be major if there wouldn't be any breaking changes to external clients.

millerds: I'm not sure what you mean by "this package depends on cli".

In general any current project can have a dependency on cli or usage-data or both and we don't have control over that.  I'm not sure how introducing a dependency in cli on usage-data will prevent breaks.  Even if the package is installed because we depend on it, it doesn't mean the reference in code can get logError from usage-data by way of cli.

igor-ribeiiro: Yeah, I agree that `office-addin-cli` would have breaking changes as `logError` is not there anymore.

What I don't get is how another package, such as `office-addin-debugging` which used to depend on `office-addin-cli` would have breaking changes. We pretty much only changed the dependencies, with no external signatures changing.

millerds: Just like in this PR . . . you have to change package.json to get the usage-data package if it doesn't already get it and you have to change the import statement in commands.ts (and other similar files) which currently says 'import { logErrorMessage, parseNumber, getPackageJsonScript } from "office-addin-cli";' but would need to use the updated package structure.  If you simply reved the cli version the build would break.  

This PR does that for our packages in this repo, but that doesn't automatically do if for other repos and or project.  If you didn't update the major version of the debugging package and npm installed it then it would ask for the new major version of the cli package and the project may have references to the cli separately that wouldn't be ready for this update.

There may be something I don't understand about npm package management that might take care of this automatically (having both version of the cli package?), but I'm currently under the impression that both should be major version bumps.

igor-ribeiiro: Makes sense, thanks for the explanation

igor-ribeiiro:
Added an `deprecated` to logError in `office-addin-cli`. This way we don't need to do any major updates, as the function still exists.
@millerds PR is ready to be reviewed.

packages/custom-functions-metadata/package.json
```diff
@@ -41,6 +40,7 @@
     "mocha": "^9.1.1",
     "office-addin-lint": "^2.0.0",
     "office-addin-prettier-config": "^1.1.4",
+    "office-addin-usage-data": "^1.4.6",
```
millerds, line 43: Is this not a regular dependency anymore?

packages/office-addin-test-server/package.json
```diff
@@ -37,6 +37,7 @@
         "concurrently": "^6.3.0",
         "mocha": "^9.1.1",
         "office-addin-lint": "^2.0.0",
+        "office-addin-usage-data": "^1.4.6",
```
millerds, line 40: Is the cli supposed to be removed?

millerds (changes requested):

packages/office-addin-test-server/package.json
igor-ribeiiro: It previous only imported cli for the `logError` function, which is moving packages

packages/custom-functions-metadata/package.json
igor-ribeiiro: What do you mean? It appears as a dependency here.
Or do you mean devDependency vs other types?

packages/office-addin-test-server/package.json
millerds: Yes, but you're keeping the package import for the cli as well as adding the usage-data package that has the methods new location.

packages/custom-functions-metadata/package.json
millerds: cli was a dependency and you're replacing it with usage-data as a dev dependency.  If it was just getting the same method from a different package, I would think the nature of the dependency of it wouldn't change.

packages/office-addin-test-server/package.json
igor-ribeiiro: Yeah, both would work, but we are now deprecating "logError" from the cli package, so this is just us moving all calls from it to usage-data.

packages/custom-functions-metadata/package.json
igor-ribeiiro: My mistake, making it a regular dependency again

packages/office-addin-test-server/package.json
millerds: If you are moving the calls and referencing the new pacakge why do you still need the old package import?

packages/office-addin-debugging/src/cli.ts
```diff
@@ -4,7 +4,7 @@
 // Licensed under the MIT license.
 
 import * as commander from "commander";
-import { logErrorMessage } from "office-addin-cli";
+import { logErrorMessage } from "office-addin-usage-data";
```
millerds, line 7: shouldn't there be a matching package.json change to import usage-data?

packages/office-addin-dev-certs/src/cli.ts
```diff
@@ -4,7 +4,7 @@
 // Licensed under the MIT license.
 
 import * as commander from "commander";
-import { logErrorMessage } from "office-addin-cli";
+import { logErrorMessage } from "office-addin-usage-data";
```
millerds, line 7: matching package.json change?

packages/office-addin-sso/src/cli.ts
```diff
@@ -4,7 +4,7 @@
 // Licensed under the MIT license.
 
 import * as commander from "commander";
-import { logErrorMessage } from "office-addin-cli";
+import { logErrorMessage } from "office-addin-usage-data";
```
millerds, line 7: package.json change?

packages/office-addin-debugging/src/cli.ts
igor-ribeiiro: I already had it imported

packages/office-addin-dev-certs/src/cli.ts
igor-ribeiiro: It is there, probably already had it imported

packages/office-addin-sso/src/cli.ts
igor-ribeiiro: It is there as well. Probably already imported.

packages/office-addin-test-server/package.json
igor-ribeiiro: For this package it still needs `office-addin-cli` because it use the `parseNumber` function from it.

packages/office-addin-debugging/src/cli.ts
millerds: But you can now remove cli from package.json

packages/office-addin-dev-certs/src/cli.ts
millerds: But you can now remove cli from package.json

packages/office-addin-sso/src/cli.ts
millerds: But you can now remove cli from package.json

igor-ribeiiro: This one still had a call for parseNumber

packages/office-addin-dev-certs/src/cli.ts
igor-ribeiiro: This one still had a call for parseNumber

packages/office-addin-debugging/src/cli.ts
igor-ribeiiro: This one still had a call for parseNumber and for getPackageJsonScript

millerds (approved):

Issue #587, opened 2022-01-21
Mocking properties values in constructor of OfficeMockObject not working for outlook


igor-ribeiiro:
https://stackoverflow.com/questions/70798215/mocking-properties-using-office-addin-mock-yields-error-property-was-not-loade

PR #588, opened 2022-01-21
[Minor] Creating an isOutlook option for the object

igor-ribeiiro:
Creating an outlook flag to pass to the mocking object in order to bypass the need for a property wait for load and sync calls.
Previous to this commit, a property assigned in the constructor of `OfficeAddinMock` would always require a `load` call, which is not needed for Outlook.

Solves this issue: https://github.com/OfficeDev/Office-Addin-Scripts/issues/587

### DIFF
--- a/packages/office-addin-mock/README.md
+++ b/packages/office-addin-mock/README.md
@@ -115,9 +115,23 @@ Represents a mock Office object.
 #### Constructor
 
 The object parameter provides initial values for the mock object. (Optional)
+The host parameter identifies the host of the tests. (Optional)
 
 ```
-constructor(object?: Object); 
+constructor(object?: Object, host?: Host); 
+```
+
+Host can be any of the following:
+
+```Javascript
+Host {
+  excel = "excel",
+  outlook = "outlook",
+  powerpoint = "powerpoint",
+  word = "word",
+  other = "other",
+  notFound = "notFound",
+}
 ```
 
 #### Methods
--- a/packages/office-addin-mock/README.md
+++ b/packages/office-addin-mock/README.md
@@ -126,11 +126,13 @@ Host can be any of the following:
 ```Javascript
 Host {
   excel = "excel",
+  onenote = "onenote",
   outlook = "outlook",
   powerpoint = "powerpoint",
+  project = "project",
+  visio = "visio",
   word = "word",
-  other = "other",
-  notFound = "notFound",
+  unknow = "unknow",
 }
 ```
 
--- a/packages/office-addin-mock/README.md
+++ b/packages/office-addin-mock/README.md
@@ -118,21 +118,19 @@ The object parameter provides initial values for the mock object. (Optional)
 The host parameter identifies the host of the tests. (Optional)
 
 ```
-constructor(object?: Object, host?: Host); 
+constructor(object?: Object, host?: OfficeApp | undefined); 
 ```
 
 Host can be any of the following:
 
 ```Javascript
-Host {
-  excel = "excel",
-  onenote = "onenote",
-  outlook = "outlook",
-  powerpoint = "powerpoint",
-  project = "project",
-  visio = "visio",
-  word = "word",
-  unknow = "unknow",
+OfficeApp {
+  Excel = "excel",
+  OneNote = "onenote",
+  Outlook = "outlook",
+  PowerPoint = "powerpoint",
+  Project = "project",
+  Word = "word",
 }
 ```
 
--- a/packages/office-addin-mock/src/host.ts
+++ b/packages/office-addin-mock/src/host.ts
@@ -2,22 +2,23 @@ import { ObjectData } from "./objectData";
 
 export enum Host {
   excel = "excel",
+  onenote = "onenote",
   outlook = "outlook",
   powerpoint = "powerpoint",
+  project = "project",
+  visio = "visio",
   word = "word",
-  other = "other",
-  notFound = "notFound",
+  unknow = "unknow",
 }
 
 export function getHostType(object: ObjectData | undefined): Host {
+  let validHost = Host.unknow;
   if (object && object["host"]) {
-    let validHost = Host.other;
     Object.values(Host).forEach((host: string) => {
       if (object["host"] === host) {
         validHost = object["host"];
       }
     });
-    return validHost;
   }
-  return Host.notFound;
+  return validHost;
 }
--- a/packages/office-addin-mock/src/host.ts
+++ b/packages/office-addin-mock/src/host.ts
@@ -1,0 +1,23 @@
+import { ObjectData } from "./objectData";
+
+export enum Host {
+  excel = "excel",
+  outlook = "outlook",
+  powerpoint = "powerpoint",
+  word = "word",
+  other = "other",
+  notFound = "notFound",
+}
+
+export function getHostType(object: ObjectData | undefined): Host {
+  if (object && object["host"]) {
+    let validHost = Host.other;
+    Object.values(Host).forEach((host: string) => {
+      if (object["host"] === host) {
+        validHost = object["host"];
+      }
+    });
+    return validHost;
+  }
+  return Host.notFound;
+}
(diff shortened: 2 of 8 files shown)


millerds:
Are you sure this is not needed across the board for Outlook?  It might be worth it to check with the others who are more familiar with Office.js.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -5,9 +5,10 @@ import { isValidError, PossibleErrors } from "./possibleErrors";
  * @param object Object structure to provide initial values for the mock object (Optional)
  */
 export class OfficeMockObject {
-  constructor(object?: ObjectData) {
+  constructor(object?: ObjectData, isOutlook: boolean = false) {
```
millerds, line 8: I'm not sure I like having to put "true" here for all Outlook usage of the mock library.  It would be nice if there were some other way to figure that out and do the right thing automatically.

igor-ribeiiro:
> Are you sure this is not needed across the board for Outlook? It might be worth it to check with the others who are more familiar with Office.js.

I was pretty certain that `sync` and `load` are not needed in Outlook, but yeah, I'm not 100% sure this goes for all their APIs (there are a lot of them for me to look at).
I looked at the  [Outlook API reference](https://docs.microsoft.com/en-us/javascript/api/outlook?view=outlook-js-1.11) and it doesn't mention either `sync` or `load` and that's where I'm coming from.

packages/office-addin-mock/src/officeMockObject.ts
igor-ribeiiro: Agree, I will investigate where we can get this information

igor-ribeiiro: I searched on what options we have and doesn't seem likely that detecting automatically is feasible.
There is a function in `office-addin-manifest` that detects the adding type given a manifest path. 
The problem is that we don't know the manifest path in the tests. If we go into that route, I don't think passing the manifest path is easier than passing an Outlook flag, so I don't think we can detect the adding type automatically.

millerds: Is there some we could deduce it from the test json? Like is "mailbox" always the top level?

igor-ribeiiro: We could go and detect this example you said (or similar) and it could cover some cases automatically.
But the current design gives the freedom for the client to mock lower-level objects, instead of having to start from the top layer, so the mailbox might not be present sometimes.

The more I think of it, the less I think it can be done automatically. Maybe we can improve this proposed design if we think of a better way.

millerds: How would you mock child objects without mocking the parents?

This isn't something we should put out there if we want a better way later since it changes the method signature.  Customer s will put in the boolean for their outlook tests as the documentation tells them, but when we change it, the boolean goes away.  If we're going to require something (boolean in the signature or mailbox or something in the json) then I would prefer the json.

igor-ribeiiro: "How would you mock child objects without mocking the parents?"
You could just mock some lower-level object right away (like `range`). Outlook doesn't have `sync` so there is no need to start mocking from a high-level object.

I agree that some things in JSON look better.
I think of enforcing the following key:
```
mockJson = {
outlook: true
}```
So this outlook key would have to exist for outlook objects.

Comment (millerds) [reply packages/office-addin-mock/src/officeMockObject.ts:8]: I'm still not clear on the lower level stuff. 'range' isn't an Outlook object and does require sync.  You couldn't mock userProfile.DisplayName without putting context.mailbox in front of it could you?  My thought is that if outlook requires mailbox at the top (or some limited set of top level objects unique to outlook) then we could figure out it's outlook based on the structure of the json.

If you wanted to add a property to the json (I might be convinced to go with this) to indicate Outlook I wouldn't say "outlook: true" as that is too specific and unique.  I would say something like "host: outlook" which would allow us to use the same filed in the future for other hosts as well (maybe host specific api structure validation would be nice).

Comment (igor-ribeiiro) [reply packages/office-addin-mock/src/officeMockObject.ts:8]: You could userProfile.DisplayName without putting context.mailbox in front

I agree that outlook: true is too specific. But I am liking the host: outlook idea

Comment (millerds) [reply packages/office-addin-mock/src/officeMockObject.ts:8]: What's the scenario in which you would have userProfile at the top level of the office.js api in the code you are testing?  We aren't technically preventing in the mock . . . but the mock is for replacing the office.js api calls in tests and so it should match the api structure shouldn't it?

Comment (igor-ribeiiro) [reply packages/office-addin-mock/src/officeMockObject.ts:8]: it should match the API structure, you just don't need to start from the upper most layer.
You can just mock some lower-level API without having to go through the whole path until you get to it.

**Comment by igor-ribeiiro** [comment]
@millerds I just added the `host` option, so the PR is ready to be reviewed

File: packages/office-addin-mock/src/host.ts [hunk 2]
```diff
@@ -0,0 +1,23 @@
+import { ObjectData } from "./objectData";
+
+export enum Host {
+  excel = "excel",
+  outlook = "outlook",
+  powerpoint = "powerpoint",
+  word = "word",
+  other = "other",
+  notFound = "notFound",
+}
```
millerds, line 10: I don't think we should have the "other" option.  It should something we know or unknow (which should be used instead of "notFound).  We should also include the other apps that Office.js supports . . . we may focus on these 4 a lot, but I believe there are other, less used, hosts that can also work.

igor-ribeiiro: Sure! I just added all the supported hosts (following documentation) and changed the other/noFound to unknow!

millerds: Looks like I spelled that wrong (left off a letter) and you used that bad spelling :-).  Is should be "unknown"

packages/office-addin-mock/README.md
```diff
@@ -115,9 +115,23 @@ Represents a mock Office object.
...
+
+```Javascript
+Host {
+  excel = "excel",
+  outlook = "outlook",
+  powerpoint = "powerpoint",
+  word = "word",
+  other = "other",
+  notFound = "notFound",
+}
```
millerds, line 134: Need to update to match the enum.

packages/office-addin-mock/src/officeMockObject.ts
```diff
@@ -1,13 +1,21 @@
...
  */
 export class OfficeMockObject {
-  constructor(object?: ObjectData) {
+  constructor(object?: ObjectData, host?: Host) {
     this.properties = new Map<string, OfficeMockObject>();
     this.loaded = false;
+    if (host) {
+      this.host = host;
+    } else {
+      this.host = getHostType(object);
```
millerds, line 17: should we be checking for a valid host?  As it stands if we get "unknown" we still continue on our merry way.

igor-ribeiiro: I didn't want to make it mandatory for people to put the host option, so that's why I am not caring if someone doesn't define the host.
In this case, the default behavior is: "not outlook"

packages/office-addin-mock/src/host.ts
akrantz: You don't really need an enum option for other or notFound. Just use `undefined` for that case rather than the enum value, e.g. the type would be "Host | undefined". By the way, there is a similar enum in office-addin-debugging or dev-settings.

igor-ribeiiro: Sure, I am now using the existing enum for hosts, which was in "manifest" package.

packages/office-addin-mock/src/officeMockObject.ts
millerds: That's an ambiguous state, but I guess it's no different than we already are.

millerds (approved):

PR #589, opened 2022-01-21
Removing deprecated mock functions

igor-ribeiiro:
Removing deprecated `addMock`, `setMock` and `addMockFunction`. 
A Major update when releasing will be needed.

### DIFF
--- a/packages/office-addin-mock/src/officeMockObject.ts
+++ b/packages/office-addin-mock/src/officeMockObject.ts
@@ -15,32 +15,6 @@ export class OfficeMockObject {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


millerds:
This isn't just removing deprecated code.  You have an add for a function with the same name (and it looks like the same body) . . . is that intentional?

igor-ribeiiro:
> This isn't just removing deprecated code. You have an add for a function with the same name (and it looks like the same body) . . . is that intentional?

It is the same function, but right now it is private 😄

igor-ribeiiro:
Also, I change the function place for organizational sake. First public functions then private, all in alphabetical order.

millerds (approved):

PR #597, opened 2022-01-24
[Minor] Convert command on office-addin-cli

igor-ribeiiro:
Creating an upgrade project command to `office-addin-cli`


packages/office-addin-cli/src/commands.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { logErrorMessage } from "./log";
+import { usageDataObject } from "./defaults";
+
+export async function upgrade(manifestPath: string) {
+  try {
+    // Convert manifest .xml to manifest.json
```
akrantz, line 9: I would suggest having this functionality be in office-addin-manifest, e.g. in a function convertManifestToJson().

The CLI command would normally part of that package. There was an ideal that office-addin-cli could expose a CLI across the packages to make it easier but that wasn't ever done, and the package service to just have some CLI utility code which can be shared across the other packages.

akrantz: "upgrade" is a very generial term, so it might be better to use something more specific. I could see the user typing: `npx office-addin-manifest convert manifest.xml`

akrantz: Usually with NPM packages, it is related to upgrading the package version, so I worry about user confusion if upgrade is used as the term for converting the manifest from an xml format to a json format.

millerds: The goal of this command is more than just converting the manifest to json.  I've already created a command (that needs to get expanded) in office-addin-manifest that will do the conversion part, but the idea for MetaOS is that the projects will go through a onetime conversion that includes converting the manifest to json and updating package.json to have the right supporting packages and possibly some other things (if it's discovered that more is needed).  Since it's targeting the project it doesn't make sense to put it in the package targeting the manifest.

We could see that the cli package ended up being a place for a few utility functions and that it was going to be the central cli.  While the usage data package is currently targeting just telemetry we thought it made sense for it to be a general logging package and include logging to the console, or telemetry, or other places if need be and this would allow us to start making the cli package a central project cli and possibly have it be the point of entry rather than the various package entry points we have today (not sure if or how far we'd take that idea).

millerds: The verb 'upgrade' could potentially be confusing . . . any thoughts on a better term?

igor-ribeiiro: > The verb 'upgrade' could potentially be confusing . . . any thoughts on a better term?

@akrantz said `convert` previously. I think it is already a better term, but maybe `convertProject` would be more meaningful.

millerds: I think the cli should give context so that detailed verbs aren't needed.  We don't say "office-addin-mainfest validateManifest manifest.xml" we say "office-addin-manifest validate manifest.xml" because the context of manifest is understood.  If we're going to use this as a central cli then it would be nice if it could be understood that the context is the Office Addin project.

igor-ribeiiro: Resolving this thread as the name is now changed to `convert`

packages/office-addin-cli/src/commands.ts
```diff
@@ -4,15 +4,15 @@
 import { logErrorMessage } from "./log";
 import { usageDataObject } from "./defaults";
 
-export async function upgrade(manifestPath: string) {
+export async function convert(manifestPath: string) {
   try {
     // Convert manifest .xml to manifest.json
     // Do any more needed modifications on the code?
```
millerds, line 10: I know we don't have the xml -> json conversion library yet, but you should consider the other things that should be done and start adding the code for those.  
One example would be verifying the package.json has the needed package versions for the json manipulation and publish command (even if you don't have the final package versions, you can write the code to check for the current versions so that we just have to update the version numbers in the future).

igor-ribeiiro:
What you mean on checking version numbers?
Like, when we update the cli​ package to have this convert functionality it would already work.
I don't get what you mean on checking package versions, this should maninly be done normally by the package.json right?

Thank you,
Igor Ribeiro
DevX Team
________________________________

akrantz:
This functionality should go into office-addin-manifest. We don't want all functionality in one package. This is related to converting an XML manifest over to json. The ability to import a function from the package, and the CLI for that functionality should be part of specific packages.

igor-ribeiiro:
> This functionality should go into office-addin-manifest. We don't want all functionality in one package. This is related to converting an XML manifest over to JSON. The ability to import a function from the package and the CLI for that functionality should be part of specific packages.

Do you mean the function to check if the user has the necessary package versions?
The idea of that function is to make sure the user has updated some versions of their dependencies before making the conversion from the manifest XML to JSON

For example, if the user simply converts their project and doesn't update `office-addin-manifest` for example, then their code would not work anymore. So I guess this is more project-related than manifest-specific.

akrantz:
Not sure I really follow what y'all are trying to do. Might be best to chat about it.

millerds:
> Not sure I really follow what y'all are trying to do. Might be best to chat about it.

This is the command that will convert a package from the xml manifest based office add-in framework to the json manifest based shared (with teams) extension platform.  So far the conversion will include a call to convert the manifest from xml to json (library being built by OPC) and updating package.json to include the correct packages and versions to support the converted project (notably the version of office-addin-manifest package that can work with json manifest).  If there are any other updates to projects that need to happen in order to work with the new platform they will also be part of this command.

millerds:
Igor . . . some other things that should be added to this command is a confirmation prompt indicating the changes we are making to the project so developers know what's going to happen, and check for a project that may have already been converted (i.e. what happens if the command is run multiple times).

igor-ribeiiro:
> Igor . . . some other things that should be added to this command is a confirmation prompt indicating the changes we are making to the project so developers know what's going to happen, and check for a project that may have already been converted (i.e. what happens if the command is run multiple times).

Still missing the confirmation prompt, but I added some error checking as long as some tests to better check corner cases.

akrantz:
The function and command to invoke it for converting manifests from XML to JSON should be in office-addin-manifest.

For ensuring packages are up-to-date there is a bette way with NPM dependencies. First, if NPX is used to run a command, it will use the latest version automatically. If you want to have a command to look at the project package.json and update a depedency in it, there's a better way to go about it.

akrantz (changes requested):
I've made comments with my concerns.

packages/office-addin-cli/src/cli.ts
```diff
@@ -18,6 +19,15 @@ commander.on("command:*", function () {
   commander.help();
 });
 
+commander
+  .command("convert")
+  .option("-m, --manifest <manifest-path>", "Specify the location of the manifest file.  Default is './manifest.xml'")
+  .option(
+    "-p, --packageJson <packageJson-path>",
+    "Specify the location of the package.json file.  Default is './package.json'"
```
millerds, line 27: Doesn't this always have to been in the root?  Do we need to make this an option?

packages/office-addin-cli/src/convert.ts
```diff
@@ -0,0 +1,51 @@
...
+  Object.keys(content.devDependencies).forEach(function (key) {
+    if (depedentPackages.has(key)) {
+      const minVersion: string = depedentPackages.get(key) ?? "";
+      const version = semver.coerce(content.devDependencies[key]);
+
+      if (version && !semver.gte(version, minVersion)) {
+        throw new ExpectedError(
+          `Your version of the package ${key} should be at least ${depedentPackages.get(
+            key
+          )}. To update your package, run:\nnpm i ${key}@latest --save-dev`
```
millerds, line 46: My thought it to update the package reference for them, not just tell them to update it.

millerds (changes requested):

millerds:
> The function and command to invoke it for converting manifests from XML to JSON should be in office-addin-manifest.
> 
> For ensuring packages are up-to-date there is a bette way with NPM dependencies. First, if NPX is used to run a command, it will use the latest version automatically. If you want to have a command to look at the project package.json and update a depedency in it, there's a better way to go about it.

Converting the manifest is just one part of this command and the actual conversion code will come from OPC.  This is targeting the project, not just the manifest.

What would be the better way to update required project dependencies?

akrantz:
Currently, it's really only office-addin-debugging that really matters, because it specifies to use the dependent packages, office-addin-manifest, dev-settings, etc. So once that package is updated, it will use the newer values even if the project.json doesn't specific the latest. You could double-check if I'm wrong but I'm pretty sure that is right.

akrantz:
I've never seen other projects do this for developers. But if you want to have a command to update dependencies, that should be separate from converting manifests. Also, it can be run using "npx" which can get the latest package. The docs can include the switch that ensures using the latest rather than the package.json one. This is something else to double-check.

millerds:
> Currently, it's really only office-addin-debugging that really matters, because it specifies to use the dependent packages, office-addin-manifest, dev-settings, etc. So once that package is updated, it will use the newer values even if the project.json doesn't specific the latest. You could double-check if I'm wrong but I'm pretty sure that is right.

For the npm commands in the project both office-addin-debugging and office-addin-manifest (for the validation command) are referenced so it seems that both of those would need to be updated (I hadn't thought of office-addin-debugging yet).

millerds:
> I've never seen other projects do this for developers. But if you want to have a command to update dependencies, that should be separate from converting manifests. Also, it can be run using "npx" which can get the latest package. The docs can include the switch that ensures using the latest rather than the package.json one. This is something else to double-check.

The Teams toolkit (both the VS Code plugin and cli) do this for their project and they go even further (updating api calls in the customer's code to the new calling pattern).  I can't say that I've looked for other projects that might do something similar so I'm not sure how common it might be.  I also can't say if the Teams approach is the best practice.  I would, however, like to avoid having a long set of instructions on what customers need to do in order to convert their project to work on the new platform.

packages/office-addin-project/src/commands.ts
```diff
@@ -0,0 +1,20 @@
...
+import * as commander from "commander";
+import { logErrorMessage } from "office-addin-usage-data";
+import { usageDataObject } from "./defaults";
+import { convertProject } from "./convert";
+
+export async function convert(command: commander.Command) {
+  try {
+    const manifestPath: string = command.manifest ?? "./manifest.xml";
+
+    await convertProject(manifestPath);
```
millerds, line 13: I think we might want to inform the developers the changes that this command will make to their project (i.e. changing the manifest and updating the project dependencies) and ask for confirmation of the changes just so they are aware of what is going to happen.  Git will help let them see the changes before committing them, but a warning might be useful.

packages/office-addin-project/src/convert.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (i === depedentPackages.length - 2) {
+      messageToBePrinted += " and";
+    } else {
+      messageToBePrinted += ",";
+    }
+  }
+
+  command += ` --save-dev`;
+  console.log(messageToBePrinted.slice(0, -1));
+  exec(command);
```
millerds, line 50: Should we also trigger the 'npm install' command like we do in generator office?

millerds (changes requested):

packages/office-addin-project/src/convert.ts
igor-ribeiiro: I think it is not needed, as I'm already running the npm install for the new packages

packages/office-addin-project/src/convert.ts
```diff
@@ -0,0 +1,51 @@
...
+    "@microsoft/teamsfx",
+    "office-addin-debugging",
+    "office-addin-manifest",
+  ];
+  let command: string = "npm install";
+  let messageToBePrinted: string = "Installing latest versions of";
+
+  for (let i = 0; i < depedentPackages.length; i++) {
+    const depedentPackage = depedentPackages[i];
+    command += ` ${depedentPackage}@latest`;
```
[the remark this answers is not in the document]
igor-ribeiiro: Yeah, I was first doing that way, but then I realized that npm already does that (check the versions and only upgrade if necessary).
So we might as well use what is already implemented

packages/office-addin-project/src/convert.ts
millerds: Sorry . . . I didn't look carefully at the command itself.  For some reason I got it in my mind that it was just updating package.json.

packages/office-addin-project/src/convert.ts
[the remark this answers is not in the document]
igor-ribeiiro: Yes, I think that's the behavior of `npm`.
But I am always using `@latest`, so I don't think we could run into cases of downgrade

packages/office-addin-project/src/commands.ts
igor-ribeiiro: Sure, I just added a confirmation prompt

packages/office-addin-project/src/convert.ts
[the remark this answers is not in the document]
igor-ribeiiro: I agree. 
But after some investigation, I don't think npm has a command built in to specify minimum versions including major versions.
Because of that, I think it doesn't make sense for us to spend much time on this, as @latest I think is enough already (even if not the best option).

packages/office-addin-project/src/convert.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  if (!fs.existsSync(manifestPath)) {
+    throw new ExpectedError(
+      `The manifest file '${manifestPath}' does not exist`
+    );
+  }
+
+  const shouldContinue = await asksForUserConfirmation();
+  console.log(`Should return returned: ${shouldContinue}`);
```
millerds, line 25: What's this for?

packages/office-addin-project/src/convert.ts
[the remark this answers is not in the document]
millerds: In the near term it works . . . but the longer this package exist the more we run the risk of a problem (i.e. upgrading to a breaking change).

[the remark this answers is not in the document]
igor-ribeiiro: After looking more into the npm official docs: https://docs.npmjs.com/cli/v8/commands/npm-install#description
There is no "upgrade if needed" command. Or we make the command upgrade all the way (until some specific upper boundary) or we hard code some boundary: only upgrade until 2.x.x major (but it might have downgrades).

If we want an "upgrade if needed" command, we would have to implement it ourselves (reading and parsing the JSON). I don't think it is worth taking the time to do this work.

igor-ribeiiro:
> It occurred to me that another thing needing to be changes in the project is the package.json scripts section. There are references to "manifest.xml" that will need to be updated to "manifest.json".

Where are those references? I don't see any `.xml` reference in the `office-addin-project` `package.json`

millerds:
> > It occurred to me that another thing needing to be changes in the project is the package.json scripts section. There are references to "manifest.xml" that will need to be updated to "manifest.json".
> 
> Where are those references? I don't see any `.xml` reference in the `office-addin-project` `package.json`

Sorry . . . not the project package package.json . . . the projects the script is being run against.  The templates output npm scripts like 'start' that include a reference to manifest.xml file.  You'll need to crack the project that is being updated's package.json and switch those to manifest.json.

igor-ribeiiro:
> > > It occurred to me that another thing needing to be changes in the project is the package.json scripts section. There are references to "manifest.xml" that will need to be updated to "manifest.json".
> > 
> > 
> > Where are those references? I don't see any `.xml` reference in the `office-addin-project` `package.json`
> 
> Sorry . . . not the project package package.json . . . the projects the script is being run against. The templates output npm scripts like 'start' that include a reference to manifest.xml file. You'll need to crack the project that is being updated's package.json and switch those to manifest.json.

Now the package is updating the `.xml` references in the `package.json`!

igor-ribeiiro:
@millerds can you review this PR again? I think it is in a decent place to merge

packages/office-addin-project/README.md
```diff
@@ -0,0 +1,28 @@
...
+https://docs.microsoft.com/en-us/office/dev/add-ins/develop/add-in-manifests).
+
+## Command-Line Interface
+* [convert](#info)
+
+#
+
+### convert
+
+Converts the Office Addin Code.
```
millerds, line 15: Convert it t6o what?

millerds (dismissed):
I take it you'll add the backup folder and add the package.json file to it when you don the xml=>json conversion integration.

packages/office-addin-project/README.md
igor-ribeiiro: Not sure what to put here yet. I will just mention from XML to json manifest for now.
We will probably just publish this package when we will convert more stuff.

igor-ribeiiro:
> I take it you'll add the backup folder and add the package.json file to it when you don the xml=>json conversion integration.

Sure, I will add a task for us to not forget about this backup work

PR #600, opened 2022-02-03
Adjust the date used to check cert validity

millerds:
Customers have reported an error in the host taskpane saying no valid cert was found, but the tooling says the cert is valid.  We haven't been able to repro this . . . until yesterday when I encountered this.  I as investigated it became clear that my cert was expired at 2:30pm, but office-addin-dev-certs said it was still valid when checking against a time of 12AM on the same day (anything after is considered expired).


igor-ribeiiro:
What's the context for this PR? Why is this needed?

igor-ribeiiro (approved):

PR #608, opened 2022-02-25
Update registering the add-in to publish a packaged json manifest.

millerds:
This should allow for sideloading the json version of the manifest.

### DIFF
--- a/packages/office-addin-dev-settings/package.json
+++ b/packages/office-addin-dev-settings/package.json
@@ -22,6 +22,7 @@
     "Office Add-in"
   ],
   "dependencies": {
+    "@microsoft/teamsfx-cli": "0.11.0-rc.0",
     "child_process": "^1.0.2",
     "commander": "^6.2.0",
     "fs-extra": "^9.0.1",
--- a/packages/office-addin-dev-settings/src/publish.ts
+++ b/packages/office-addin-dev-settings/src/publish.ts
@@ -1,0 +1,18 @@
+import * as childProcess from "child_process";
+
+/* global console */
+
+export async function publish(zipPath: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const command = `npx @microsoft/teamsfx-cli@0.11.0-rc.0 provision manifest --file-path ${zipPath}`;
+
+    childProcess.exec(command, (error, stdout) => {
+      if (error) {
+        reject(stdout);
+      } else {
+        console.log("Successfully registered package with https://dev.teams.microsoft.com/apps");
+        resolve();
+      }
+    });
+  });
+}
--- a/packages/office-addin-dev-settings/src/publish.ts
+++ b/packages/office-addin-dev-settings/src/publish.ts
@@ -2,7 +2,7 @@ import * as childProcess from "child_process";
 
 /* global console */
 
-export async function publish(zipPath: string): Promise<void> {
+export async function registerWithTeams(zipPath: string): Promise<void> {
   return new Promise((resolve, reject) => {
     const command = `npx @microsoft/teamsfx-cli@0.11.0-rc.0 provision manifest --file-path ${zipPath}`;
 
--- a/packages/office-addin-dev-settings/src/publish.ts
+++ b/packages/office-addin-dev-settings/src/publish.ts
@@ -1,18 +1,26 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
 import * as childProcess from "child_process";
+import * as fs from "fs";
 
 /* global console */
 
 export async function registerWithTeams(zipPath: string): Promise<void> {
   return new Promise((resolve, reject) => {
-    const command = `npx @microsoft/teamsfx-cli@0.11.0-rc.0 provision manifest --file-path ${zipPath}`;
+    if (zipPath.endsWith(".zip") && fs.existsSync(zipPath)) {
+      const command = `npx @microsoft/teamsfx-cli@0.11.0-rc.0 provision manifest --file-path ${zipPath}`;
 
-    childProcess.exec(command, (error, stdout) => {
-      if (error) {
-        reject(stdout);
-      } else {
-        console.log("Successfully registered package with https://dev.teams.microsoft.com/apps");
-        resolve();
-      }
-    });
+      childProcess.exec(command, (error, stdout) => {
+        if (error) {
+          reject(stdout);
+        } else {
+          console.log("Successfully registered package with https://dev.teams.microsoft.com/apps");
+          resolve();
+        }
+      });
+    } else {
+      reject(new Error(`The file '${zipPath}' is not valid`));
+    }
   });
 }
--- a/packages/office-addin-dev-settings/test/files/manifests/manifest.json
+++ b/packages/office-addin-dev-settings/test/files/manifests/manifest.json
@@ -1,0 +1,121 @@
+{
+  "$schema": "https://raw.githubusercontent.com/OfficeDev/microsoft-teams-app-schema/preview/DevPreview/MicrosoftTeams.schema.json",
+  "id": "df2fa15f-332a-47aa-87c2-e30c20686eee",
+  "manifestVersion": "m365DevPreview",
+  "version": "1.0.0",
+  "name": {
+    "short": "My Office Add-in",
+    "full": "My Office Add-in"
+  },
+  "description": {
+    "short": "A template to get started.",
+    "full": "A template to get started."
+  },
+  "icons": {
+    "outline": "",
+    "color": ""
+  },
+  "accentColor": "#FFFFFF",
+  "developer": {
+    "name": "Contoso",
+    "websiteUrl": "https://www.contoso.com",
+    "privacyUrl": "https://www.contoso.com/privacy",
+    "termsOfUseUrl": "https://www.contoso.com/terms"
+  },
+  "localizationInfo": {
+    "defaultLanguageTag": "en-us"
+  },
+  "extension": {
+    "requirements": {
+      "scopes": ["Mailbox"],
+      "capabilities": [{ "name": "Mailbox", "minVersion": "1.3" }]
+    },
+    "runtimes": [
+      {
+        "id": "ShowTaskpane",
+        "type": "general",
+        "code": {
+          "page": "https://localhost:3000/taskpane.html"
+        },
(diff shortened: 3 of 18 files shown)


packages/office-addin-dev-settings/package.json
```diff
@@ -22,6 +22,7 @@
     "Office Add-in"
   ],
   "dependencies": {
+    "@microsoft/teamsfx-cli": "0.11.0-rc.0",
```
igor-ribeiiro, line 25: I am not aware of what this `-rc.0` mean. Can't we use the normal `^0.11` version?

packages/office-addin-dev-settings/src/publish.ts
```diff
@@ -0,0 +1,18 @@
+import * as childProcess from "child_process";
+
+/* global console */
+
+export async function publish(zipPath: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const command = `npx @microsoft/teamsfx-cli@0.11.0-rc.0 provision manifest --file-path ${zipPath}`;
```
akrantz, line 7: Rather than use npx, is there a way to import and use the @microsoft/teamsfx-cli package to do this? It would be more efficient.

packages/office-addin-dev-settings/src/publish.ts
```diff
@@ -0,0 +1,18 @@
+import * as childProcess from "child_process";
+
+/* global console */
+
+export async function publish(zipPath: string): Promise<void> {
```
akrantz, line 5: `publish` here is very generic. Would `publishToTeams` or `registerWithTeams` be better?

packages/office-addin-dev-settings/src/publish.ts
```diff
@@ -0,0 +1,18 @@
+import * as childProcess from "child_process";
```
igor-ribeiiro, line 1: Can we add the:
```
```
In the beginning of the file for consistency?

packages/office-addin-dev-settings/test/unit-tests/test.ts
```diff
@@ -319,28 +319,6 @@ describe("Registration", function() {
         assert.strictEqual(first.id, firstManifestId);
         assert.strictEqual(first.manifestPath, firstRegisteredManifestPath);
       });
-      if (process.platform === "win32") {
```
igor-ribeiiro, line 322: Why did we remove these tests?

packages/office-addin-dev-settings/src/publish.ts
igor-ribeiiro, line 7: Given that this function is exported, can we add error checking for the `zipPath`?
Like seeing if it exists and checking if it is really a zip.

If the `teams-fx-cli` already does this, that's fine

packages/office-addin-dev-settings/test/unit-tests/test.ts
```diff
@@ -319,28 +319,6 @@ describe("Registration", function() {
...
-          const registeredAddins = await devSettings.getRegisterAddIns();
-          const [first, second] = registeredAddins;
-          assert.strictEqual(registeredAddins.length, 2);
-          assert.strictEqual(first.id, firstManifestId);
-          assert.strictEqual(second.id, secondManifestId);
-          assert.strictEqual(first.manifestPath, firstManifestPath);
-          assert.strictEqual(second.manifestPath, secondManifestPath);
-        });
-      }
     });
```
igor-ribeiiro, line 322: Can we add some tests for the register/publish steps with JSON files?
I know we can't automate the happy path, but we can add tests for non-existent zip files and wrong formats for example

packages/office-addin-dev-settings/src/publish.ts
millerds: While I don't have all the details, the way their libraries are designed you have to provide the auth token for methods that end up calling the authenticated endpoint like is happening for this publish.  In order to get the auth prompting they already have built-in, it currently has to go through the cli or the VS Code extension.  Doing npx and using the cli was the simplest option we could get without doing a bunch of auth work ourselves or getting them to rearchitect their libraries to put the pieces together.

packages/office-addin-manifest/src/cli.ts
```diff
@@ -28,7 +28,6 @@ commander
 commander
   .command("export")
   .option("-m, --manifest <manfest-path>", "Specify the location of the manifest file.  Default is './manifest.json'")
-  .option("-a, --assets <assets-folder-path>", "Specify the location of the assets folder.  Default is './assets'")
```
igor-ribeiiro, line 31: Why don't we need to specify the assets folder anymore? Did the design change?
Also, please add this command to the README of this package.

packages/office-addin-dev-settings/src/publish.ts
millerds: Good point.  I can do this now, but note that eventually it will be some yet to be named shared endpoint not specific to Teams.

packages/office-addin-dev-settings/test/files/manifests/manifest.json
```diff
@@ -0,0 +1,121 @@
...
+  "name": {
+    "short": "My Office Add-in",
+    "full": "My Office Add-in"
+  },
+  "description": {
+    "short": "A template to get started.",
+    "full": "A template to get started."
+  },
+  "icons": {
+    "outline": "",
```
igor-ribeiiro, line 15: Can we also add the values for these fields?
```
    "outline": "assets/icon-32.png",
    "color": "assets/icon-32.png"
```

igor-ribeiiro (changes requested):
Can we add the project type to the telemetry in the `office-addin-dev-settings` project?  
It would be nice to have the data of people using the register command for `.xml` vs `.json` based projects. 

The telemetry could be put in: `packages\office-addin-dev-settings\src\commands.ts` in the `register` function.
We then add the project type as a parameter to the telemetry.

packages/office-addin-dev-settings/package.json
millerds: Presumably "rc" stands for "release candidate".  This is basically putting an unofficial version out there for us to try out before they make it part of an official release.  We'll need to update this once the official release includes the command we need.

packages/office-addin-dev-settings/test/unit-tests/test.ts
millerds: These test are for the Win32 specific method that uses two arguments allowing for the registry to have a key other than the add-in id in the manifest.  However we currently don't expose anyway to do that directly and the only place we use it we are reading the manifest specifically.  I change the method signature add moved the id reading down a level to be more efficient and so the test no longer makes sense as the method only has a single argument.

packages/office-addin-dev-settings/src/publish.ts
millerds: Good point.  While the "export" in this case is only internal to this package (you can't call it directly from outside the package) it would be good to make sure the zip file does exist.

packages/office-addin-dev-settings/test/unit-tests/test.ts
millerds: Hmm . . . there is some value in verifying failure cases.  What do you mean by "wrong formats"?

packages/office-addin-manifest/src/cli.ts
millerds: I had understood that certain icon needed to be a part of the zip file.  However now that I'm actually publishing a zip I can see that it's not required.  On top of that, the ones they need are referenced in the json manifest and these may not be the right ones.  On top of that the register command only passes in the manifest file and I don't have an assets directory to reference and so it wouldn't work unless the assets were in a specific location.

I'm going to switch to using the paths in the manifest (which requires manifest manipulation work) and optionally include them in the zip file . . . no need to specify them directly.

packages/office-addin-dev-settings/test/unit-tests/test.ts
igor-ribeiiro: By the wrong format I mean the file is not a `.zip` or it might be corrupted.
Either way, maybe the `jszip` package already handles these exemptions, but we could add tests to verify if these cases throw errors

millerds: Since this is sending it to teams cli I'll let them check file format (we don't need to crack open the file), but I'll do some other basic checking.

igor-ribeiiro (approved):

igor-ribeiiro (changes requested):

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -12,8 +12,8 @@
     "full": "A template to get started."
   },
   "icons": {
-    "outline": "assets/icon-32.png",
-    "color": "assets/icon-32.png"
+    "outline": "test/assets/icon-32.png",
```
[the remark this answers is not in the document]
millerds: There are tests for making the zip file and so the test assets (already there) do get used.

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -30,151 +30,92 @@
...
-                ]
-            }
-        ],
-        "ribbons": [
-            {
-                "contexts": [
-                    "MessageReadCommandSurface"
+        "lifetime": "short",
+        "actions": [
+          {
```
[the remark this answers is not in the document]
millerds: Seems like a preference.  I've just gotten in the habit of 2 spaces for everything because that is what eslint is telling us to do everywhere.

packages/office-addin-manifest/src/export.ts
```diff
@@ -29,34 +30,35 @@ function createZip(manifestPath: string = "manifest.json"): jszip {
   }
 
   // Need json manifest manipulation before this can work.
-  // const manifest = readManifest(manifestPath);
-  // addIconFile(manifest.icon?.color, zip);
-  // addIconFile(manifest.icon?.outline, zip);
+  const manifest: TeamsAppManifest = await ManifestUtil.loadFromPath(manifestPath);
+  addIconFile(manifest.icons?.color, zip);
+  addIconFile(manifest.icons?.outline, zip);
```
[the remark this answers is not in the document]
millerds: Not when I tried it . . . the zip library seems to handle (probably by updating the existing entry).

igor-ribeiiro (approved):

PR #609, opened 2022-02-25
Added JSON manifest manipulation

igor-ribeiiro:
Added manifest JSON manipulation code.

### DIFF
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
@@ -2,8 +2,6 @@
 // Licensed under the MIT license.
 
 import { ManifestInfo } from "../manifestInfo";
-import * as util from "util";
-import * as fs from "fs";
 
 export abstract class ManifestHandler {
   /* eslint-disable no-unused-vars */
@@ -11,7 +9,6 @@ export abstract class ManifestHandler {
     this.manifestPath = manifestPath;
     this.fileData = "";
   }
-
 
   abstract modifyManifest(guid?: string, displayName?: string): Promise<any>;
   abstract parseManifest(): Promise<ManifestInfo>;
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
@@ -12,7 +12,7 @@ export abstract class ManifestHandler {
     this.fileData = "";
   }
 
-  public async updateFileData(): Promise<string> {
+  public async readFromManifestFile(): Promise<string> {
     try {
       const fileData: string = await util.promisify(fs.readFile)(this.manifestPath, {
         encoding: "utf8",
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
@@ -7,13 +7,12 @@ import * as fs from "fs";
 
 export abstract class ManifestHandler {
   /* eslint-disable no-unused-vars */
-  constructor() {
-    this.manifestPath = "";
+  constructor(manifestPath: string) {
+    this.manifestPath = manifestPath;
     this.fileData = "";
   }
 
   public async readFromManifestFile(manifestPath: string): Promise<string> {
-    this.manifestPath = manifestPath;
     try {
       const fileData: string = await util.promisify(fs.readFile)(manifestPath, {
         encoding: "utf8",
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
@@ -12,15 +12,15 @@ export abstract class ManifestHandler {
     this.fileData = "";
   }
 
-  public async readFromManifestFile(manifestPath: string): Promise<string> {
+  public async updateFileData(): Promise<string> {
     try {
-      const fileData: string = await util.promisify(fs.readFile)(manifestPath, {
+      const fileData: string = await util.promisify(fs.readFile)(this.manifestPath, {
         encoding: "utf8",
       });
       this.fileData = fileData;
       return fileData;
     } catch (err) {
-      throw new Error(`Unable to read data for manifest file: ${manifestPath}. \n${err}`);
+      throw new Error(`Unable to read data for manifest file: ${this.manifestPath}. \n${err}`);
     }
   }
 
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
@@ -4,16 +4,12 @@
 import { ManifestInfo } from "../manifestInfo";
 
 export abstract class ManifestHandler {
-  /* eslint-disable no-unused-vars */
   constructor(manifestPath: string) {
     this.manifestPath = manifestPath;
-    this.fileData = "";
   }
 
   abstract modifyManifest(guid?: string, displayName?: string): Promise<any>;
   abstract parseManifest(): Promise<ManifestInfo>;
   abstract writeManifestData(manifestData: any): Promise<void>;
   manifestPath: string;
-  fileData: string;
-  /* eslint-enable no-unused-vars */
 }
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
@@ -12,17 +12,6 @@ export abstract class ManifestHandler {
     this.fileData = "";
   }
 
-  public async readFromManifestFile(): Promise<string> {
-    try {
-      const fileData: string = await util.promisify(fs.readFile)(this.manifestPath, {
-        encoding: "utf8",
-      });
-      this.fileData = fileData;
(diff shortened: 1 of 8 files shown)


packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
```diff
@@ -7,21 +7,20 @@ import * as fs from "fs";
...
-  constructor() {
-    this.manifestPath = "";
+  constructor(manifestPath: string) {
+    this.manifestPath = manifestPath;
     this.fileData = "";
   }
 
-  public async readFromManifestFile(manifestPath: string): Promise<string> {
-    this.manifestPath = manifestPath;
+  public async updateFileData(): Promise<string> {
```
millerds, line 15: updateFileData?  Looks like this is a read not a write.

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
```diff
@@ -1,29 +1,69 @@
...
     try {
-      const file = await JSON.parse(this.fileData);
+      const file: TeamsAppManifest = await JSON.parse(this.fileData);
+      return this.translateTeamsAppManifestToManifestInfo(file);
     } catch (err) {
       throw new Error(`Unable to read data for manifest file: ${this.manifestPath}. \n${err}`);
     }
+  }
+
+  translateTeamsAppManifestToManifestInfo(teamsAppManifest: TeamsAppManifest): ManifestInfo {
```
millerds, line 37: Nit: Shorter to the point name might be a little better.  Like "getManifestInfo" or something to that effect.

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
```diff
@@ -1,29 +1,69 @@
...
+    await this.updateFileData();
   }
   /* eslint-enable @typescript-eslint/no-unused-vars */
 }
+
+/**
+ * Save manifest to .json file
+ * @param filePath path to the manifest.json file
+ */
+async function writeToPath(path: string, manifest: TeamsAppManifest): Promise<void> {
```
millerds, line 67: Isn't this going to be part of the teams-lib?

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
```diff
@@ -7,21 +7,20 @@ import * as fs from "fs";
...
 export abstract class ManifestHandler {
   /* eslint-disable no-unused-vars */
-  constructor() {
-    this.manifestPath = "";
+  constructor(manifestPath: string) {
+    this.manifestPath = manifestPath;
     this.fileData = "";
   }
 
-  public async readFromManifestFile(manifestPath: string): Promise<string> {
```
millerds, line 15: Any particular reason for the name change?

packages/office-addin-manifest/src/manifestOperations.ts
```diff
@@ -49,15 +48,15 @@ export namespace OfficeAddinManifest {
...
   } else if (manifestPath.endsWith(".xml")) {
-    manifestHandler = new ManifestHandlerXml();
+    manifestHandler = new ManifestHandlerXml(manifestPath);
   } else {
     const extension: string = manifestPath.split(".").pop() ?? "<no extension>";
     throw new Error(
       `Manifest operations are not supported in .${extension}.\nThey are only supported in .xml and in .json.`
     );
   }
-  await manifestHandler.readFromManifestFile(manifestPath);
```
millerds, line 61: is readFromManifestFile being used anywhere anymore?

millerds (changes requested):

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
igor-ribeiiro: Yeah, this is the exact copy of the team's lib. I'm just copying-pasting, for now, to be able to develop against it.

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
igor-ribeiiro: Yeah, it is a read, not a writer. The name got confusing
My idea is that this function updates the internal `fileData` variable, which is needed after making modifications to the manifest.
I will see if I can improve this design to avoid these confusions.

packages/office-addin-manifest/src/manifestOperations.ts
igor-ribeiiro: This was just a side effect of the name change. The function is still being used here

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
igor-ribeiiro: I don't think there is a strong reason to, I will just keep the name the same

packages/office-addin-manifest/src/manifestOperations.ts
millerds: Sorry . . . I was mixing this up with "readManifestFile" which is in manifestOperations instead the handler's "readFromManifestFile".  It looked like it was just used by the tests now . . . but I just checked and it's exported outside the package (not just internally).  So we're good.

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
```diff
@@ -1,29 +1,61 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
+import { TeamsAppManifest } from "@microsoft/teamsfx-api";
```
millerds, line 4: So the TeamsAppManifest interface is not part of the new teams-manifest pacakge?

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
```diff
@@ -1,29 +1,61 @@
...
+import { v4 as uuidv4 } from "uuid";
 import { ManifestInfo } from "../manifestInfo";
 import { ManifestHandler } from "./manifestHandler";
 
 export class ManifestHandlerJson extends ManifestHandler {
   /* eslint-disable @typescript-eslint/no-unused-vars */
-  async modifyManifest(guid?: string, displayName?: string): Promise<any> {
-    throw new Error("Manifest cannot be modified in .json files");
+  async modifyManifest(guid?: string, displayName?: string): Promise<TeamsAppManifest> {
+    const teamsAppManifest: TeamsAppManifest = await JSON.parse(this.fileData);
```
millerds, line 13: Isn't the new library supposed to have manifest reading as part of it?

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
```diff
@@ -7,21 +7,20 @@ import * as fs from "fs";
...
-    this.manifestPath = manifestPath;
+  public async readFromManifestFile(): Promise<string> {
     try {
-      const fileData: string = await util.promisify(fs.readFile)(manifestPath, {
+      const fileData: string = await util.promisify(fs.readFile)(this.manifestPath, {
```
millerds, line 17: Is there a reason we are reading the fileData and then the TeamsAppManifest objects in separate steps?

millerds (changes requested):

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
igor-ribeiiro: I thought it would be needed previously. I just changed the code structure so we now just use the teams read package.

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
igor-ribeiiro: Sure, I just added the reading with the library now

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
igor-ribeiiro: Fixed

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
```diff
@@ -2,27 +2,12 @@
...
-    try {
-      const fileData: string = await util.promisify(fs.readFile)(manifestPath, {
-        encoding: "utf8",
-      });
-      this.fileData = fileData;
-      return fileData;
-    } catch (err) {
-      throw new Error(`Unable to read data for manifest file: ${manifestPath}. \n${err}`);
-    }
+    this.fileData = "";
```
millerds, line 10: Why do we need this intermediate object?  It's never read and only stored in one method (which just returns the same data anyway.

Unless you change the method that accepts or return the data as arguments to just use the internal property instead . . . but I'm not sure if that would be a breaking change from how things work right now (exported API methods).

millerds (changes requested):

packages/office-addin-manifest/src/manifestHandler/manifestHandler.ts
igor-ribeiiro: It is only currently needed for the `xml` case. I will move the manipulation of this variable to the `xml` class.

millerds (approved):

PR #612, opened 2022-03-03
Manifest validation

igor-ribeiiro:
Add manifest JSON manifest to the package.


packages/office-addin-manifest/src/validate.ts
```diff
@@ -65,13 +66,24 @@ export async function validateManifest(
   manifestPath: string,
   verifyProduction: boolean = false
 ): Promise<ManifestValidation> {
-  try {
```
millerds, line 68: Is there a reason we are removing the try/catch?

packages/office-addin-manifest/src/validate.ts
```diff
@@ -65,13 +66,24 @@ export async function validateManifest(
...
-    const validation: ManifestValidation = new ManifestValidation();
-    const clientId: string = verifyProduction ? "Default" : "devx";
-
-    // read the manifest file to ensure the file path is valid
-    await OfficeAddinManifest.readManifestFile(manifestPath);
-
+  const validation: ManifestValidation = new ManifestValidation();
+  const clientId: string = verifyProduction ? "Default" : "devx";
```
millerds, line 70: This is specific to the xml checking

packages/office-addin-manifest/src/validate.ts
```diff
@@ -65,13 +66,24 @@ export async function validateManifest(
...
+  await OfficeAddinManifest.readManifestFile(manifestPath);
+
+  if (manifestPath.endsWith(".json")) {
+    const manifest: TeamsAppManifest = await ManifestUtil.loadFromPath(manifestPath);
+    const validationResult: string[] = await ManifestUtil.validateManifest(manifest);
+
+    if (validationResult.length !== 0) {
+      // There are errors
+      validation.isValid = false;
+      validation.jsonErrors = validationResult;
```
millerds, line 82: It may be quicker to just add a json manifest version of the results, but I don't think it's good design to have one off properties like this.  We aren't going to have json and xml results at the same time.

Other things to consider . . . the json manifest results output should looks as similar to the xaml ones (as much as possible) and should allow for warnings vs. error (I'm not sure if there are warnings in the json manifest validation results . . . but you should check).

millerds (changes requested):

packages/office-addin-manifest/src/validate.ts
igor-ribeiiro: We already have the try/catch on the `commands.ts` (the cli) for telemetry. So I thought this telemetry just had a lot of false positives because it keeps getting logged in our unit tests.
So now we just have the telemetry on the CLI part of the command

packages/office-addin-manifest/src/validate.ts
igor-ribeiiro: The json validation is literally just an array of strings. Each string is an error, they just have errors, no other case.
But I agree they should be more similar, so I adapted this new validation format to ours.

packages/office-addin-manifest/src/commands.ts
```diff
@@ -92,7 +92,9 @@ function logManifestValidationWarnings(warnings: ManifestValidationIssue[] | und
 }
 
 function logManifestValidationIssue(issue: ManifestValidationIssue) {
-  console.log(`${issue.title}: ${issue.content}` + (issue.helpUrl ? ` (link: ${issue.helpUrl})` : ``));
+  console.log(
+    (issue.title ? `${issue.title}: ` : ``) + `${issue.content}` + (issue.helpUrl ? ` (link: ${issue.helpUrl})` : ``)
+  );
```
millerds, line 97: What are the cases where the title is going to be empty?

packages/office-addin-manifest/src/validate.ts
millerds: This different is that the method can be called outside of the cli (like the test cases themselves do) and since the method is exported outside of the package then anyone can call this method as an API instead of via the cli.  We'd be losing all that telemetry as well.

millerds (changes requested):

packages/office-addin-manifest/src/validate.ts
igor-ribeiiro: Yeah, we can keep this telemetry

packages/office-addin-manifest/src/commands.ts
igor-ribeiiro: None! I will remove this, as I'm already adding titles to the json validation

millerds (approved):

PR #614, opened 2022-03-15
Removing file base path

igor-ribeiiro:
Sideload code was assuming the manifestPath served had always the same name from the one from the file, which isn't necessarily the case. 

There is no issue with this right now, but using `basename` didn't make sense, so I'm just removing it before it can become a problem.

### DIFF
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -360,7 +360,7 @@ export async function sideloadAddIn(
         if (!isSideloadingSupportedForWebHost(app)) {
           throw new ExpectedError(`Sideload to the ${getOfficeAppName(app)} web app is not supported.`);
         }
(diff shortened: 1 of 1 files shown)


millerds:
In looking at the history of the changes this was made as part of adding support for web based debugging.  The "path.basename" just gets the file name from the path as opposed to using the whole path (relative or otherwise).  I don't see how removing this helps given that in both cases the assumptions is that the path being given in the query parameter is where webpack is serving it from whether that's just the file name or with that path.

I don't think changing this now helps us and we risk unexpectedly breaking something else.  We should come up with a whole solution to providing the query parameter with the webpack location of the manifest.

igor-ribeiiro:
Yeah, this by itself is not really enough. It would be needed when we have the webpack having the built manifest name as a parameter (when we support name modifications by webpack).
We can leave this change until we fund the work completely.

PR #629, opened 2022-04-22
Json xml convert

millerds:
When calling the sideload command in office-addin-dev-setting if it's a json file then we will call the exe tool to convert it to xml (stored in the temp directory) and then sideload to Outlook as normal xml would.  This shouldn't change any json manifest handling.


packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -386,3 +399,63 @@ export async function sideloadAddIn(
...
+    return foundDotNet;
+  } catch (err) {
+    return false;
+  }
+}
+
+async function convertJsonToXmlManifest(manifestPath: string): Promise<string> {
+  return new Promise((resolve, reject) => {
+    if (manifestPath.endsWith(".json") && fs.existsSync(manifestPath)) {
+      console.log("Converting json to back compat xml");
```
igor-ribeiiro, line 426: Does the client need to be aware we are doing this conversion? I feel we don't need to log that this is happening

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -386,3 +399,63 @@ export async function sideloadAddIn(
...
+          resolve(newManifestPath);
+        }
+      });
+    } else {
+      reject(new Error(`The file '${manifestPath}' is not valid`));
+    }
+  });
+}
+
+export default function stripBom(manifestPath: string) {
```
igor-ribeiiro, line 446: What does `Bom` mean? Why do we need this function?

packages/office-addin-dev-settings/test/files/manifests/manifest.json
```diff
@@ -1,121 +1,147 @@
...
+                }
+              ]
+            }
+          ]
+        }
+      ]
+    }
   }
-}
+
```
igor-ribeiiro, line 147: NIT: endl here

igor-ribeiiro (changes requested):

packages/office-addin-dev-settings/test/files/manifests/manifest.json
millerds: ??

millerds: ??

igor-ribeiiro: end line* here
I meant to say to add a enter/break/end line in the end of the file

packages/office-addin-dev-settings/src/sideload.ts
millerds: https://docs.microsoft.com/en-us/globalization/encoding/byte-order-mark

It was getting in the way of outlook loading the add-in because the resulting xml had the BOM.  utf-8 doesn't need it so I removed it.

packages/office-addin-dev-settings/src/sideload.ts
millerds: It's getting logged out in the debug console . . . I'm not sure it will be output when used in the template, but I think it would.

It would be nice to keep this aspect under the covers and hidden . . . however, if there is something wrong with the json that breaks the conversion process that's how well know what to report.  If we completely hid this we'll have to come up with some way of helping developers know to fix their json.

igor-ribeiiro (approved):

PR #632, opened 2022-04-27
Additional parameters in JSON metadata file

p0n3:
We should be able to set up [additional properties](https://docs.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-json#metadata-reference) (like `allowErrorForDataTypeAny`) in the JSON metadata file using `custom-functions-metadata-plugin`.
Soon we could also add `allowCustomDataForDataTypeAny` (the property is currently available in public preview).

### DIFF
--- a/packages/custom-functions-metadata-plugin/README.md
+++ b/packages/custom-functions-metadata-plugin/README.md
@@ -5,13 +5,18 @@ A WebPack plugin which generates the metadata for custom functions.
 
 # Webpack example (webpack.config.js)
 
+```
 const CustomFunctionsPlugin = require("custom-functions-metadata-plugin");
 
 plugins: [
       new CustomFunctionsPlugin({
         input: './src/functions/functions.ts',
-        output: 'functions.json'})
+        output: 'functions.json',
+        options: {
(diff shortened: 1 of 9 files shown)


akrantz:
Actually. I think it's not a return value but an input that denotes the error state. It would help to have an example. Rather than jsut allow type "any" -- really `unknown` now in TypeScript as `any` is discouraged -- it would be better to have a type defined for it, e.g. CustomFunctions.InputError with that type defined in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/custom-functions-runtime/index.d.ts.

millerds:
Looks like someone else is suggesting a different solution to the same problem using the meta data package the some more comments in the functions file rather than making them command options.

https://github.com/OfficeDev/Office-Addin-Scripts/pull/654

Seems like the general challenge that needs to be solved is that there are top level properties in the json metadata used by Office that aren't being generated by the tooling (office-addin-customfuctions-metadata and office-addin-customfunctions-metadata-plugin).

akrantz:
This is no longer relevant.

PR #633, opened 2022-04-27
Manipulating json

igor-ribeiiro:
Updated manifest manipulation code to test and get all fields.

### DIFF
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
@@ -46,7 +46,7 @@ export class ManifestHandlerJson extends ManifestHandler {
     manifestInfo.description = teamsAppManifest.description.short;
     manifestInfo.displayName = teamsAppManifest.name.short;
     manifestInfo.highResolutionIconUrl = teamsAppManifest.icons.color;
-    manifestInfo.hosts = teamsAppManifest.extension?.runtimes[0]?.requirements?.capabilities?.map((capability: any) => capability.name); // We need for the new schema to be published
+    manifestInfo.hosts = teamsAppManifest.extension?.requirements?.scopes; // We need for the new schema to be published
     manifestInfo.iconUrl = teamsAppManifest.icons.color;
     manifestInfo.officeAppType = teamsAppManifest.extension?.requirements?.capabilities[0]?.name; // We need for the new schema to be published
     manifestInfo.permissions = teamsAppManifest.authorization?.permissions?.resourceSpecific[0]?.name; // Still need to find an example where Permissions work. Or the manifest doesn't have it, or it is a wrong vlaue.
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
@@ -37,7 +37,6 @@ export class ManifestHandlerJson extends ManifestHandler {
     }
   }
 
-  // getManifestInfo(teamsAppManifest: TeamsAppManifest): ManifestInfo {
   getManifestInfo(teamsAppManifest: any): ManifestInfo {
     const manifestInfo: ManifestInfo = new ManifestInfo();
     manifestInfo.id = teamsAppManifest.id;
@@ -45,11 +44,11 @@ export class ManifestHandlerJson extends ManifestHandler {
     manifestInfo.defaultLocale = teamsAppManifest?.localizationInfo?.defaultLanguageTag;
     manifestInfo.description = teamsAppManifest?.description?.short;
     manifestInfo.displayName = teamsAppManifest?.name?.short;
-    manifestInfo.highResolutionIconUrl = teamsAppManifest?.icons.color;
-    manifestInfo.hosts = teamsAppManifest?.extension?.requirements?.scopes; // We need for the new schema to be published
+    manifestInfo.highResolutionIconUrl = teamsAppManifest?.icons?.color;
+    manifestInfo.hosts = teamsAppManifest?.extension?.requirements?.scopes;
     manifestInfo.iconUrl = teamsAppManifest?.icons?.color;
-    manifestInfo.officeAppType = teamsAppManifest?.extension?.requirements?.capabilities[0]?.name; // We need for the new schema to be published
-    manifestInfo.permissions = teamsAppManifest?.authorization?.permissions?.resourceSpecific[0]?.name; // Still need to find an example where Permissions work. Or the manifest doesn't have it, or it is a wrong vlaue.
+    manifestInfo.officeAppType = teamsAppManifest?.extension?.requirements?.capabilities[0]?.name;
+    manifestInfo.permissions = teamsAppManifest?.authorization?.permissions?.resourceSpecific[0]?.name;
     manifestInfo.providerName = teamsAppManifest?.developer?.name;
     manifestInfo.supportUrl = teamsAppManifest?.developer?.websiteUrl;
     manifestInfo.version = teamsAppManifest?.version;
--- a/packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
+++ b/packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
@@ -37,18 +37,19 @@ export class ManifestHandlerJson extends ManifestHandler {
     }
   }
 
-  getManifestInfo(teamsAppManifest: TeamsAppManifest): ManifestInfo {
+  // getManifestInfo(teamsAppManifest: TeamsAppManifest): ManifestInfo {
+  getManifestInfo(teamsAppManifest: any): ManifestInfo {
     const manifestInfo: ManifestInfo = new ManifestInfo();
     manifestInfo.id = teamsAppManifest.id;
-    manifestInfo.appDomains = teamsAppManifest.validDomains;
+    // manifestInfo.appDomains = teamsAppManifest.validDomains; // Don't know what this is
     manifestInfo.defaultLocale = teamsAppManifest.localizationInfo?.defaultLanguageTag;
     manifestInfo.description = teamsAppManifest.description.short;
     manifestInfo.displayName = teamsAppManifest.name.short;
     manifestInfo.highResolutionIconUrl = teamsAppManifest.icons.color;
-    // manifestInfo.hosts = teamsAppManifest.extension.requirements.scopes; // We need for the new schema to be published
+    manifestInfo.hosts = teamsAppManifest.extension?.runtimes[0]?.requirements?.capabilities?.map((capability: any) => capability.name); // We need for the new schema to be published
     manifestInfo.iconUrl = teamsAppManifest.icons.color;
-    // manifestInfo.officeAppType = teamsAppManifest.extension.requirements.capabilities.name; // We need for the new schema to be published
-    // manifestInfo.permissions = teamsAppManifest.Permissions; // Still need to find an example where Permissions work. Or the manifest doesn't have it, or it is a wrong vlaue.
+    manifestInfo.officeAppType = teamsAppManifest.extension?.requirements?.capabilities[0]?.name; // We need for the new schema to be published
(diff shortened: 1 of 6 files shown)


packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
```diff
@@ -37,21 +37,21 @@ export class ManifestHandlerJson extends ManifestHandler {
     }
   }
 
-  getManifestInfo(teamsAppManifest: TeamsAppManifest): ManifestInfo {
+  getManifestInfo(teamsAppManifest: any): ManifestInfo {
```
millerds, line 40: Why are we changing this to "any"?  Typescript discourages the use of any suggesting that "unknown" is better . . . but a strong type is even more desirable.

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,120 +1,82 @@
 {
   "$schema": "https://raw.githubusercontent.com/OfficeDev/microsoft-teams-app-schema/preview/DevPreview/MicrosoftTeams.schema.json",
-  "id": "df2fa15f-332a-47aa-87c2-e30c20686eee",
-  "manifestVersion": "m365DevPreview",
+  "id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
```
millerds, line 3: What's the reason for this id?

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,120 +1,82 @@
...
-  },
-  "accentColor": "#FFFFFF",
   "developer": {
     "name": "Contoso",
     "websiteUrl": "https://www.contoso.com",
     "privacyUrl": "https://www.contoso.com/privacy",
-    "termsOfUseUrl": "https://www.contoso.com/terms"
+    "termsOfUseUrl": "https://www.contoso.com/servicesagreement"
+  },
+  "icons": {
```
millerds, line 20: The conversion tool may require this to be a full url (i.e. using localhost:3000 as the domain)

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,120 +1,82 @@
...
-    "outline": "test/assets/icon-32.png",
-    "color": "test/assets/icon-32.png"
-  },
-  "accentColor": "#FFFFFF",
   "developer": {
     "name": "Contoso",
     "websiteUrl": "https://www.contoso.com",
     "privacyUrl": "https://www.contoso.com/privacy",
-    "termsOfUseUrl": "https://www.contoso.com/terms"
+    "termsOfUseUrl": "https://www.contoso.com/servicesagreement"
```
millerds, line 18: Can we add supportUrl here as well (https://www.contoso.com/help)?

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,120 +1,82 @@
...
+  },
+  "icons": {
+    "outline": "test/assets/icon-32.png",
+    "color": "test/assets/icon-32.png"
   },
+  "accentColor": "#230201",
   "localizationInfo": {
-    "defaultLanguageTag": "en-us"
+    "defaultLanguageTag": "en-us",
+    "additionalLanguages": [
```
millerds, line 27: I think we need to start with this as an empty array since we don't have language files as part of our template (or do we?).

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,120 +1,82 @@
...
+  "validDomains": [
+    "contoso.com"
+  ],
+  "extension": {
+      "requirements": {
+          "scopes": [
+            "mail"
+          ],
+          "capabilities": [
+             { "name": "AddinCommands", "minVersion": "1.1" }
```
millerds, line 53: Do we not need Mailbox?  Why are we Adding "AddinCommands"?

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,120 +1,82 @@
...
-        "actions": [
-          {
-            "id": "ExecuteFunction.action",
-            "type": "execution",
-            "name": "action"
-          }
-        ]
-      }
-    ],
-    "ribbons": [
```
millerds, line 67: So we have no definition of the ribbon button anymore?

millerds (changes requested):
Can you explain more of the reasoning behind the reduction of the runtime section and the removal of the ribbon secation?

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
igor-ribeiiro: This should be temporary. `TeamsAppManifest` is not updated to include some of the fields here.
So I put any while it is not update to get latest schema changes.

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: Can be any value. Just liked to be easy to test

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: What is the exact name of this property?
I didn't find anymore URLs other than the ones here, so I didn't find a `supportUrl` field

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: All the examples I saw used https URLs.
Might be just for examples, but we can change these if we found out it later.

packages/office-addin-manifest/test/manifests/manifest.json
millerds: "supportUrl": "https://www.contoso.com/help"

That's the one used in the adobe example in that repo and by the conversion tool.

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
millerds: You should leave a comment to that effect and log a bug to fix it.

packages/office-addin-manifest/test/manifests/manifest.json
millerds: As long as it's just the test value.

packages/office-addin-manifest/test/manifests/manifest.json
millerds: If the examples are using https URLs then we should to.

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: Do we? I'm not sure. As said previously, I am mostly basing myself on the Alpha example.
It doesn't define Mailbox here, it just on `runtime` later on.

packages/office-addin-manifest/src/manifestHandler/manifestHandlerJson.ts
igor-ribeiiro: Sure, added a TODO comment and created a bug to fix it

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: Apparently, this `supportUrl` is now in the schema, I was looking at a little older version. So I just added it

Also, as FYI:
From what I understand, we shouldn't base ourselves on the Adobe example. It is not up to date
The most up-to-date example is always the alpha example and that is what I'm basing myself on.
The schema should also be up to date.

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: I don't think these definitions are needed here, as this is just for testing for manifest manipulation.
We will need it in the task pane, as well some other definitions there, but they can get pretty lengthy and as we cannot test right now, I don't know which are necessary or not

packages/office-addin-manifest/test/manifests/manifest.json
millerds: Capabilities is an array of requirements . . . I wouldn't think the names of those are tied directly to the schema since it really depends on the host support vs. the add-in requirements and should be independent from the schema.

packages/office-addin-manifest/test/manifests/manifest.json
millerds: I would think we want the test manifest to match the taskpane manifest as much as possible in order to cover the same things.  There could be bugs in that wouldn't be caught by the test if they were a result of things in the taskpane manifest but not in the test manifest.

packages/office-addin-manifest/test/manifests/manifest.json
millerds: Use:

    "icons": {
      "outline": "https://localhost:3000/assets/icon-64.png",
      "color": "https://localhost:3000/assets/icon-128.png"
    },

millerds (changes requested):

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: Sure, I will add the "Mailbox" as a dependency

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: I added all the sections that I thought would be relevant for the outlook test now.

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,121 +1,550 @@
...
+                        {
+                            "id": "dashboard",
+                            "label": "Controls",
+                            "controls": [
+                                {
+                                    "id": "uppercase",
+                                    "type": "button",
+                                    "label": "To uppercase",
+                                    "icons": [
+                                        { "size": 16, "file": "test_16.png" },
```
millerds, line 342: I think the references need to be full localhost urls don't they?

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,121 +1,550 @@
...
+        }
+      ],
+      "runtimes": [
+        {
+          "requirements": {
+              "capabilities": [
+                { "name": "MailBox", "minVersion": "1.10" }
+              ]
+          },
+          "id": "text",
```
millerds, line 70: Can we use a better id?

millerds (changes requested):

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,121 +1,550 @@
...
+          "type": "general",
+          "code": {
+              "page": "https://localhost:3000/taskpane.html",
+              "script": "https://localhost:3000/taskpane.js"
+          },
+          "lifetime": "short",
+          "actions": {
```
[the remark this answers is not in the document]
igor-ribeiiro: I will put that capability back. I guess the `.json` manifest is indeed getting complex

packages/office-addin-manifest/test/manifests/manifest.json
igor-ribeiiro: Changed to "mail"

packages/office-addin-manifest/test/manifests/manifest.json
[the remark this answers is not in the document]
igor-ribeiiro: Added the runtime for it

packages/office-addin-manifest/test/manifests/manifest.json
```diff
@@ -1,121 +1,550 @@
...
+        {
+            "requirements": {
+              "formFactors": ["desktop"]
+            },
+            "title": "Get Started",
+            "description": "Your sample add-in loaded succesfully. Click the buttons to get started.",
+            "learnMoreUrl": "https://www.contoso.com/learnmore"
+        }
+      ],
+      "runtimes": [
```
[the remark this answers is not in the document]
igor-ribeiiro: Is this the same as the showtaskpane runtime? 
I just added that option

millerds (approved):

PR #646, opened 2022-06-03
[Patch] yo Office is now responsible for data writting

igor-ribeiiro:
Usage data was wrongfully always setting usage-data for yo-office to "no" and not asking the client what they prefer.
Now, the responsibility of writing the usage-data level is on Generator Office and it will not be a special case in `office-addin-usage-data`.

    No

    No

Tested and validated that Yo-Office now asks and raises data when accepted by the customer. Yo Office now asks if it can collect data and it doesn't assume it is a no.

### DIFF
--- a/packages/office-addin-usage-data/src/usageData.ts
+++ b/packages/office-addin-usage-data/src/usageData.ts
@@ -99,16 +99,6 @@ export class OfficeAddinUsageData {
         this.options.usageDataLevel = jsonData.readUsageDataLevel(this.options.groupName);
       }
 
-      // Generator-office will not raise a prompt because the yeoman generator creates the prompt.  If the projectName
-      // is defaults.generatorOffice and a office-addin-usage-data file hasn't been written yet, write one out.
-      if (
-        this.options.projectName === defaults.generatorOffice &&
-        this.options.instrumentationKey === defaults.instrumentationKeyForOfficeAddinCLITools &&
-        jsonData.needToPromptForUsageData(this.options.groupName)
-      ) {
-        jsonData.writeUsageDataJsonData(this.options.groupName, this.options.usageDataLevel);
-      }
-
       if (
         !this.options.isForTesting &&
         this.options.raisePrompt &&


akrantz:
What if you use an office-addin-scripts package via NPX without using Yo Office? I think previously it worked by asking the question if there was not a setting yes or no. If you also removed the saved setting, it should prompt again, and this also doesn't involve Yo Office.

akrantz:
I think I was misled by the PR title. It seems this only removes code from Yo Office to write anything.  Doesn't that mean that Yo Office is NO LONGER responsible for writing data?

igor-ribeiiro:
> I think I was misled by the PR title. It seems this only removes code from Yo Office to write anything. Doesn't that mean that Yo Office is NO LONGER responsible for writing data?

Yeah, this change alone doesn't really solve the problem. I have a PR ready for Generator Office to pretty much call that  `writeUsageDataJsonData` function and that will solve the problem.

So this PR is more to remove the responsibility of `office-addin-usage-data`, so I can publish the package and use it in the generator office. Then I will make the change for the generator to be responsible for this function call.

akrantz:
But you don't want to do that. It SHOULD be the responsiblity of office-addin-usage-data.

igor-ribeiiro:
This is the PR on the generator office: https://github.com/OfficeDev/generator-office/pull/709 
It will only add the reference to the published package once available

igor-ribeiiro:
> My recollection was that when yo office is run, there is a prompt which comes up, and the user choice can be saved. Is the bug just that the user choice isn't being saved correctly and was always no?

No, the bug was that the question was never asked and the default response would be no.

igor-ribeiiro:
> 

Just to give some more context on the reason our data collection was bugged. This is the line that broke the bug collection:
https://github.com/OfficeDev/generator-office/pull/693/files#diff-e234283aa9cd1cdf0803a13194964671ae1dfd31064f5fe35429df36a51ef17bR117
It was instead here: https://github.com/OfficeDev/generator-office/pull/693/files#diff-e234283aa9cd1cdf0803a13194964671ae1dfd31064f5fe35429df36a51ef17bL221

This difference is where the constructor was placed was what created the bug. This is because the constructor has the `writeUsageDataJsonData` call in it. 
- If the constructor is before the yo questions are asked (how it is now), this will create a blank `usageData` file that defaults the option of data collection to `no` without asking anything. Yo Office sees that this file exists and proceeds to not collect data.
- If the constructor is after the questions (before the bug), yo Office will make a question if you accept data-collection or not (as the cache usageData file does not exist). Then, it will `writeUsageDataJsonData` with the user's answer.

Addressing the comments:
- The generator office prompt itself wasn't changed in that other PR so it should be the same as is.
The prompt didn't change. It was just not getting triggered because of the scenarios I described above, where the `usageData` file would already be written before asking any question.

- What does moving the call to "writeUsageDataJsonData" have on anything (I'm not sure what this call does . . . just sounds like it writes data rather than choose something).
It is the "breaking" function in the `usageData` constructor that caused the bug. It was only called in the special case for yo office and was kind of confusing to be there. So I changed its place to generator office code (which is the other PR: https://github.com/OfficeDev/generator-office/pull/709)

- What does the comment "Generator-office will not raise a prompt because the yeoman generator creates the prompt." mean?
`office-addin-usage-data` has code to generate the prompt about data usage. Yo Office is treated as a special case because `yo` can already create prompts and we instead chose to use `yo` prompts instead of the defaults of `office-addin-usage-data`.

millerds:
Thanks for the clarification.  If moving the constructor broke of the usage data object created the bug, why aren't we just moving it back to where it was?

millerds:
So previously we wouldn't be creating the file in generator-office if the prompt would answer no, but then then office-addin-usage-data would create the file anyway with some default settings.  After the constructor was moved the file gets created by generator-office all the time, but always with default settings.  So this change is to get rid of the creation of the file by office-addin-usage-data on behalf of generator-office (which shouldn't be needed anymore) and the other PR will update the file with non default settings when the prompt get's answered?

Are we sure there weren't any other side effects that are now broken because the constructor was moved?

igor-ribeiiro:
- So previously we wouldn't be creating the file in generator-office if the prompt would answer no, but then then office-addin-usage-data would create the file anyway with some default settings. After the constructor was moved the file gets created by generator-office all the time, but always with default settings. So this change is to get rid of the creation of the file by office-addin-usage-data on behalf of generator-office (which shouldn't be needed anymore) and the other PR will update the file with non default settings when the prompt get's answered?

Yeah, that all correct

- Are we sure there weren't any other side effects that are now broken because the constructor was moved?
I can say that I don't think there will be any side effects from non-generator office projects because there is an `if` for `this.options.projectName === defaults.generatorOffice`.
I also tested those 2 PRs together and the telemetry and file creation were correct. I also investigated both codes and, to my knowledge, this would fix this issue and others wouldn't arise from it.

millerds:
> * Are we sure there weren't any other side effects that are now broken because the constructor was moved?
>   I can say that I don't think there will be any side effects from non-generator office projects because there is an `if` for `this.options.projectName === defaults.generatorOffice`.
>   I also tested those 2 PRs together and the telemetry and file creation were correct. I also investigated both codes and, to my knowledge, this would fix this issue and others wouldn't arise from it.

Sorry . . . I was referring to the generator-office change done earlier and not these two.  If the constructor was done after all prompting before, then may the write you are adding now should be done after everything as well.  This would cover any other potential changes the prompting triggers . . . but then again if something happened in the other prompts that causes us to bail out then any updates wouldn't happen to the file.  I just want to be sure that if we're fixing this, we aren't just fixing one spot while others may be there as well.

igor-ribeiiro:
> are adding now should be done after everything as well

I think you mean this other PR: https://github.com/OfficeDev/generator-office/pull/709/files#diff-e234283aa9cd1cdf0803a13194964671ae1dfd31064f5fe35429df36a51ef17bR138, right?

If so, I think the current place makes sense. We right the use option right after they choose it. If they cancel Yo Office process in the middle, the option is already salved,
This seems like reasonable behavior to me. In the prompt, we already tell the user how to change the option manually, so they can change after if they want.

millerds:
> > are adding now should be done after everything as well
> 
> I think you mean this other PR: https://github.com/OfficeDev/generator-office/pull/709/files#diff-e234283aa9cd1cdf0803a13194964671ae1dfd31064f5fe35429df36a51ef17bR138, right?
> 
> If so, I think the current place makes sense. We right the use option right after they choose it. If they cancel Yo Office process in the middle, the option is already salved, This seems like reasonable behavior to me. In the prompt, we already tell the user how to change the option manually, so they can change after if they want.

Yes . . . I agree writing the option after the choice is made make sense and avoid the repeat of the question.  I just worry that there are other things that should be written (that were before the constructor moved and aren't now).  And what about the case where the constructor was executed, the prompt not yet answered, and the process stopped (error or yo office cancelled) . . . then we end up with a file using just the defaults as we do now.

PR #652, opened 2022-06-14
[patch] Improve dev server startup

millerds:

---

**Change Description**:
When 'npm start' is run we check if the required port is already being used (bound to) and assume the dev server is already running if it is.  However it turns out that when an add-in is run in the host a whole bunch of port bindings to the same port are made and when the host is closed some of those bindings linger attached to pid 0 (the idle process) before finally getting cleaned up.  This has the effect of blocking 'npm start' while these bindings are waiting to be cleaned up.  It seems safe to ignore these bindings on pid 0 and allow 'npm start' to re-setup the dev server without having to wait so long.

    No.

    No.

    Ran test suit.

### DIFF
--- a/packages/office-addin-debugging/src/port.ts
+++ b/packages/office-addin-debugging/src/port.ts
@@ -54,7 +54,7 @@ export function getProcessIdsForPort(port: number): Promise<number[]> {
   return new Promise((resolve, reject) => {
     const isWin32 = process.platform === "win32";
     const isLinux = process.platform === "linux";
-    const command = isWin32 ? `netstat -ano` : isLinux ? `netstat -tlpna | grep :${port}` : `lsof -n -i:${port}`;
+    const command = isWin32 ? `netstat -ano | findstr :${port}` : isLinux ? `netstat -tlpna | grep :${port}` : `lsof -n -i:${port}`;
 
     childProcess.exec(command, (error, stdout) => {
       if (error) {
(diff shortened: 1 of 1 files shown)


packages/office-addin-debugging/src/port.ts
```diff
@@ -54,7 +54,11 @@ export function getProcessIdsForPort(port: number): Promise<number[]> {
   return new Promise((resolve, reject) => {
     const isWin32 = process.platform === "win32";
     const isLinux = process.platform === "linux";
-    const command = isWin32 ? `netstat -ano` : isLinux ? `netstat -tlpna | grep :${port}` : `lsof -n -i:${port}`;
+    const command = isWin32
+      ? `netstat -ano | findstr :${port}`
```
igor-ribeiiro, line 58: do we need this `findstr` statement?

millerds: This brings Win32 in line with the other platforms which filter the list down before looping over all the port bindings.  It's just an optimization and not critical.

packages/office-addin-debugging/src/port.ts
```diff
@@ -71,7 +75,7 @@ export function getProcessIdsForPort(port: number): Promise<number[]> {
           lines.forEach((line) => {
             /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
             const [protocol, localAddress, foreignAddress, status, processId] = line.split(" ").filter((text) => text);
-            if (processId !== undefined) {
+            if (processId !== undefined && processId !== '0') {
```
igor-ribeiiro, line 78: My guess is that this `0` only works for Windows right?
Could we add also test on other platforms if possible and, if it doesn't work on them, can we add some `if` for checking if we are on Windows.

igor-ribeiiro (changes requested):

packages/office-addin-debugging/src/port.ts
igor-ribeiiro: Can we then make this `0` option only be checked for Win32?
We don't want this to accidentally create some bug on other platforms now or in the future.

millerds: It already is (if statement starts on line 74)

igor-ribeiiro (approved):

PR #660, opened 2022-07-11
[Minor] Convertion tool

igor-ribeiiro:
Adding beginning support for converting XML to JSON manifests. Currently, it is a draft PR because the tool used is not 100% ready.

Yes, adding one more option `output` to the convert command.

Not yet, as the documentation for these commands are yet to be written.

Manual runs of the convert tool and `npm run test` while the tool is in the root directory.


packages/office-addin-project/src/convert.ts
```diff
@@ -68,6 +70,10 @@ async function backupProject(backupPath: string) {
   });
 }
 
+export function convertManifest(manifestPath: string, outputPath: string) {
```
millerds, line 73: Does this need to be exported?

millerds (changes requested):
We shouldn't be making a PR directly against master . . . it should be in a branch (even for draft cases).

packages/office-addin-project/src/convert.ts
```diff
@@ -68,6 +70,10 @@ async function backupProject(backupPath: string) {
...
+  execSync(`node ./converter/lib/main.js --input ${manifestPath} --output ${outputPath} --verbose false --debug true`);
```
millerds, line 74: I don't see this converter code.

akrantz: The conversion code should be in a separate package which is a dependency, and then "npx <package name>" can invoke the CLI for the package.  

You can also have it run an NPM script, which allows the script to have different commands if needed. An example of this is "npx run dev-server" where the "dev-server" script does the appropriate thing, but it isn't tightly coupled.

akrantz: Even when being developed locally, you can use "npm link" to have the package as a dependency.

millerds:
But there is no reference in the PR to a package to create a link for.  That's my question . . . where is "./converter/lib/main.js" coming from?  It's not in the PR itself and the PR doesn't contain and reference to a package that suggest the manifest converter we are trying to integrate.  So if I were to clone and run this then I don't think it would work because I'm missing something not included here.

igor-ribeiiro:
You both @akrantz and @millerds are correct, this PR itself does not work right now. I am using some local code on my machine to be able to use the converter calls, that's why I opened as a draft PR, sorry for not being more clear that it is not working right now.

To have the final PR version, I am just waiting to receive the package version of the local files I'm using. As suggested, I would change the CLI call to use `npx package_name`.

millerds:
> You both @akrantz and @millerds are correct, this PR itself does not work right now. I am using some local code on my machine to be able to use the converter calls, that's why I opened as a draft PR, sorry for not being more clear that it is not working right now.
> 
> To have the final PR version, I am just waiting to receive the package version of the local files I'm using. As suggested, I would change the CLI call to use `npx package_name`.

I'm not sure what suggestion relates to "npx package_name" . . . I was suggesting we don't want a cli command for converting a manifest and that is should just be the project convert.

igor-ribeiiro:
> > You both @akrantz and @millerds are correct, this PR itself does not work right now. I am using some local code on my machine to be able to use the converter calls, that's why I opened as a draft PR, sorry for not being more clear that it is not working right now.
> > To have the final PR version, I am just waiting to receive the package version of the local files I'm using. As suggested, I would change the CLI call to use `npx package_name`.
> 
> I'm not sure what suggestion relates to "npx package_name" . . . I was suggesting we don't want a cli command for converting a manifest and that is should just be the project convert.

Agree, ideally we have some APIs to call instead of using the CLI. What I meant is that there is not NPM package for us to import yet. When we have it, we can work on making this PR work.

packages/office-addin-project/README.md
```diff
@@ -20,6 +20,7 @@ Syntax:
 
 `manifest-path`: path to manifest file.
 `backup-path`: path to the generated backup of the project.
+`output-directory`: path to the output folder for the converted manifest.
```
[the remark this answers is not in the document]
igor-ribeiiro: That is just an option, there is a default directory `./manifest` if nothing is chosen. Do you think it is better to not even have the option?

packages/office-addin-project/src/convert.ts
igor-ribeiiro: The idea is the customer could call this function directly and not use the backup and `package.json` check functions

packages/office-addin-project/src/convert.ts
```diff
@@ -68,6 +71,10 @@ async function backupProject(backupPath: string) {
   });
 }
 
+export function convertManifest(manifestPath: string, outputPath: string) {
+  convert(manifestPath, outputPath);
```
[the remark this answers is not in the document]
igor-ribeiiro: Just to be exported to clients

packages/office-addin-project/README.md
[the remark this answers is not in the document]
millerds: I don't think we want any new option as it can get confusing and make us responsible for making sure everythign works with the new manifest location (like our npm scripts).  We should have a standard way of updating the project and if customer want to re-organize their project after the fact they can do it.

packages/office-addin-project/src/convert.ts
millerds: When would that be necessary?  The more pieces of the puzzle we expose the more we have to support customers usage of them and that limits how we change things in the future.  The goal is to get customer project from a xml based manifest project to a json base manifest project and we don't need to allow all the intermediate steps to be done separately.

Even if we did want to do this (and I'm saying we don't) this would go in the manifest package and not the project package.

packages/office-addin-project/src/convert.ts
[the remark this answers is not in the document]
millerds: The convert method itself is already being exported from the package we are consuming it from . . . which is also ours.  We don't need to create a new entry point to the same thing and then have to maintain its exposure.  The goal here is to convert the project, not to provide the breakdown of step to convert the project.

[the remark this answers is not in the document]
igor-ribeiiro: Sure, I removed the method then

igor-ribeiiro:
@millerds This is a non-erroring version. I think we can move with the PR even if still have some bugs so it becomes easier for people to test.

packages/office-addin-project/README.md
```diff
@@ -6,13 +6,15 @@ For more information, see the [documentation](
...
 ## Command-Line Interface
+
 * [convert](#info)
 
 #
 
 ### convert
 
 Converts the Office Addin Code from xml to json based manifest.
+The outputed conversion is avaible at `./manifest`.
```
millerds, line 17: This is converting the project and not just the manifest so this could also be misleading.

packages/office-addin-project/src/convert.ts
```diff
@@ -15,6 +16,7 @@ export async function convertProject(
   manifestPath: string = "./manifest.xml",
   backupPath: string = "./backup.zip"
 ) {
+  const outputPath: string = "./manifest";
```
millerds, line 19: We should convert to the same location as the xml manifest . . . which may not be "./manifest".  (Note that the only reason the json template had a manifest folder was because of a quirk in the json => xml converter)

millerds (changes requested):

millerds:
Note that we can't check this in until after private preview.

packages/office-addin-project/README.md
igor-ribeiiro: Agree, the documentation could have the details explaining the resulting project structure (like where is the manifest).

packages/office-addin-project/src/convert.ts
igor-ribeiiro: But what about the fact that the JSON manifest could have multiple files?
Like it can be split up into multiple files (for localization is the easiest example). It does seem nice to have it already in a specific folder where all those files can live.

millerds: Maybe, but we don't want to enforce a project structure if we can avoid it.  We would also have to change all the references to the manifest location built into the project (like in the package.json) which may not be a big deal, but it still extra work.

It might be nice to put the language specific file in a subdirectory, but that might require updating the manifest references to the lang files and I'm not sure it's required.  I think it's best to leave it to the developer to move things around in the project.  If we receive feedback to the contrary, we can add this later.

millerds: The outputPath needs to be the location of the input manifest path.

millerds (changes requested):

igor-ribeiiro:
@millerds just made the output in the same folder as the input manifest.
This PR is now functional and working. Can we merge it? I think we should merge it and hold on to publishing.
Then we will see how to publish/release to the people that need it.

millerds:
I checked and we are still using versions <1.0.0 which basically means pre-release so I guess we should be OK checking this in and even publishing it as long as we support it like pre-release versions.

millerds (approved):

PR #666, opened 2022-08-15
[Major] Move access token retrieval server side

millerds:

---

**Change Description**:
Move graph token retrieval from the code running in the client (taskpane) to the code running on the service (middle-tier).  Remove the get auth token call so the client can't get it.  Adjust the other supporting flows to handle errors.

No.

Maybe.  The behavior changes to the script package such that places consuming it might need to adjust their description of if's functionality.

Using local adjustments to the template that consumes this I verified two different identities were still able to get the data requested.


packages/office-addin-sso/src/middle-tier/app.ts
```diff
@@ -76,19 +74,32 @@ export class App {
     });
 
     this.appInstance.get("/getuserdata", async function (req: any, res: any, next: any) {
```
davidchesnut, line 76: The access token passed in from the client must be validated here.

packages/office-addin-sso/src/middle-tier/app.ts
```diff
@@ -76,19 +74,32 @@ export class App {
...
-
-      // If Microsoft Graph returns an error, such as invalid or expired token,
-      // there will be a code property in the returned object set to a HTTP status (e.g. 401).
-      // Relay it to the client. It will caught in the fail callback of `makeGraphApiCall`.
-      if (graphData.code) {
-        next(createError(graphData.code, "Microsoft Graph error " + JSON.stringify(graphData)));
+      const graphTokenResponse = await getAccessToken(req.get("Authorization"));
+      if (graphTokenResponse.claims || graphTokenResponse.error) {
+        graphTokenResponse.claims
+          ? usageDataObject.reportEvent("CliamsResponse")
```
davidchesnut, line 80:
```suggestion
          ? usageDataObject.reportEvent("ClaimsResponse")
```

packages/office-addin-sso/src/middle-tier/app.ts
```diff
@@ -76,19 +74,32 @@ export class App {
...
+          : usageDataObject.reportError(
+              "AccessTokenError",
+              new Error("Access Token Error: " + graphTokenResponse.error)
+            );
+        res.send(graphTokenResponse);
```
davidchesnut, line 85: Seems like this should set an HTTP 400 or 401 based on which error.

packages/office-addin-sso/src/middle-tier/app.ts
```diff
@@ -76,19 +74,32 @@ export class App {
...
+        const graphQueryParamSegment: string = process.env.QUERY_PARAM_SEGMENT || "";
+
+        const graphData = await getGraphData(graphToken, graphUrlSegment, graphQueryParamSegment);
+
+        // If Microsoft Graph returns an error, such as invalid or expired token,
+        // there will be a code property in the returned object set to a HTTP status (e.g. 401).
+        // Relay it to the client. It will caught in the fail callback of `makeGraphApiCall`.
+        if (graphData.code) {
+          usageDataObject.reportException("getuserdata()", graphData.code);
+          next(createError(graphData.code, "Microsoft Graph error " + JSON.stringify(graphData)));
```
davidchesnut, line 98: Does this set 400 or 401 error in the response back to the client?

packages/office-addin-sso/src/middle-tier/ssoauth-helper.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import * as fetch from "node-fetch";
+import * as form from "form-urlencoded";
+
+/* global process */
+
+export async function getAccessToken(authorization: string): Promise<any> {
+  const scopeName: string = process.env.SCOPE || "User.Read";
+  if (authorization == null) {
```
davidchesnut, line 14:
```suggestion
  if (authorization === null) {
```

packages/office-addin-sso/src/server.ts
```diff
@@ -5,7 +5,7 @@
 import * as https from "https";
 import * as devCerts from "office-addin-dev-certs";
 import { OfficeAddinManifest } from "office-addin-manifest";
-import { App } from "./app";
+import { App } from "./middle-tier/app";
```
davidchesnut, line 8: Just FYI after discussing with Ricky I am moving the term from "middle-tier" back to server, web application, or similar. I'll be updating docs appropriately (in case you are trying to map to docs.)

packages/office-addin-sso/src/msgraph-helper.ts
```diff
@@ -3,46 +3,15 @@
...
+export async function makeGraphApiCall(bootstrapToken: string): Promise<any> {
   try {
     const response = await $.ajax({
       type: "GET",
       url: "/getuserdata",
-      headers: { access_token: accessToken },
+      headers: { Authorization: "Bearer " + bootstrapToken },
       cache: false,
     });
     return response;
```
davidchesnut, line 17: Here too I am wondering if the server is returning 400 or 401 on errors to force this try/catch to happen.

davidchesnut (changes requested):
Looks good overall! A few suggestions to please take a look at. Thanks!

Rick-Kirkham (dismissed):

igor-ribeiiro (approved):

packages/office-addin-sso/src/middle-tier/app.ts
```diff
@@ -76,19 +73,24 @@ export class App {
...
-      const graphQueryParamSegment = process.env.QUERY_PARAM_SEGMENT || "";
-
-      const graphData = await getGraphData(graphToken, graphUrlSegment, graphQueryParamSegment);
-
-      // If Microsoft Graph returns an error, such as invalid or expired token,
-      // there will be a code property in the returned object set to a HTTP status (e.g. 401).
-      // Relay it to the client. It will caught in the fail callback of `makeGraphApiCall`.
-      if (graphData.code) {
-        next(createError(graphData.code, "Microsoft Graph error " + JSON.stringify(graphData)));
+      const graphTokenResponse = await getAccessToken(req.get("Authorization"));
```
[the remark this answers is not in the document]
millerds: Added

packages/office-addin-sso/src/middle-tier/app.ts
davidchesnut: Sorry I meant 403 (which is what Graph will return) although probably unlikely since you know the URL will always be correct.

millerds: This a little bit fallout from the access token fetching being on the client before.  I moved it to the middle-tier, but handling the claims and errors that require client action (alternate bootstrap token retrieval) is still in the client and still based on the response.  I'd have to pick apart the response to create a new one with enough information for the client to do what it needs.

packages/office-addin-sso/src/middle-tier/app.ts
millerds: I will use the same error code that graph indicates . . . if that is 400 or 401 then that is what will be reported to the client.

packages/office-addin-sso/src/msgraph-helper.ts
millerds: I don't think a 400 or 401 causes an exception . . . it should just report the response to the method caller and they should handle the response.  This is how the taskpane code will handle the request for claims as well.

davidchesnut: cool

packages/office-addin-sso/src/middle-tier/app.ts
davidchesnut: sound good.

packages/office-addin-sso/src/middle-tier/app.ts
millerds: I believe I've added the validation correctly.

davidchesnut: My bad, I see it now. thanks!

davidchesnut (approved):
Looks great!

packages/office-addin-sso/src/middle-tier/app.ts
millerds: I did a little digging to understand how this compared to you sample and this essentially where you wrap the response in a "500" error and the client pulls apart the interior info in order to retry if needed.  The difference here is I'm not wrapping the error and my client handling is part of the sso package and not in the taskpane project code.

I think it could be organized better here . . . but I will leave that for a future change.

PR #669, opened 2022-08-22
[minor] Reduce string_decoder dependencies

millerds:

---

**Change Description**:
Component governace is reporting that we shouldn't have dependencies on string_decoder@1.1.1.  There are a few packages we use that have nested dependencies that include string_decoder.  Most of them we can switch away from and use other packages with similar functionality.  In one case I had to override the nested dependency by adding a root dependency to the newer versions.  In the case of lerna, we don't have an alternative option so that is the one remaining case we'll take the low risk for.

I also had to update the typescript and ts-node dependencies to fix some test execution errors triggered by some dependency changes previously made.

No.

No.

Ran build and test multiple time to look for breaks.  Also connected to the taskpane project locally and tested running the add-in locally.

### DIFF
--- a/packages/office-addin-dev-settings/.vscode/launch.json
+++ b/packages/office-addin-dev-settings/.vscode/launch.json
@@ -142,7 +142,7 @@
       "type": "node",
       "request": "launch",
       "program": "${workspaceFolder}/lib/cli.js",
-      "args": ["sideload", "c:\\dev\\git\\Office-Addin-TaskPane\\manifest.excel.xml", "web", "-d", "https://1drv.ms/x/s!Au78PotIZz2Q5OEmJ-xvfBs3ODNUYg?e=F64J5i"]
+      "args": ["sideload", ".\\test\\files\\manifests\\manifest.xml"]
     },
     {
       "name": "Command Line: source-bundle-url",
--- a/packages/office-addin-manifest/test/test.ts
+++ b/packages/office-addin-manifest/test/test.ts
@@ -489,11 +489,11 @@ describe("Unit Tests", function() {
         assert.strictEqual(info.defaultLocale, "en-us");
         assert.strictEqual(info.description, "A template to get started.");
         assert.strictEqual(info.displayName, "Contoso Task Pane Add-in");
-        assert.strictEqual(info.highResolutionIconUrl, "https://localhost:3000/assets/icon-128.png");
+        assert.strictEqual(info.highResolutionIconUrl, "test/assets/icon-128.png", "highResolutionIconUrl");
         assert.strictEqual(info.hosts instanceof Array, true);
         assert.strictEqual(info.hosts!.length, 1);
         assert.strictEqual(info.hosts![0], "mail");
-        assert.strictEqual(info.iconUrl, "https://localhost:3000/assets/icon-128.png");
+        assert.strictEqual(info.iconUrl, "test/assets/icon-128.png", "iconUrl");
         assert.strictEqual(info.officeAppType, "AddinCommands");
         assert.strictEqual(info.permissions, "Mailbox.ReadWrite");
         assert.strictEqual(info.providerName, "Contoso");
--- a/packages/office-addin-project/src/convert.ts
+++ b/packages/office-addin-project/src/convert.ts
@@ -12,6 +12,8 @@ import { ExpectedError } from "office-addin-usage-data";
 
 /* global console */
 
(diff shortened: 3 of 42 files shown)


packages/office-addin-dev-settings/.vscode/launch.json
```diff
@@ -142,7 +142,7 @@
       "type": "node",
       "request": "launch",
       "program": "${workspaceFolder}/lib/cli.js",
-      "args": ["sideload", "c:\\dev\\git\\Office-Addin-TaskPane\\manifest.excel.xml", "web", "-d", "https://1drv.ms/x/s!Au78PotIZz2Q5OEmJ-xvfBs3ODNUYg?e=F64J5i"]
```
igor-ribeiiro, line 145: What is this change?

packages/office-addin-manifest/test/test.ts
```diff
@@ -489,11 +489,11 @@ describe("Unit Tests", function() {
         assert.strictEqual(info.defaultLocale, "en-us");
         assert.strictEqual(info.description, "A template to get started.");
         assert.strictEqual(info.displayName, "Contoso Task Pane Add-in");
-        assert.strictEqual(info.highResolutionIconUrl, "https://localhost:3000/assets/icon-128.png");
```
igor-ribeiiro, line 492: What made this necessary?

packages/office-addin-project/src/convert.ts
```diff
@@ -47,25 +48,25 @@ async function asksForUserConfirmation(): Promise<boolean> {
...
 
-    stream.on("close", () => {
-      console.log(
-        `A backup of your project was created to ${path.resolve(backupPath)}`
-      );
-      resolve();
-    });
-    archive.finalize();
+  const files: string[] = fs.readdirSync(__dirname);
+  files.forEach((entry) => {
```
igor-ribeiiro, line 55: Does this work for multiple nested directories? I feel like this needed to be somehow recursive on the directories right?

igor-ribeiiro (changes requested):

packages/office-addin-dev-settings/.vscode/launch.json
millerds: This changes the debug launching for office-addin-dev-settings to a manifest that is part of the project rather than assuming another location on the dev box outside of the current project.

packages/office-addin-manifest/test/test.ts
millerds: I updated the test manifest to using local references to icons to validate their inclusion in the zip file for json manifest add-ins.  This then validates those paths rather than the localhost paths.  I also added the extra parameter to distinguish between this and the other entry with the same value.

packages/office-addin-project/src/convert.ts
millerds: The adm-zip module automatically does nested folders when you call "addLocalFolder".  The reason for the loop is because I don't want every folder . . . this this reminds me that I forgot to add the code that skips some folders (i.e. node_modules)

packages/office-addin-project/src/convert.ts
```diff
@@ -51,15 +51,18 @@ async function asksForUserConfirmation(): Promise<boolean> {
...
   files.forEach((entry) => {
-    const entryStats = fs.lstatSync(entry);
+    const fullPath = path.resolve(entry)
+    const entryStats = fs.lstatSync(fullPath);
+
     if (skipBackup.includes(entry)) {
       // Don't add it to the backup
     } else if (entryStats.isDirectory()) {
-      zip.addLocalFolder(entry);
+      zip.addLocalFolder(entry, entry);
```
igor-ribeiiro, line 65: Can you put a comment on that to explain the need for two entries? (What each entry is doing).

packages/office-addin-project/src/convert.ts
millerds: Looks like I didn't make it quite right in general.  I have now tested it better and the updated code should work.

packages/office-addin-project/src/convert.ts
millerds: That would be part of the adm-zip documentation.  The first argument is the path to pull the file from (in this case relative to the project), and the second argument is the path in the zip file to put it.  I don't know that we want to get into the habit of redocumenting library apis locally . . . it could get out of date and we'd have tons of extra comments all over out code.

igor-ribeiiro (approved):

PR #670, opened 2022-08-24
[patch] Remove sso .gitignore file

millerds:

---

**Change Description**:
Remove the .gitignore file from office-addin-sso which is preventing the build output from being published.

No.

No.

Ran tests.

### DIFF
--- a/packages/office-addin-sso/.gitignore
+++ b/packages/office-addin-sso/.gitignore
@@ -1,5 +1,0 @@
-# Build output
-lib/
-
-# Node modules folder
-node_modules/


igor-ribeiiro:
Why is the building complaining about the `.gitignore` file?

millerds:
> Why is the building complaining about the `.gitignore` file?

Building doesn't complain for office-addin-sso . . . that works fine.  However, because of the .gitignore npm publish doesn't publish the built lib folder in the tgz file.  And so with the add in project consumes it there isn't the js module files (form the lib folder) and so it fails to find the office-addin-sso module.

igor-ribeiiro:
> > Why is the building complaining about the `.gitignore` file?
> 
> Building doesn't complain for office-addin-sso . . . that works fine. However, because of the .gitignore npm publish doesn't publish the built lib folder in the tgz file. And so with the add in project consumes it there isn't the js module files (form the lib folder) and so it fails to find the office-addin-sso module.

Can we just remove the lib folder line from the `.gitignore`? Then, it should stop ignoring the lib folder.

millerds:
> > > Why is the building complaining about the `.gitignore` file?
> > 
> > 
> > Building doesn't complain for office-addin-sso . . . that works fine. However, because of the .gitignore npm publish doesn't publish the built lib folder in the tgz file. And so with the add in project consumes it there isn't the js module files (form the lib folder) and so it fails to find the office-addin-sso module.
> 
> Can we just remove the lib folder line from the `.gitignore`? Then, it should stop ignoring the lib folder.

I added the git ignore in order to change VS Code behavior that that is was broke things.  npm and VS Code ignore the node_modules folder be default so there is no need to have that as part of a .gitignore and all the other packages in this repo don't use a .gitignore (and now I know why).

igor-ribeiiro (approved):

PR #675, opened 2022-09-06
[Patch] Adding bin executable to office-addin-project

igor-ribeiiro:

---

**Change Description**:

Added executable to the package for it to able to be run in other packages (with `npx` for example).
Added a specific version for the teams-fx package to be installed.

Tested conversion on Office-Addin-Manifest and checked that the conversion could be run and the updates were as expected.


packages/office-addin-project/src/convert.ts
```diff
@@ -91,7 +91,11 @@ function updatePackages(): void {
 
   for (let i = 0; i < depedentPackages.length; i++) {
     const depedentPackage = depedentPackages[i];
-    command += ` ${depedentPackage}@latest`;
+    if (depedentPackage === "@microsoft/teamsfx-cli") {
```
millerds, line 94: Currently, teamsfx should be a nested dependency and not a direct dependency so this shouldn't be necessary and the teamsfx entry in the array should be removed.

packages/office-addin-project/package.json
```diff
@@ -11,6 +11,9 @@
     "prettier": "office-addin-lint prettier",
     "test": "mocha -r ts-node/register test/**/*.ts"
   },
+  "bin": {
+    "office-addin-project": "lib/cli.js"
```
millerds, line 15: According to comments in other packages, a cli.js should be in the root that references the build js in the lib folder to be loaded dynamically.  This avoids problems when the lib folder doesn't exist.

millerds (changes requested):

packages/office-addin-project/src/convert.ts
igor-ribeiiro: We still need the dependency on @microsoft/teamsfx-cli right? I just updated to require it

millerds: The project has a dependency on office-addin-debugging which depends on office-addin-dev-settings which depends of teamsfx-cli.  I don't think we need to require teamsfx-cli at the project level package.json directly.

akrantz: You should specify the dependencies based on the direct package that the code uses. If it DIRECTLY gets code from teamsfx-cli, even if another dependent package, always does, you should still have a dependency on it.

packages/office-addin-project/package.json
```diff
@@ -19,9 +22,9 @@
     "commander": "^6.2.1",
     "fs-extra": "^7.0.1",
     "inquirer": "^7.3.3",
-    "office-addin-manifest": "^1.10.1",
-    "office-addin-manifest-converter": "^0.1.4",
-    "office-addin-usage-data": "^1.6.1",
+    "office-addin-manifest": "^1.10.0",
```
millerds, line 25: Why are we down grading these versions?

millerds (changes requested):

packages/office-addin-project/package.json
igor-ribeiiro: Are they released? They are not the latest and I just followed what VSCode recommended

millerds: I didn't check what the specific current versions are, but they shouldn't be updated just because.  The version that has been set should remain unless there is a reason to change it.  Presumably if they were changed to that version there was a reason for it and we don't want to break that unless we have a specific reason to do so.

millerds (dismissed):

millerds (approved):

PR #677, opened 2022-09-07
[Minor] Adding unique DeviceID to telemetry

igor-ribeiiro:
[Major] is because I'm removing some deprecated functions which were tagged like that for a while.
---

**Change Description**:

    Describe what the PR is about. 

No

Yes. We need to briefly explain the additional data that is being collected and where it is stored.
The data `SessionID` is a unique identifier generated when a device uses the tooling. That ID is stored at `os.homedir()/office-addin-usage-data.json`.
This information needs to be added to the documentation.

Manually tested with linking that the data generated is being logged and includes sessionIDs
Verified and fixed the unit test for them to pass.
I checked the scenarios where the file doesn't exist and when it does exist.


millerds (changes requested):
I'm a little confused by a couple of things . . .

I thought we were looking at making an ID that would stick with the machine.  A session for an individual command is that singular execution, isn't it?  Since we aren't a singular app we don't really have a "session"

I'm also confused about how the id is being stored outside the execution.  From what I can tell (and I'm not confident in my understanding) the id is being generated and then store with the telemetry output . . . but I don't see how that is persisted for use in another execution.

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -166,28 +167,20 @@ export class OfficeAddinUsageData {
    * Reports error to usage data structure
    * @param errorName Error name sent to usage data structure
    * @param err Error sent to usage data structure
+   * @Deprecated Use reportException instead
```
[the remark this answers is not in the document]
igor-ribeiiro: There were both in the code in the current time.
From my investigations, they were trying to do the exact same thing. ReportError was a little older and didn't have support for the data object, while reportException had.
So, we should just have one or the other.

Regarding naming, I personally would prefer `reportException` as that is what is currently being used in other parts of the telemetry (`reportExpectedException` for example).

igor-ribeiiro:
> 

The id is being saved in the file "C:\Users\<username>\office-addin-usage-data.json". That file already existed and is what controls the collection of data.
You can check in your machine, that file should exist and be there.

Then, I just reused the existing logic to make this information persist across multiple uses.

igor-ribeiiro:
@millerds just solved the suggestion, can you have a look at the PR?

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -166,28 +167,20 @@ export class OfficeAddinUsageData {
...
-      this.reportErrorApplicationInsights(errorName, err);
-    }
+    this.reportException(errorName, err);
   }
 
   /**
    * Reports error to Application Insights
    * @param errorName Error name sent to Application Insights
    * @param err Error sent to Application Insights
+   * @Deprecated Use reportException instead
```
millerds, line 180: I looked at this more closely . . . they are not the same thing, though they are close.  This is lighter weight and a fall back for the reportException method and you are removing that safe guard.

millerds (changes requested):

packages/office-addin-usage-data/src/usageData.ts
igor-ribeiiro: I will make it not public them, so it is only used in one place

millerds: It might already be used in other places given that is has been public.  Before making a change to this at all we should look more carefully at why the two are being used and where they are being used outside of this project.

igor-ribeiiro: I would also make a major update. I will keep it public to make it safe.

igor-ribeiiro:
@Rick-Kirkham, can we have a small section in the documentation that mentions this new DeviceID tracking?

packages/office-addin-usage-data/src/usageData.ts
millerds: You've still trying to deprecate it and use the other method.  I'm saying there shouldn't be any logging method changes in the PR that aren't specific to the new ID.

Rick-Kirkham (dismissed):

igor-ribeiiro:
@millerds PR is ready to be reviewed again

millerds (changes requested):
I guess I'm not explaining myself very well.  Don't change reportError or reportException to deprecate them or replace one with the other.  Leave them as is and just make the change for deviceId.

igor-ribeiiro:
> I guess I'm not explaining myself very well. Don't change reportError or reportException to deprecate them or replace one with the other. Leave them as is and just make the change for deviceId.

I reverted most of the changes, I will still want to put the deprecated tag as it is mostly instead for internal use

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -453,11 +456,12 @@ export class OfficeAddinUsageData {
         eventTelemetryObj.properties = {
           ...this.defaultData,
           ...data,
+          deviceID: this.options.deviceID,
         };
         this.usageDataClient.trackEvent(eventTelemetryObj);
         this.eventsSent++;
       } catch (e) {
-        this.reportError("sendUsageDataEvent", e);
+        this.reportException("sendUsageDataEvent", e);
```
millerds, line 464: Don't include this change in this PR

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -290,7 +293,7 @@ export class OfficeAddinUsageData {
 
       return err;
     } catch (err) {
-      this.reportError("maskFilePaths", err);
+      this.reportException("maskFilePaths", err);
```
millerds, line 296: Don't include this change in this PR

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -210,7 +213,7 @@ export class OfficeAddinUsageData {
       }
       jsonData.writeUsageDataJsonData(this.options.groupName, this.options.usageDataLevel);
     } catch (err) {
-      this.reportError("UsageDataOptIn", err);
+      this.reportException("UsageDataOptIn", err);
```
millerds, line 216: Don't include this change in this PR

packages/office-addin-usage-data/src/usageData.ts
```diff
@@ -152,11 +153,11 @@ export class OfficeAddinUsageData {
           usageDataEvent.properties[key] = value;
           usageDataEvent.measurements[key + " durationElapsed"] = elapsedTime;
         }
-
+        usageDataEvent.properties["deviceID"] = this.options.deviceID;
         this.usageDataClient.trackEvent(usageDataEvent);
         this.eventsSent++;
       } catch (err) {
-        this.reportError("sendUsageDataEvents", err);
+        this.reportException("sendUsageDataEvents", err);
```
millerds, line 160: Don't include this change in this PR

packages/office-addin-usage-data/src/usageData.ts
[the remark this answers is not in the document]
millerds: Don't include this change in this PR

packages/office-addin-usage-data/src/usageData.ts
millerds: Don't include this change in this PR

millerds:
> > I guess I'm not explaining myself very well. Don't change reportError or reportException to deprecate them or replace one with the other. Leave them as is and just make the change for deviceId.
> 
> I reverted most of the changes, I will still want to put the deprecated tag as it is mostly instead for internal use

The lack of knowledge of the intent is why this needs to be done separately after some investigation into the logging methods.  Just a quick look at the history shows that reportError has been around longer and keep after some "code cleanup" done by Adam 3 years ago, while reportException was added as part of a change by you a little over a year ago as part of a change to be able to report an expected exception.  This suggests that the original intent of reportError was to be used to report errors everywhere while reportException was a utility to be able to report and expected exception as opposed to a replacement for reportError.  Given the similarities we ought to reconcile the two cases (including a decision about what to deprecate), but that should be done in another PR with some careful consideration of the side effects.

millerds (changes requested):

igor-ribeiiro:
> > > I guess I'm not explaining myself very well. Don't change reportError or reportException to deprecate them or replace one with the other. Leave them as is and just make the change for deviceId.
> > 
> > 
> > I reverted most of the changes, I will still want to put the deprecated tag as it is mostly instead for internal use
> 
> The lack of knowledge of the intent is why this needs to be done separately after some investigation into the logging methods. Just a quick look at the history shows that reportError has been around longer and keep after some "code cleanup" done by Adam 3 years ago, while reportException was added as part of a change by you a little over a year ago as part of a change to be able to report an expected exception. This suggests that the original intent of reportError was to be used to report errors everywhere while reportException was a utility to be able to report and expected exception as opposed to a replacement for reportError. Given the similarities we ought to reconcile the two cases (including a decision about what to deprecate), but that should be done in another PR with some careful consideration of the side effects.

Makes sense, we can discuss about removing the `reportError` later if needed.

millerds (changes requested):
Still have a new deprecation marking that shouldn't be added.

millerds (approved):

PR #684, opened 2022-09-16
[Patch] Making conversion not leave project on bad state

igor-ribeiiro:

---

**Change Description**:

  Made the conversion check for any errors before making changes to the project (such as in the `packge.json`.

No

No

Tested conversion of the project with an invalid manifest.

### DIFF
--- a/packages/office-addin-project/package.json
+++ b/packages/office-addin-project/package.json
@@ -5,7 +5,7 @@
   "main": "./lib/main.js",
   "scripts": {
     "build": "rimraf lib && concurrently \"tsc -p tsconfig.json\"",
-    "cli": "node lib/cli.js convert -m test/invalid.manifest.xml",
+    "cli": "node lib/cli.js",
     "lint": "office-addin-lint check",
     "lint:fix": "office-addin-lint fix",
     "prettier": "office-addin-lint prettier",
--- a/packages/office-addin-project/package.json
+++ b/packages/office-addin-project/package.json
@@ -5,7 +5,7 @@
   "main": "./lib/main.js",
   "scripts": {
     "build": "rimraf lib && concurrently \"tsc -p tsconfig.json\"",
-    "cli": "node lib/cli.js",
+    "cli": "node lib/cli.js convert -m test/invalid.manifest.xml",
     "lint": "office-addin-lint check",
     "lint:fix": "office-addin-lint fix",
     "prettier": "office-addin-lint prettier",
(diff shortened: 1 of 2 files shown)


packages/office-addin-project/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "./lib/main.js",
   "scripts": {
     "build": "rimraf lib && concurrently \"tsc -p tsconfig.json\"",
-    "cli": "node lib/cli.js",
+    "cli": "node lib/cli.js convert -m test/invalid.manifest.xml",
```
millerds, line 8: I don't think we want to package.json script command to have specific arguments like this.  Those should come from the command line (I believe that is how the others work).

packages/office-addin-project/src/convert.ts
```diff
@@ -37,9 +37,14 @@ export async function convertProject(
     return;
   }
   await backupProject(backupPath);
+  try {
+    await convert(manifestPath, outputPath, false, false);
+  } catch (err: any) {
+    console.log(`Error in conversion.`);
+    throw err;
+  }
```
millerds, line 45: Is this all we need to do?  Isn't there any clean up from a partially completed call to convert.

millerds (changes requested):

packages/office-addin-project/src/convert.ts
igor-ribeiiro: There would still be a `.zip` backup file that was created. Other than that. you shouldn't see other changes in the project (Just tested again and it is like that).

millerds: Right now that may be the case in our limited experience, but there may be something we miss or something that get's changed that alters this.  The backup may be there, but we're leaving it up to the customer to switch back even when we detect something went wrong.  It is a better customer experience to report the error, but them back into the state the started with and the way to be sure of that is to restore the backup for them.

packages/office-addin-project/src/convert.ts
```diff
@@ -37,9 +37,15 @@ export async function convertProject(
...
   await backupProject(backupPath);
+  try {
+    await convert(manifestPath, outputPath, false, false);
+  } catch (err: any) {
+    console.log(`Error in conversion. Restoring project initial state.`);
+    await restoreBackup(backupPath);
+    throw err;
+  }
   updatePackages();
   await updateManifestXmlReferences();
```
millerds, line 48: These two lines should be part of the try statement.

millerds (changes requested):

millerds (approved):

PR #689, opened 2022-09-27
[Patch] Deleting old .xml manifest

igor-ribeiiro:
Deleting the `.xml` manifest after the main conversion is done.

---

**Change Description**:

    Describe what the PR is about. 

Testing conversion of the project and notice the `.xml` manifest is removed even if is not in the root of the project.

### DIFF
--- a/packages/office-addin-project/src/convert.ts
+++ b/packages/office-addin-project/src/convert.ts
@@ -40,7 +40,7 @@ export async function convertProject(
   try {
     await convert(manifestPath, outputPath, false, false);
     updatePackages();
-    await updateManifestXmlReferences();
+    await updateManifestXmlReferences(manifestPath);
   } catch (err: any) {
     console.log(`Error in conversion. Restoring project initial state.`);
     await restoreBackup(backupPath);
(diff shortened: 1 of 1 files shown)


packages/office-addin-project/src/convert.ts
```diff
@@ -132,4 +134,5 @@ async function updateManifestXmlReferences(): Promise<void> {
   // write updated json to file
   const writeFileAsync = util.promisify(fs.writeFile);
   await writeFileAsync(packageJson, JSON.stringify(content, null, 2));
+  fs.unlinkSync(manifestPath);
```
millerds, line 137: Seems like the wrong place to put this.  The method is about updating the references to the manifest (currently only in the package.json file) . . . deleting it doesn't match.  We are also slightly complicating things by adding a parameter just to make one call to delete the file.

Seems like it really belongs after the convert call in the main convertProject method or in a private method called from the main one that does the convert and file delete.

millerds (changes requested):

millerds (approved):

PR #712, opened 2022-11-19
[minor] office-addin-dev-certs: add --domains option to install

akrantz:
This adds --domains as an option to the office-addin-dev-certs install command.

This allows the set of domains to be customized if needed. By default, the domains are "127.0.0.1" and "localhost".

An example where customization is useful is if "localhost" is not allowed as a domain name. You can map a different name to the local port, but the SSL certificate needs to have that name listed.

For example, on Windows, the hosts file can map "dev-host" to 127.0.0.1 (IPv4) and ::1 (IPv6) by adding these lines to %systemroot%\system32\drivers\etc\hosts:

127.0.0.1 	dev-host
::1		dev-host

Then, use the --domains option to specify the IP addresses and host names:

office-addin-dev-certs install --domains 127.0.0.1,::1,localhost,dev-host

Opening the browser to https://dev-host will show that the SSL certificate is valid.

Yes, 

1. I validated the default behavior of "office-addin-dev-certs install" is unchanged. 
2. I validated that with the --domains option added, the SSL certificate has those domains listed. 
    For example: office-addin-dev-certs install --domains 127.0.0.1.::1,localhost,dev-host
3. I also validated the browser shows the url https://dev-host contains a valid SSL certificate and is trusted.

### DIFF
--- a/packages/office-addin-dev-certs/src/cli.ts
+++ b/packages/office-addin-dev-certs/src/cli.ts
@@ -20,6 +20,7 @@ commander
     "--days <days>",
     `Specifies the validity of CA certificate in days. Default: ${defaults.daysUntilCertificateExpires}`
   )
+  .option("--domains <domains>", "List of IP address and domains separated by commas.")
   .description(`Generate an SSL certificate for "localhost" issued by a CA certificate which is installed.`)
   .action(commands.install);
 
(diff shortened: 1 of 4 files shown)


akrantz:
@Rick-Kirkham You can decided the amount of documentation which is needed for this. It has not come from other user feedback but rather a scenario I've encountered where we need to use a different name besides "localhost".

millerds (approved):

PR #756, opened 2023-03-10
[minor] Cf support multiple function files

millerds:

---

**Change Description**:
The custom function metadata plugin assumes only one source file and generates a json for just that one file.  This will add the capability to accept an array of file paths and generate a json for the all of the functions in all of the files.  This needs to get published before making the plugin change that will add the support to webpack specifically.

Yes.  The file path parameter can now accept a comma separated list of paths as the string.

Yes.  The documentation could include the information about being able to use multiple files.

Tried using the package in a CF test project with different inputs.  Used the cli with different inputs.  Add a test that included multiple files as inputs.


packages/custom-functions-metadata/src/cli.ts
```diff
@@ -3,29 +3,44 @@
...
+program.name("custom-functions-metadata").version(process.env.npm_package_version || "(version not available)");
 
-commander
-  .command("generate <source-file> [output-file]")
-  .description("Generate the metadata for the custom functions from the source code.")
+program
+  .command("generate")
+  .argument(
+    "<source-files>",
+    "The path to the source file (or comma seperated list of files) for custom functions.",
```
Rick-Kirkham, line 19: spelling:

```suggestion
    "The path to the source file (or comma separated list of files) for custom functions.",
```

Rick-Kirkham (dismissed):
Approving with a recommended spelling fix.

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -140,31 +141,43 @@ type CustomFunctionsSchemaDimensionality = "invalid" | "scalar" | "matrix";
...
-    parseTreeResult.extras.forEach((extra) => extra.errors.forEach((err) => generateResults.errors.push(err)));
-
-    if (generateResults.errors.length > 0) {
-      if (wantConsoleOutput) {
-        console.error("Errors in file: " + inputFile);
-        generateResults.errors.forEach((err) => console.error(err));
+  if (input && inputFiles.length > 0) {
+    inputFiles.forEach((inputFile) => {
+      if (!inputFile) {
+        throw new Error(`Input file name is empty or has a comma at the beginning or end.\n Input: ${input}`);
```
kwame-ansong, line 158: I'm not familiar with this code, so this comment is probably a nit :),

This error seems very specific but also very useful.
Should we say that the filename is invalid, and then also say "For example, it has a comma at the beginning or end", something like that? Or are the filenames always valid except when there's a comma somewhere? In other words, that's the only error we can ever have in a filename from `inputFiles`?

akrantz:
We had originally supported multiple files, but this was done by requirig the functions in other files to be imported into the main functions.ts file.  This seems to allow being able to have the webpack plugin specify multiple files.

packages/custom-functions-metadata/src/generate.ts
millerds: It is specific and maybe too much so (I wasn't sure myself).  With this change I'm introducing using a comma delimited list of file paths being input as a string and if the comma is at the beginning or the end it gets converted into an array of two strings with one being empty which could easily get overlooked (I know I could overlook it).  This was intended to help point that out.  Other solutions could be to automatically trim the comma, automatically skip empty strings in the new array, or make the error general and hope they figure it out be reading documentation.  Thoughts on a better way?

millerds: Went ahead and changed the error checking to trim and ignore empty strings instead of throwing an error message (except for the existing check for no input at all).

Rick-Kirkham (approved):

zhngx1 (approved):

kwame-ansong (approved):

PR #759, opened 2023-03-18
[patch] Fix dev-certs verify.ps1

parched:
**Change Description**:

Previously, it could be expired for up to 2 days but still verify.

No

No

I can now run my addin from vscode and not see
image

but now I just see blank and it's stuck saying "We're loading the add-ins run time" Not sure if it's related or not.

### DIFF
--- a/packages/office-addin-dev-certs/scripts/verify.ps1
+++ b/packages/office-addin-dev-certs/scripts/verify.ps1
@@ -30,7 +30,7 @@ if ($PSVersionTable.PSVersion.Major -le 5) {
 if(Get-Command -name Import-Certificate -ErrorAction SilentlyContinue){
     $result = Get-ChildItem cert:\\CurrentUser\\Root | Where-Object Issuer -like "*CN=$CaCertificateName*"
     if (!$ReturnInvalidCertificate) {
-        $result = $result | Where-Object { $_.NotAfter -gt (Get-Date) }
+        $result = $result | Where-Object { $_.NotAfter -gt (Get-Date).AddDays(1) }
         if ($result -and ($result.Length -eq 1) -and (Test-Path $CaCertificatePath) -and (Test-Path $LocalhostCertificatePath)) {
             # Check that CA certificate in store is the same as ca.crt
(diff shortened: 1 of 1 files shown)


millerds:
The way I read it the certificate will be expiring (_NotAfter) soon (after today minius -1) rather than up to a day ago.

parched:
Not sure I follow, isn't today-1 a day ago?

parched:
I edited the description, it's actually up to 2 days. I suspect the `-1` was intended to be `+1`. But checking the actual time of day with `Get-Date` seems more correct.

akrantz:
The intent was that renewing the certificate happens up to a day before it actually is due to expire. This was to prevent starting the dev server 1 hour before expiration and having it expire while the dev server is running. A day of padding seems sufficient.

I think the command is should be looking for certificates which are valid after the current day/time plus 1 day.

It does seem like there is a logic error, in that the "valid until" date should be calculated as current time plus 1 day instead of minus 1 day.

It might be better for the code for dev-certs install command to pass the desired "valud until" date, and the verify command by default could just use the current time.

parched:
Ok, I've added 1 day buffer now. But feel free to close if you'd like to handle this a different way.

millerds:
I see it now . . . the +1 seems the correct logic, however I believe the "get-date" only returns the date and not the time while the expiration includes a time element.  That is probably why "[datetime]::today" was used.  You actually end up measuring less than a day and more so the closer you get to midnight.

parched:
It looks like it's the other way round, that's why I changed it to `Get-Date`.
image
Perhaps the intent of using `today` was to give some buffer (assuming developers aren't working at midnight) but it was actually negative buffer.

akrantz:
I'm not clear why you want to switch to today so the time is always midnight, rather than use get-date to use the current date and time. today + 1 day is the upcoming midnight, so close to midnight there is only a small amount of time. With get-date + 1 day, you get a 24 hour buffer.

millerds:
I got the two backwards . . . get-date is better.

PR #761, opened 2023-03-23
[patch] Prevent false positive in Root CA certificate check by ignoring extraneous PowerShell output

apolakipso:
The Root CA certificate wasn't properly installed when running `npm start`, turns out the installation was skipped because the code interpreted PowerShell output from my profile as a successful check for an installed certificate.

The change: A PowerShell command might generate initial output (in my case it was some information echo'd in my profile) - this broke the check since the code assumes the cert exists if *any content* is returned. The code now only returns true if the name of the certificate is contained in the returned output:
```bash
Subject      : O=Developer CA for Microsoft Office Add-ins, L=Redmond, S=WA, C=US, CN=Developer CA for Microsoft
               Office Add-ins
```
The code matches the **`O=Developer CA for Microsoft Office Add-ins,`** part.

No

No

I ran `npm start` without the change and the Root CA certificate was not installed. After the change, `npm start` correctly determined the certificate wasn't present and installed it.

### DIFF
--- a/packages/office-addin-dev-certs/src/verify.ts
+++ b/packages/office-addin-dev-certs/src/verify.ts
@@ -9,13 +9,16 @@ import * as defaults from "./defaults";
 import { usageDataObject } from "./defaults";
 import { ExpectedError } from "office-addin-usage-data";
 
+// On win32 this is a unique hash used with PowerShell command to reliably delineate command output
+export const outputMarker = process.platform === "win32" ? `[${crypto.createHash("md5").update(`${defaults.certificateName}${defaults.caCertificatePath}`).digest("hex")}]` : "";
+
 /* global process, Buffer, __dirname */
 
 function getVerifyCommand(returnInvalidCertificate: boolean): string {
   switch (process.platform) {
     case "win32": {
       const script = path.resolve(__dirname, "..\\scripts\\verify.ps1");
-      const defaultCommand = `powershell -ExecutionPolicy Bypass -File "${script}" -CaCertificateName "${defaults.certificateName}" -CaCertificatePath "${defaults.caCertificatePath}" -LocalhostCertificatePath "${defaults.localhostCertificatePath}"`;
+      const defaultCommand = `powershell -ExecutionPolicy Bypass -File "${script}" -CaCertificateName "${defaults.certificateName}" -CaCertificatePath "${defaults.caCertificatePath}" -LocalhostCertificatePath "${defaults.localhostCertificatePath}" -OutputMarker "${outputMarker}"`;
       if (returnInvalidCertificate) {
         return defaultCommand + ` -ReturnInvalidCertificate`;
       }
@@ -38,6 +41,10 @@ export function isCaCertificateInstalled(returnInvalidCertificate: boolean = fal
 
   try {
     const output = execSync(command, { stdio: "pipe" }).toString();
+    if (process.platform === "win32") {
+      // Remove any PowerShell output that preceeds invoking the actual certificate check command
+      return output.slice(output.lastIndexOf(outputMarker) + outputMarker.length).trim().length !== 0;
+    }
     // script files return empty string if the certificate not found or expired
     if (output.length !== 0) {
       return true;
--- a/packages/office-addin-dev-certs/scripts/verify.ps1
+++ b/packages/office-addin-dev-certs/scripts/verify.ps1
@@ -14,9 +14,19 @@ Param (
     [string]
     $LocalhostCertificatePath,
 
+    [Parameter(Mandatory = $false)]
+    [ValidateNotNullOrEmpty()]
+    [string]
+    $OutputMarker,
+
     [switch]
     $ReturnInvalidCertificate
 )
+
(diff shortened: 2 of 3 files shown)


akrantz:
Does this work even for non-Powershell on Windows or on other platforms?  Could other platforms use a different string in the response?

millerds:
The way the check is written it's always calling PowerShell on Windows.  However, mac uses shell script and linux uses direct command and I'm sure that the mac case doesn't return the cert name in the response . . . rather it returns the encrypted cert string.

apolakipso:
The change only affects the `win32` platform now. I used a simple `if` (instead of the full `switch` rigmarole) as there is only the one special case for win32. I also returned the result directly. Let me know if this is sufficient.

millerds (approved):

millerds:
Looks like it failed a test case (the one specifically trying to verify a cert is installed).

apolakipso:
I'm not sure why the tests failed, but have rewritten the change anyway.
Previously, I've tried these:

### Use PowerShell's `-NoProfile` parameter to disable the profile
Found this not to be a good option since whatever is configured in the profile might be necessary for the command to succeed

### Checking for the certificate name
I thought this should work (it did locally), but getting even a bit specific (like searching the certificate name in the output) was enough to break the tests (I think). Depending on the actual output as little as possible seems the way to go.

I finally went with this:

### Pass unique marker string to `verify.ps1` script which can be used to separate PowerShell profile output from the command's output

This was an easy way to ensure we're only evaluating the command's output and not anything the PowerShell prints on startup. I've found numerous ways people use to prevent PowerShell output, but all of those seemed too hacky and fragile. This introduces an additional parameter to the `verify.ps1` script, which I found acceptable.

apolakipso:

packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -9,13 +9,15 @@ import * as defaults from "./defaults";
 import { usageDataObject } from "./defaults";
 import { ExpectedError } from "office-addin-usage-data";
 
+const powershellOutputMarker = `[${new Date().getTime()}]`
```
akrantz, line 12: This should be for Win32 only. Can you move it before defaultCommand variable below which is inside case "win32"?

apolakipso: The value should be unique and is used later on in `isCaCertificateInstalled` to check against the actual result string. 
Not sure how to rewrite this and keeping it readable. I could use a generator function (see below), but that seems way less readable. Please advise.

```js
function* generateOutputMarker() {
  const value = `[${new Date().getTime()}]`;
  while (true) {
    yield value;
  }
}

// Usage in `getVerifyCommand` and `isCaCertificateInstalled`
const outputMarker = generateOutputMarker().next().value
```

apolakipso: Please let me know if you disagree with creating anything in that scope that relates to a single platform or just with the variable being created even when we might not use it.

This is simpler still, since `crypto` is already in here:

```js
function getOutputMarker() {
    return `[${crypto.createHash("md5").update(JSON.stringify(defaults).join('')).digest("hex")}]`
}
```
This would then only be called in the `win32` paths in  `getVerifyCommand` and `isCaCertificateInstalled` like so:
```js
const outputMarker  = getOutputMarker()
```

Simply putting this in two places seems needlessly fragile:
```js
const outputMarker  = `[${crypto.createHash("md5").update(JSON.stringify(defaults).join('')).digest("hex")}]`
```

packages/office-addin-dev-certs/src/verify.ts
```diff
@@ -38,6 +43,12 @@ export function isCaCertificateInstalled(returnInvalidCertificate: boolean = fal
 
   try {
     const output = execSync(command, { stdio: "pipe" }).toString();
+    // script files return the certificate name if the certificate is found and hasn't expired
```
apolakipso, line 46: Remove comment

millerds:
I think using the hash as the marker and using the output marker const instead of the method is cleaner.  I don't think we need a local copy of the marker in each places it's used.

apolakipso:
@millerds Thank you for your feedback, I changed the commit accordingly. 

NB: This creates the hash for all platforms, which was the point of @akrantz earlier comment I think:
> This should be for Win32 only. Can you move it before defaultCommand variable below which is inside case "win32"?

If that is a no-go, I'd guard it with this (or a full `if` if ternary operators are not wanted in the code base)
```js
const outputMarker = process.platform === "win32" ? `[${crypto.createHash("md5").update(`${defaults.certificateName}${defaults.caCertificatePath}`).digest("hex")}]` : null;
```

Please let me know 😃

millerds (dismissed):

apolakipso:
I guess this failed since the test doesn't account for the output marker, I'll have a look and see whether I can make sense of the tests. No idea yet how the errors regarding invalid JSON relate to this 🙁

apolakipso:
Exported the output marker and have the test use it

millerds (approved):

millerds:
Thanks for the contribution!

PR #819, opened 2023-11-17
[minor] Fix 'uninstall' and 'verify' for Linux platform in office-addin-dev-certs package.

nwhitehead:
Fixes a bug with office-addin-dev-certs on Linux. The verify command always thought the cert was not installed, causing every run of a generic sample app to generate and install new certificates. This was frustrating when trying to debug certificate issues when running a sample app for the first time.


nwhitehead:

PR #821, opened 2023-12-11
Fix check for Microsoft Azure CLI substring

davidchesnut:
Microsoft Azure CLI (32-bit) and similar strings will no longer match the check. Modified to check for the substring instead.

This is a patch.
Needs someone to double-check testing that this works correctly when Azure CLI is installed, and when it is uninstalled.


millerds (approved):

millerds:
Looks like linux is failing because it's already using node v20 and the dependency we have on teams package (and it's nested dependencies) creates a warning on install which fails the tests.  This change is specific to Win32 (an needed) so I'm going to merge it any way.

PR #824, opened 2024-01-03
[patch] Mark support for npm v10

millerds:

---

**Change Description**:

    Indicate support for npm v10

No.

No.

Ran npm install and automated tests with npm v10 installed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "engines": {
     "node": ">=16 <19",
-    "npm": ">=6 <=9"
+    "npm": ">=6 <=11"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "engines": {
     "node": ">=16 <19",
```
akrantz, line 5: You should also support Node LTS which is v20 now, so bump it to <21.

millerds: I would, except we have a dependency that doesn't support v20 yet and we need an update for that first.

PR #825, opened 2024-01-08
Fix axios and other vulnerabilities

kwame-ansong:
Upgrade axios and other packages to address vulnerabilities.

### DIFF
--- a/packages/office-addin-dev-settings/package-lock.json
+++ b/packages/office-addin-dev-settings/package-lock.json
@@ -2328,22 +2328,47 @@
       }
     },
     "node_modules/@microsoft/dev-tunnels-management": {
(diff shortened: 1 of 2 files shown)


lsmith77:
I think it would also make sense to bump up the min versions in the `package.json`

millerds:
Can you also update the direct reference to teamsfx-cli in packages/office-addin-dev-settings/src/publish.ts.  The string used for the cli command is currently referencing the alpha version to get some preview fixes that should now be part of the release teamsfx-cli so the version reference can become 2.1.0 instead of the current alpha reference.

kwame-ansong:
Abandoning this for now. A different PR will be sent out in place of this one.

PR #829, opened 2024-01-17
[Patch] Set allowCustomDataForDataTypeAny to true by default for custom function metadata generation

xai-msft:
**Change Description**:

  Custom function metadata generation now adds `allowCustomDataForDataTypeAny:true` by default to functions.json.

    No

    On page [Use data types with custom functions in Excel](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-data-types-concepts), it is mentioned that "To use the data types integration with custom functions, the custom functions JSON metadata file must be manually updated to include the property allowCustomDataForDataTypeAny. Set this property to true.". This paragraph can be removed since manually adding allowCustomDataForDataTypeAny is no longer needed.

npm run test passes

### DIFF
--- a/packages/custom-functions-metadata/src/generate.ts
+++ b/packages/custom-functions-metadata/src/generate.ts
@@ -178,8 +178,8 @@ export async function generateCustomFunctionsMetadata(
 
     if (functions.length > 0) {
       const metadata = {
-        functions: functions,
         allowCustomDataForDataTypeAny: true,
+        functions: functions,
       }
       generateResults.metadataJson = JSON.stringify(metadata, null, 4);
     }
--- a/packages/custom-functions-metadata/src/generate.ts
+++ b/packages/custom-functions-metadata/src/generate.ts
@@ -177,10 +177,11 @@ export async function generateCustomFunctionsMetadata(
     });
 
     if (functions.length > 0) {
-      generateResults.metadataJson = JSON.stringify({
+      const metadata = {
         functions: functions,
         allowCustomDataForDataTypeAny: true,
-      }, null, 4);
+      }
+      generateResults.metadataJson = JSON.stringify(metadata, null, 4);
     }
   }
 
--- a/packages/custom-functions-metadata/src/generate.ts
+++ b/packages/custom-functions-metadata/src/generate.ts
@@ -177,7 +177,10 @@ export async function generateCustomFunctionsMetadata(
     });
 
     if (functions.length > 0) {
-      generateResults.metadataJson = JSON.stringify({ functions: functions }, null, 4);
+      generateResults.metadataJson = JSON.stringify({
+        functions: functions,
+        allowCustomDataForDataTypeAny: true,
+      }, null, 4);
     }
   }
 
--- a/packages/custom-functions-metadata-plugin/package.json
+++ b/packages/custom-functions-metadata-plugin/package.json
@@ -1,6 +1,6 @@
 {
   "name": "custom-functions-metadata-plugin",
-  "version": "1.5.3",
+  "version": "1.5.4",
   "author": "Office Dev",
   "bugs": {
     "url": "https://github.com/OfficeDev/Office-Addin-Scripts/issues"
@@ -26,7 +26,7 @@
     "watch": "tsc -p tsconfig.json -w"
   },
   "dependencies": {
-    "custom-functions-metadata": "^1.5.3"
+    "custom-functions-metadata": "^1.5.4"
   },
   "devDependencies": {
     "@types/node": "^14.17.2",
--- a/packages/custom-functions-metadata/package.json
+++ b/packages/custom-functions-metadata/package.json
@@ -1,6 +1,6 @@
 {
   "name": "custom-functions-metadata",
-  "version": "1.5.3",
+  "version": "1.5.4",
   "description": "Generate metadata for Excel Custom Functions.",
   "main": "./lib/main.js",
   "scripts": {
--- a/packages/custom-functions-metadata/test/cases/basic/expected.json
+++ b/packages/custom-functions-metadata/test/cases/basic/expected.json
@@ -7,5 +7,6 @@
             "parameters": [],
             "result": {}
         }
-    ]
+    ],
+    "allowCustomDataForDataTypeAny": true
 }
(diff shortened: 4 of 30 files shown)


packages/custom-functions-metadata/src/generate.ts
```diff
@@ -177,7 +177,10 @@ export async function generateCustomFunctionsMetadata(
     });
 
     if (functions.length > 0) {
-      generateResults.metadataJson = JSON.stringify({ functions: functions }, null, 4);
+      generateResults.metadataJson = JSON.stringify({
+        functions: functions,
+        allowCustomDataForDataTypeAny: true,
```
akrantz, line 182: This is certainly simple enough to turn this on by default. However, I would like to know more about the implication of this. If this is enabled, does this affect custom functions which do not return entity data types? 

If custom functions are being used to return entity data types, I believe this is an object type which relates to [CellValue](https://learn.microsoft.com/en-us/javascript/api/excel/excel.cellvalue). I believe the ideal thing here is to have type info for these types and then it can validate that the object is one of the acceptable types. But it might be easiest to start to just say if `object` type is returned for any custom function, that triggers setting allowCustomDataForDataTypeAny to true.

akrantz (changes requested):
In an email discussion I raised the question of whether allowCustomDataForDataTypeAny can be true by default, and that might be the case, but I want to make sure we understand the ramifications of this, especially since once the change is made, it will automatically apply for all developers when they get the updated package, and for ScriptLab when it is integrated.

Another alternative is to check if any `object` type is returned for any custom function, and only set it when that is the case. This means that rather than say the return type is `any` (or `unknown`), it would know that it is an object type.

packages/custom-functions-metadata/src/generate.ts
akrantz: @xai-msft Should we discuss this offline? Please schedule a call if that is more helpful to discuss this. Thanks.

xai-msft: > @xai-msft Should we discuss this offline? Please schedule a call if that is more helpful to discuss this. Thanks.

@akrantz Sorry, I was OOF last week. I will schedule a meeting soon.

akrantz:
By the way, you shouldn't use master on your fork. You should use `git checkout -b <name>` to create a named branch in your fork, and then have the PR use that branch. This ensures you don't have problems when you pull changes back to the master branch in your fork.

akrantz:
Tests will need to be updated. Make sure to run `npm run test` on your changes.

xai-msft:
> By the way, you shouldn't use master on your fork. You should use `git checkout -b <name>` to create a named branch in your fork, and then have the PR use that branch. This ensures you don't have problems when you pull changes back to the master branch in your fork.

Thanks for the comment. I'll make sure to do it next time.

xai-msft:
Tests are updated. We believe that there is little risk since ScriptLab has it enabled by default too.

akrantz (changes requested):
Should allowErrorForDataTypeAny also be set to true?  I think it would only call into functions which have a parameter type of any in the functions.json.

There may also be other fixes needed in custom-functions-metadata as I saw an issue with the webpack plugin with running build or dev-server scripts in Excel-Custom-Functions project. You can build office-addin-scripts from the root, then use `npm link c:\dev\office-addin-scripts\packages\custom-functions-metadata-plugin` in Excel-Custom-Functions project to have it use the code.

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -177,7 +177,11 @@ export async function generateCustomFunctionsMetadata(
     });
 
     if (functions.length > 0) {
-      generateResults.metadataJson = JSON.stringify({ functions: functions }, null, 4);
+      const metadata = {
+        functions: functions,
+        allowCustomDataForDataTypeAny: true,
```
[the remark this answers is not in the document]
xai-msft: Sure. I'll update this.

xai-msft:
> Should allowErrorForDataTypeAny also be set to true? I think it would only call into functions which have a parameter type of any in the functions.json.
> 
> There may also be other fixes needed in custom-functions-metadata as I saw an issue with the webpack plugin with running build or dev-server scripts in Excel-Custom-Functions project. You can build office-addin-scripts from the root, then use `npm link c:\dev\office-addin-scripts\packages\custom-functions-metadata-plugin` in Excel-Custom-Functions project to have it use the code.

`allowErrorsForDataTypeAny` hasn't been set to true in ScriptLab, and we don't really know if there had been a reason for that. `allowCustomDataForDataTypeAny ` enables errors as input too, so functionally setting `allowCustomDataForDataTypeAny` is already enough. I don't feel confident about how safe `allowErrorsForDataTypeAny` is, since it hasn't been widely used, unlike `allowCustomDataForDataTypeAny` with ScriptLab.

Regarding the issue with using the plug-in in a yo office project: I reproduced it locally and will look into that.

akrantz:
One suggestion for allowErrorForDataTypeAny. We could have a option for the plugin as to whether to enable it, e.g. `allowErrorDataTypes: boolean` and that can be passed to custom-functions-metadata such that it adds `allowErrorForDataTypeAny: true` to the metadata only when true. That would allow a user to update webpack config to enable the option. This could be done in a separate PR.

akrantz (changes requested):
Let me know once updates are made and I'll approve. Thanks!

xai-msft:
@akrantz I updated the code & tested.

Regarding the issue with building add-in code with npm linked build: I tried with the original master branch and encountered the same error as well, so I think it's ok/unrelated.

Thank you!

akrantz (approved):
Looks good. There are some suggestions that I've made which could be done in other pull requests. Thanks @xai-msft !

mliu-rgare:
Hello @akrantz and @xai-msft , I'm working on the development of an Excel Add-in and I'd like to be able to pass ISO8601 dates to Excel and have them be picked up as dates rather than general formatting. It appears as if this patch to 1.5.4 wasn't published to npm. Is there additional work to be done at this time?

millerds:
There is the penitential for another change to go into this repo before we publish all the changes together.

PR #836, opened 2024-02-15
[patch] custom-function-metadata

wandyezj:
** Change Description **

Refactor custom-function-metadata to separate `parseTree` from `generateCustomFunctionMetadata`.
This allows `parseTree` to be used in the browser without having to polyfill `fs`.

break generate.ts into two files: `parseTree.ts` and `generateCustomFunctionMetadata.ts`.

update main.ts exports to specify exact types that are exported.

    No

    No

    Ran unit tests.


packages/custom-functions-metadata/src/commands.ts
```diff
@@ -3,7 +3,7 @@
 
 import { writeFileSync } from "fs";
 import { logErrorMessage } from "office-addin-usage-data";
-import { generateCustomFunctionsMetadata } from "./generate";
```
akrantz, line 6: Is there really a need to rename this file?

akrantz:
Are these changes driven from ScriptLab code changes? Is there a PR there for the usage of this.

There's no description in this PR about the changes -- what they are, or why the changes are being made.

packages/custom-functions-metadata/src/generate.ts
```diff
@@ -1,1170 +1,2 @@
...
- * Log containing all the errors found while parsing
- * @param error Error string to add to the log
- */
-export function logError(error: string, position?: ts.LineAndCharacter | null): string {
-  if (position) {
-    error = `${error} (${position.line + 1},${position.character + 1})`;
-  }
-  return error;
-}
+export * from "./generateCustomFunctionsMetadata";
```
akrantz, line 1: This doesn't seem like a good change. I'd like to understand what the core motivations are here.

akrantz (changes requested):
I'd prefer to keep gnerate.ts more as it was. Would like to understand if there is some plication about parseTree functionality and integration with ScriptLab which is being made here.

packages/custom-functions-metadata/src/main.ts
```diff
@@ -1,4 +1,18 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 
-export * from "./generate";
+export { generateCustomFunctionsMetadata } from "./generateCustomFunctionsMetadata";
```
akrantz, line 4: Would prefer to keep `export *` as it was originally.

wandyezj:
> Refactor custom-function-metadata to separate `parseTree` from `generateCustomFunctionMetadata`. This allows `parseTree` to be used in the browser without having to polyfill `fs`.
> 
> break generate.ts into two files: `parseTree.ts` and `generateCustomFunctionMetadata.ts`.

See PR description:
```
Refactor custom-function-metadata to separate parseTree from generateCustomFunctionMetadata.
This allows parseTree to be used in the browser without having to polyfill fs.

break generate.ts into two files: parseTree.ts and generateCustomFunctionMetadata.ts.
```

packages/custom-functions-metadata/src/main.ts
wandyezj: I changed it because it was very difficult to understand what exactly this package was exporting.

This gives a complete list of types that are intended for export.

packages/custom-functions-metadata/src/generate.ts
wandyezj: > allow parseTree to be used in the browser without having to polyfill fs.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -0,0 +1,1114 @@
...
+  const lastParameterPosition = getPosition(lastParameter);
+
+  // Try and determine the return type.  If one can't be determined we will set to any type
+  if (isStreamingFunction) {
+    const lastParameterType = lastParameter.type as ts.TypeReferenceNode;
+    if (!lastParameterType) {
+      // Need to get result type from param {type}
+      const name = (lastParameter.name as ts.Identifier).text;
+      const ptype = jsDocParamTypeInfo[name];
+      // @ts-ignore
```
wandyezj, line 486: Copied from original: `@ts-ignore` is generally something to avoid. Generally, this should be flagged by linter rules.

wandyezj:
> I'd prefer to keep gnerate.ts more as it was. Would like to understand if there is some plication about parseTree functionality and integration with ScriptLab which is being made here.

It is challenging to make this work. 

1. It's clear that other packages are importing generate directly instead of relying on the package index - as evidenced by test suite failures when not everything is exported from `generate.ts`. Yes, referencing files directly instead of using the index is generally considered a bad practice. However, it's convenient to avoid breaking existing usage.
2. parseTree needs to be in its own file separate from files that are importing fs.

This devolves to the existing refactor of separate files for parseTree and generateCustomFunctionMetadata and then exporting these as done previously from generate.ts to avoid breakage.

It's also worth looking at root causes about why we are importing from generate.ts directly instead of the index file. The root cause here is it's __really confusing__ what is exported from this package since the exports are not clear from the index. I fix that here as well to encourage proper importing.

packages/custom-functions-metadata/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "custom-functions-metadata",
-  "version": "1.5.5",
+  "version": "1.5.6",
```
millerds, line 3: We shouldn't change the version manually . . . that is handled by Lerna when we publish the packages.

packages/custom-functions-metadata/src/main.ts
millerds: Triage says this is OK, but the change should be done in all packages in order to be consistent in the repo.  It should be done in a separate PR along with the other packages and we should continue to us `export *` for this PR.

millerds (changes requested):
A few tweaks and this would be accepted.  Also need to resolve conflicts.

millerds:
Appears to have been abandoned.

PR #839, opened 2024-02-15
custom-functions-metadata: move parseTree() to avoid fs import

akrantz:
Move parseTree() to ./src/parseTree.ts keeping generateCustomFunctionMetadata() in generate.ts.

This is to ensure ScriptLab can import parseTree() without having to import fs package,
because there's no longer a polyfill from react-scripts.

No

No

npm run test


packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -0,0 +1,1116 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as ts from "typescript";
+import XRegExp = require("xregexp");
```
wandyezj, line 5: Is it possible to import this using `import XRegExp from xregexp`

packages/custom-functions-metadata/test/src/test.ts
```diff
@@ -2,9 +2,10 @@
 // Licensed under the MIT license.
 
 import * as assert from "assert";
-import * as fs from "fs";
+import { readFileSync } from "fs";
 import * as mocha from "mocha";
-import { generateCustomFunctionsMetadata, IParseTreeResult, parseTree } from "../../src/generate";
+import { generateCustomFunctionsMetadata } from "../../src/generate";
```
wandyezj, line 7: How come import these separately instead of from main?

packages/custom-functions-metadata/src/parseTree.ts
akrantz: I suspect it needs to use this type of import.

wandyezj (approved):

packages/custom-functions-metadata/test/src/test.ts
akrantz: I think generally tests imported directly but not sure there's a reason to do it a different way.

wandyezj: Generally, it might be preferable to import from the package if testing external exports. This ensures what you intend to export is (1) exported (2) tested. Additionally, (3) it avoids having to touch unit test files to adjust imports when package internals change (less churn).

packages/custom-functions-metadata/src/parseTree.ts
wandyezj: I'm not sure that's the case?

Generally, it's probably best to use `import` unless there is a really good reason to use require. 

https://stackoverflow.com/questions/46677752/the-difference-between-requirex-and-import-x

packages/custom-functions-metadata/test/src/test.ts
akrantz: That point of the tests is not to test the external exports but rather the functionality.

packages/custom-functions-metadata/src/parseTree.ts
akrantz: This is not just a require() statement. It is a TypeScript import statement which is compatible with require. See https://www.typescriptlang.org/docs/handbook/modules/reference.html#export--and-import--require

PR #844, opened 2024-03-07
[minor] Additional exclude custom function metadata filters

jeaniehuynh:
**Change Description**:

    Add a couple more mutually exclusive tags for @linkedEntityDataProvider custom functions.

### DIFF
--- a/packages/custom-functions-metadata/Specification.md
+++ b/packages/custom-functions-metadata/Specification.md
@@ -80,14 +80,14 @@ The provided _url_ is displayed in Excel.
 ---
 ### @linkedEntityDataProvider
 
-Indicates that the function is a "special" custom function that is meant to act as the "loadFunction" for user defined `LinkedDataType`s. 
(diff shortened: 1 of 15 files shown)


akrantz (dismissed):
Looks good.

millerds:
Why the changes to the package-lock.json files?  They seem to be removing things (references to other office-addin-script packages) that shouldn't be.

jeaniehuynh:
> Why the changes to the package-lock.json files? They seem to be removing things (references to other office-addin-script packages) that shouldn't be.

Those were just the changes that autogenerated when I ran the build. That said, I did initially run a build from the root directory first (not just from the custom-functions-metadata package) if that may have caused it? I can revert and rebuild to double check.

akrantz:
When I looked at the package lock files, it pulled in the minor dependent package changes which happen. It's not the build but "npm install" at the root and the associated "lerna bootstrap" that is triggered which updates these. It's also fine to not add them.

akrantz (approved):

PR #845, opened 2024-03-14
[minor] Add TAOS sideload support for MetaOS WXP

QuanfuXiao:
When the manifest is WXP JSON, the following changes are made:
- Switch to TAOS sideload option.
- Launch host directly.

### DIFF
--- a/packages/office-addin-dev-settings/src/dev-settings-windows.ts
+++ b/packages/office-addin-dev-settings/src/dev-settings-windows.ts
@@ -9,6 +9,7 @@ import {
   ManifestInfo,
   OfficeApp,
   OfficeAddinManifest,
+  ManifestType,
 } from "office-addin-manifest";
 import { DebuggingMethod, RegisteredAddin, SourceBundleUrlComponents, WebViewType } from "./dev-settings";
 import { ExpectedError } from "office-addin-usage-data";
@@ -178,7 +179,7 @@ export async function registerAddIn(manifestPath: string, registration?: string)
   const appsInManifest = getOfficeAppsForManifestHosts(manifest.hosts);
 
   // Register using the service
-  if (manifestPath.endsWith(".json") || appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
+  if (manifest.manifestType === ManifestType.JSON || appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
     if (!registration) {
       let filePath = "";
       if (manifestPath.endsWith(".json")) {
--- a/packages/office-addin-dev-settings/src/dev-settings-windows.ts
+++ b/packages/office-addin-dev-settings/src/dev-settings-windows.ts
@@ -182,10 +182,10 @@ export async function registerAddIn(manifestPath: string, registration?: string)
   if (manifest.manifestType === ManifestType.JSON || appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
     if (!registration) {
       let filePath = "";
-      if (manifestPath.endsWith(".json")) {
+      if (manifest.manifestType === ManifestType.JSON) {
         const targetPath: string = fspath.join(process.env.TEMP as string, "manifest.zip");
         filePath = await exportMetadataPackage(targetPath, manifestPath);
-      } else if (manifestPath.endsWith(".xml")) {
+      } else if (manifest.manifestType === ManifestType.XML) {
         filePath = manifestPath;
       }
       registration = await registerWithTeams(filePath);
--- a/packages/office-addin-dev-settings/src/dev-settings-windows.ts
+++ b/packages/office-addin-dev-settings/src/dev-settings-windows.ts
@@ -178,7 +178,7 @@ export async function registerAddIn(manifestPath: string, registration?: string)
   const appsInManifest = getOfficeAppsForManifestHosts(manifest.hosts);
 
   // Register using the service
-  if (appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
+  if (manifestPath.endsWith(".json") || appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
     if (!registration) {
       let filePath = "";
       if (manifestPath.endsWith(".json")) {
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -80,11 +80,11 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
     if (!extEntry) {
       throw new ExpectedError("webextension was not found.");
     }
-    
+
     const webExtensionXml = templateZip
-    .readAsText(extEntry)
-    .replace(/00000000-0000-0000-0000-000000000000/g, manifest.id)
-    .replace(/1.0.0.0/g, manifest.version);
+      .readAsText(extEntry)
+      .replace(/00000000-0000-0000-0000-000000000000/g, manifest.id)
+      .replace(/1.0.0.0/g, manifest.version);
 
     templateZip.getEntries().forEach(function (entry) {
       var data: Buffer = entry.getData();
@@ -92,8 +92,8 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
         data = Buffer.from(webExtensionXml);
       }
       outZip.addFile(entry.entryName, data, entry.comment, entry.attr);
-      
-      const webExtensionFolderPath = webExtensionPath.replace(/(\/[^\/]*)$/, '');
+
+      const webExtensionFolderPath = webExtensionPath.substring(0, webExtensionPath.lastIndexOf("/"))
       // If manifestType is JSON, remove the web extension folder
       if (entry.entryName.startsWith(webExtensionFolderPath) && manifest.manifestType == ManifestType.JSON) {
         outZip.deleteFile(entry.entryName);
@@ -368,7 +368,7 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
     throw new ExpectedError(`Sideload to the ${getOfficeAppName(app)} app is not supported.`);
   }
 
-  // for Outlook, Word, Excel, PowerPoint open {Host}.exe; for other Office apps, open the document
+  // for Outlook, open Outlook.exe; for other Office apps, open the document
   let path: string;
   if (manifest.manifestType === ManifestType.JSON && app == OfficeApp.Outlook) {
     const version: string | undefined = await getOutlookVersion();
@@ -377,12 +377,10 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
         `The current version of Outlook does not support sideload. Please use version 16.0.13709 or greater.`
       );
     }
-    path = await getOutlookExePath();
+    await launchApp(app, await getOutlookExePath());
   } else {
-    path = await generateSideloadFile(app, manifest, document);
+    await launchApp(app, await generateSideloadFile(app, manifest, document));
   }
-
-  await launchApp(app, path);
 }
 
 async function launchWebApp(app: OfficeApp, manifestPath: string, manifest: ManifestInfo, document: string) {
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -254,6 +254,23 @@ async function getOutlookExePath(): Promise<string> {
   }
 }
 
+async function getWXPExePath(app: OfficeApp): Promise<string> {
+  try {
+    const InstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\${app}.EXE`;
+    const key = new registry.RegistryKey(`${InstallPathRegistryKey}`);
+    const ExePath: string | undefined = await registry.getStringValue(key, "");
+
+    if (!ExePath) {
+      throw new Error(".exe registry empty");
+    }
+
+    return ExePath;
+  } catch (err) {
+    const errorMessage: string = `Unable to find install location: \n${err}`;
+    throw new Error(errorMessage);
+  }
+}
+
 /**
  * Given a file path, returns a unique file path where the file doesn't exist by
  * appending a period and a numeric suffix, starting from 2.
@@ -338,7 +355,7 @@ export async function sideloadAddIn(
     switch (appType) {
       case AppType.Desktop:
         await registerAddIn(manifestPath, registration);
-        await launchDesktopApp(app, manifest, document);
+        await launchDesktopApp(app, manifestPath,manifest, document);
         break;
       case AppType.Web: {
         if (!document) {
@@ -357,7 +374,7 @@ export async function sideloadAddIn(
   }
 }
 
-async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document?: string) {
+async function launchDesktopApp(app: OfficeApp, manifestPath: string, manifest: ManifestInfo, document?: string) {
   if (!isSideloadingSupportedForDesktopHost(app)) {
     throw new ExpectedError(`Sideload to the ${getOfficeAppName(app)} app is not supported.`);
   }
@@ -371,7 +388,11 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
       );
     }
     await launchApp(app, await getOutlookExePath());
-  } else {
+  } else if (manifestPath.endsWith(".json"))
+  {
+    await launchApp(app, await getWXPExePath(app));
+  } else
+  {
     await launchApp(app, await generateSideloadFile(app, manifest, document));
   }
 }
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -255,6 +255,7 @@ async function getOfficeExePath(app: OfficeApp): Promise<string> {
         break;
       default:
         hostApp = "OUTLOOK.EXE";
+        break;
     }
 
     const InstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\${hostApp}`;
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -369,7 +369,7 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
   }
 
   // for Outlook, open Outlook.exe; for other Office apps, open the document
-  if (manifest.manifestType === ManifestType.JSON && app == OfficeApp.Outlook) {
+  if (app == OfficeApp.Outlook) {
     const version: string | undefined = await getOutlookVersion();
     if (version && !hasOfficeVersion("16.0.13709", version)) {
       throw new ExpectedError(
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -255,8 +255,7 @@ async function getOfficeExePath(app: OfficeApp): Promise<string> {
         hostApp = "powerpnt.exe";
         break;
       default:
-        hostApp = "OUTLOOK.EXE";
-        break;
+        throw new Error("The Office host not supported.");
     }
 
     const InstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\${hostApp}`;
@@ -383,7 +382,7 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
 
   // for Outlook, Word, Excel, PowerPoint open {Host}.exe; for other Office apps, open the document
   let path: string;
-  if (app == OfficeApp.Outlook) {
+  if (manifest.manifestType === ManifestType.JSON && app == OfficeApp.Outlook) {
     const version: string | undefined = await getOutlookVersion();
     if (version && !hasOfficeVersion("16.0.13709", version)) {
       throw new ExpectedError(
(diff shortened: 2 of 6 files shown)


packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -254,6 +254,23 @@ async function getOutlookExePath(): Promise<string> {
   }
 }
 
+async function getWXPExePath(app: OfficeApp): Promise<string> {
```
millerds, line 257: This is almost identical to the "getOutlookExePath" method.  I can see the difference is this has been generalized from that which is good, but we should have one method to launch exe files and use that in all cases rather on separate methods for separate cases.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -254,6 +254,23 @@ async function getOutlookExePath(): Promise<string> {
...
+  try {
+    const InstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\${app}.EXE`;
+    const key = new registry.RegistryKey(`${InstallPathRegistryKey}`);
+    const ExePath: string | undefined = await registry.getStringValue(key, "");
+
+    if (!ExePath) {
+      throw new Error(".exe registry empty");
```
millerds, line 264: message should include app name

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -254,6 +254,23 @@ async function getOutlookExePath(): Promise<string> {
...
+    }
+
+    return ExePath;
+  } catch (err) {
+    const errorMessage: string = `Unable to find install location: \n${err}`;
```
millerds, line 269: Message should include app name.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -254,6 +254,23 @@ async function getOutlookExePath(): Promise<string> {
   }
 }
 
+async function getWXPExePath(app: OfficeApp): Promise<string> {
+  try {
+    const InstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\${app}.EXE`;
```
millerds, line 259: Note that the exe names are not the same as the "app" name (Word is actually 'winword.exe').  Case sensitivity would give you problems as well ('excel.exe' != "Excel.exe".

You'll need some sort of mapping for that.  Maybe that means making a map in one of the package were the value is an object with all the information for a given host. . . like host title, exe name, etc.  It's something to think about beyond a switch just for this method.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -357,7 +374,7 @@ export async function sideloadAddIn(
   }
 }
 
-async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document?: string) {
+async function launchDesktopApp(app: OfficeApp, manifestPath: string, manifest: ManifestInfo, document?: string) {
```
millerds, line 377: Instead of adding another argument to this method, we should add a manifestType and/or path property to ManifestInfo.  We shouldn't have to keep both pieces of information everywhere when keeping it together makes more sense.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -371,7 +388,11 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
       );
     }
     await launchApp(app, await getOutlookExePath());
-  } else {
+  } else if (manifestPath.endsWith(".json"))
+  {
```
millerds, line 392: convention is this should be on the previous line.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -371,7 +388,11 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
...
+    await launchApp(app, await getWXPExePath(app));
+  } else
+  {
```
millerds, line 395: convention is that this should be on the previous like with the else statement (like it was before your change)

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -371,7 +388,11 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
       );
     }
     await launchApp(app, await getOutlookExePath());
-  } else {
+  } else if (manifestPath.endsWith(".json"))
```
millerds, line 391: There a couple of things to consider here
- Combine this with the outlook case
- Give that outlook xml manifests are also getting sideloaded via the service now that check is probably not needed anymore
- What is the xml sideloading story for WXP going forward (Outlook had to move to deploying xml via the service)?

millerds (changes requested):
Thanks for taking on this update . . . lets make it a bit more general.

packages/office-addin-dev-settings/src/sideload.ts
QuanfuXiao: Hi Darren, this draft changes I made here is for the purpose of making minimum changes to existing pipeline first. According to your suggestions, we will refactor and combine Outlook cases. For the sideload of XML, I saw a TAOS which can sideload xml (https://titles.msit.mos.microsoft.com/dev/v1/users/packages/addins), you mean we can change the sideload of xml from embedded "web-extension" to TAOS now?

packages/office-addin-dev-settings/src/sideload.ts
QuanfuXiao: Ok, will change.

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Got it! Thansk!

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: You made a good point! Thanks Darren!

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Got it! Thanks!

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -237,36 +237,34 @@ async function getOutlookVersion(): Promise<string | undefined> {
   }
 }
 
-async function getOutlookExePath(): Promise<string> {
+async function getWXPOExePath(app: OfficeApp): Promise<string> {
+  let HostApp: string = "";
```
millerds, line 241: JavaScript convention is lowercase first letter variable names.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -237,36 +237,34 @@ async function getOutlookVersion(): Promise<string | undefined> {
   }
 }
 
-async function getOutlookExePath(): Promise<string> {
+async function getWXPOExePath(app: OfficeApp): Promise<string> {
```
millerds, line 240: Don't use "wxpo" in the method name . . . it should be a general name for the general thing that it is doing (we don't want to have to change the name when there's an update)

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -237,19 +237,34 @@ async function getOutlookVersion(): Promise<string | undefined> {
...
   try {
-    const OutlookInstallPathRegistryKey: string = `HKEY_LOCAL_MACHINE\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\App Paths\\OUTLOOK.EXE`;
-    const key = new registry.RegistryKey(`${OutlookInstallPathRegistryKey}`);
-    const outlookExePath: string | undefined = await registry.getStringValue(key, "");
-
-    if (!outlookExePath) {
-      throw new Error("Outlook.exe registry empty");
+    switch (app) {
+      case OfficeApp.Excel:
+        HostApp = "EXCEL";
```
millerds, line 245: Assuming the registry look up is case sensitive I don't believe this (or the others that are not outlook) are correct.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -237,19 +237,34 @@ async function getOutlookVersion(): Promise<string | undefined> {
...
+        break;
+      case OfficeApp.Outlook:
+        HostApp = "OUTLOOK";
+        break;
+      case OfficeApp.Word:
+        HostApp = "WINWORD";
+        break;
+      case OfficeApp.PowerPoint:
+        HostApp = "POWERPNT";
+        break;
```
millerds, line 255: Should always have a default case in the switch.

packages/office-addin-dev-settings/src/sideload.ts
millerds: I'm saying that was done because Outlook needed it . . . but the Outlook case for xml was already different for xml.  Outlook doesn't have a file to open . . . so the add-ins were actually referenced in exchange and Outlook got them from there . . . so sideloading was to exchange.  Syncing exchange with the new service (which supposed to be the new source of truth) was delayed which didn't work for the developer experience.  So they added a sideload xml manifest directly to the new service instead of indirectly through exchange (and Outlook was going to remove the call to get add-ins from exchange anyway).

I'm not sure what the plan is for WXP.  They have add-in references in files (by id, not be manifest itself).  The hosts are able to load xml manifests locally via the registry.  Json manifests cannot be loaded locally because the hosts need the app definition instead of the direct manifest so they will talk to the new service for that.  

Will WXP support file references for json manifest add-ins?  I don't know why they would remove that support.  If it's still there is this change even needed (loading can still be triggered via file)?

Will WXP remove the xml manifest loading from the registry when they switch to loading from the new service?  Do so would require sideloading xml to the new service, but that is not what this change is targeting.

millerds (changes requested):

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Got it. I will change it. Thanks!

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Got it. I will change it to "getOfficeExePath". How about do you think?

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Yes, I will double check it in registry Editor. Thanks Darren!

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Thanks for Darren's heads up!

YueLi-MSFT: I have tested the registryKey, it can lunch WXPO hosts correctly.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -361,18 +380,23 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
...
     const version: string | undefined = await getOutlookVersion();
     if (version && !hasOfficeVersion("16.0.13709", version)) {
       throw new ExpectedError(
         `The current version of Outlook does not support sideload. Please use version 16.0.13709 or greater.`
       );
     }
-    await launchApp(app, await getOutlookExePath());
+    path = await getOfficeExePath(app);
+  } else if (manifest.manifestType === ManifestType.JSON) {
+    path = await getOfficeExePath(app);
```
akrantz, line 394: Is this changing things to launch Word/Excel/PPT apps based on the exe path rather than opening a document with a reference to the add-in?

packages/office-addin-dev-settings/src/dev-settings-windows.ts
```diff
@@ -178,7 +178,7 @@ export async function registerAddIn(manifestPath: string, registration?: string)
   const appsInManifest = getOfficeAppsForManifestHosts(manifest.hosts);
 
   // Register using the service
-  if (appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
+  if (manifestPath.endsWith(".json") || appsInManifest.indexOf(OfficeApp.Outlook) >= 0) {
```
akrantz, line 181: If you have the manifest info which contains the manifest type, it might be better to use that rather than just checking the extension. 

What happens for Mac with JSON manifests? Is this only being implemented on Windows now, or does Mac not also need changes?

akrantz: It might be better to add an optional parameter for this function about how the registration should be done, e.g. `registerUsingService: boolean` and the caller can have the logic about whether to use the service based on manifest type or whether Outlook or other conditions. The code here could error if Outlook and registerUsingService is false. For undefined, it would treat the value as false. But later the default could change to true unless explicitly false.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -361,18 +380,23 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
     throw new ExpectedError(`Sideload to the ${getOfficeAppName(app)} app is not supported.`);
   }
 
-  // for Outlook, open Outlook.exe; for other Office apps, open the document
+  // for Outlook, Word, Excel, PowerPoint open {Host}.exe; for other Office apps, open the document
+  let path: string;
   if (app == OfficeApp.Outlook) {
```
millerds, line 385: This should be combined with the if manifestType = json

millerds (changes requested):
Some of this would be unnecessary given WXP can load add-in using the reference in the file rather than needing to launch the host separately.  This is different from Outlook.  Is there a plan to remove loading via file reference?

Note: We need to understand the long term plan for xml sideloading in WXP before we make any changes to that code path.  So far that seems contained, but we also need to consider the whole goal when making these more targeted changes.

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Got it. Will change.

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: Yes. Will change.

packages/office-addin-dev-settings/src/sideload.ts
YueLi-MSFT: For XML add-ins we don't change the launching pattern in our changes, it is still a document with add-in embedded via web-extension.

packages/office-addin-dev-settings/src/dev-settings-windows.ts
YueLi-MSFT: Yes, I have change it to use the ManifestInfo object(manifestType). Thanks for your heads up!

packages/office-addin-dev-settings/src/sideload.ts
millerds: That doesn't answer the question of do we need to change the WXP launching pattern at all (even for json)?

millerds:
We still need to addressed the larger question of why the pattern for launch the hosts for WXP needs to change.  Adding an add-in reference to a file and then opening the file (which automatically launches the host) should sill work for add-ins that use a json manifest shouldn't it?  Why do we need to launch the host separately?

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -361,18 +368,21 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
     throw new ExpectedError(`Sideload to the ${getOfficeAppName(app)} app is not supported.`);
   }
 
-  // for Outlook, open Outlook.exe; for other Office apps, open the document
-  if (app == OfficeApp.Outlook) {
+  // for Outlook, Word, Excel, PowerPoint open {Host}.exe; for other Office apps, open the document
```
akrantz, line 371: This comment should be changed back to what was originally there.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -79,18 +80,24 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
...
 
     templateZip.getEntries().forEach(function (entry) {
       var data: Buffer = entry.getData();
-      if (entry == extEntry) {
+      if (entry == extEntry && manifest.manifestType == ManifestType.XML) {
         data = Buffer.from(webExtensionXml);
       }
       outZip.addFile(entry.entryName, data, entry.comment, entry.attr);
+      
+      const webExtensionFolderPath = webExtensionPath.replace(/(\/[^\/]*)$/, '');
```
akrantz, line 96: Is this removing a trailing slash or backslash? The regex is hard to understand with the escaping. Perhaps a comment to explain what it is supposed to be doing would help.

millerds: That or do something like webExtensionPath.substr(0, webExtensionPath.lastIndexOf("/"))

millerds (changes requested):

packages/office-addin-dev-settings/src/sideload.ts
QuanfuXiao: OK, changed.

packages/office-addin-dev-settings/src/sideload.ts
QuanfuXiao: Have replaced it with this method.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -79,18 +80,24 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
...
     }
-
+    
     const webExtensionXml = templateZip
-      .readAsText(extEntry)
-      .replace(/00000000-0000-0000-0000-000000000000/g, manifest.id)
-      .replace(/1.0.0.0/g, manifest.version);
+    .readAsText(extEntry)
+    .replace(/00000000-0000-0000-0000-000000000000/g, manifest.id)
+    .replace(/1.0.0.0/g, manifest.version);
```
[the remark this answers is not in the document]
QuanfuXiao: updated

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -361,18 +368,21 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
...
       );
     }
-    await launchApp(app, await getOutlookExePath());
+    path = await getOutlookExePath();
   } else {
-    await launchApp(app, await generateSideloadFile(app, manifest, document));
+    path = await generateSideloadFile(app, manifest, document);
   }
+
+  await launchApp(app, path);
```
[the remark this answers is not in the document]
QuanfuXiao: updated

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -79,18 +80,24 @@ export async function generateSideloadFile(app: OfficeApp, manifest: ManifestInf
...
+      if (entry == extEntry && manifest.manifestType == ManifestType.XML) {
         data = Buffer.from(webExtensionXml);
       }
       outZip.addFile(entry.entryName, data, entry.comment, entry.attr);
+      
+      const webExtensionFolderPath = webExtensionPath.replace(/(\/[^\/]*)$/, '');
+      // If manifestType is JSON, remove the web extension folder
+      if (entry.entryName.startsWith(webExtensionFolderPath) && manifest.manifestType == ManifestType.JSON) {
+        outZip.deleteFile(entry.entryName);
+      }
```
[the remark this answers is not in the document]
QuanfuXiao: Hi @millerds , there will show an error if we leave it here. Because of original web-extension, a fake add-in is auto launched with errors. So, we have to remove web extension from the zip.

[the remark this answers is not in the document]
millerds: Makes sense (I didn't know iif the host would throw an error).

You should still use the if clause I suggested instead of adding the entry only to delete it later.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -362,7 +369,8 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
   }
 
   // for Outlook, open Outlook.exe; for other Office apps, open the document
-  if (app == OfficeApp.Outlook) {
+  let path: string;
```
millerds, line 372: Path is unused.  Should remove it.

packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -362,7 +369,8 @@ async function launchDesktopApp(app: OfficeApp, manifest: ManifestInfo, document
...
+  if (manifest.manifestType === ManifestType.JSON && app == OfficeApp.Outlook) {
```
millerds, line 373: This should revert back to the original.

millerds (changes requested):

packages/office-addin-dev-settings/src/sideload.ts
QuanfuXiao: changed

packages/office-addin-dev-settings/src/sideload.ts
QuanfuXiao: changed

packages/office-addin-dev-settings/src/sideload.ts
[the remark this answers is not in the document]
QuanfuXiao: Still the same error. Have to remove it.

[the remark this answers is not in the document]
millerds: I see now . . . it's removing the other file in that directory and not just the data one.  In that case then webExtensionFolderPath should be defined outside of the foreach statement and the if clause be around the outZip.addFile statement with the logic reversed (i.e. add if not json or not folder path.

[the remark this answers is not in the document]
QuanfuXiao: Thanks @millerds , updated

millerds (dismissed):

YueLi-MSFT (approved):

millerds:
@QuanfuXiao @YueLi-MSFT looks like there's a test that needs fixing.

YueLi-MSFT (approved):

Issue #840, opened 2024-02-22
Issues with AppContainer Loopback Logic in `office-addin-debugging` and `office-addin-dev-settings`


ZYUN-MSFT:
Hi all, I recently found some issues/strange behavior with the loopback logic.

# Background

I know that AppContainer tech make some Apps likes UWP can't access localhost (loopback) by default.
In previously, I think Webview2 && Edge need that loopback exempt for Add-in side-load.
But some test I did recently shows that things may be different.

**Tools I used：**

* AppContainer Loopback Exemption Utility, which is build in with Fiddler Classic.
* `CheckNetIsolation.exe`

# Issue 1: Type `No` during Loopback question prompt

* OS: Win11 23H2

[Code](https://github.com/OfficeDev/Office-Addin-Scripts/blob/046cf733ee7291e33bc973d912592242932a8e0c/packages/office-addin-debugging/src/start.ts#L309) in `office-addin-debugging` dosen't use the result from [`ensureLoopbackIsEnabled`](https://github.com/OfficeDev/Office-Addin-Scripts/blob/046cf733ee7291e33bc973d912592242932a8e0c/packages/office-addin-dev-settings/src/appcontainer.ts#L125):

```ts
// enable loopback for Edge
if (isWindowsPlatform && parseInt(os.release(), 10) === 10) {
  const name = isDesktopAppType ? "EdgeWebView" : "EdgeWebBrowser";
  await devSettings.ensureLoopbackIsEnabled(name);
}
```

```ts
export async function ensureLoopbackIsEnabled(
  manifestPath: string,
  askForConfirmation: boolean = true
): Promise<boolean> {
  const name = await getAppcontainerNameFromManifestPath(manifestPath);
  let isEnabled = await isLoopbackExemptionForAppcontainer(name);

  if (!isEnabled) {
    if (!askForConfirmation || (await getUserConfirmation(getDisplayNameFromManifestPath(manifestPath)))) {
      await addLoopbackExemptionForAppcontainer(name);
      isEnabled = true;
    }
  }

  return isEnabled;  // if user type No, this will return false
}
```

**So, when `ensureLoopbackIsEnabled` detect that loopback is not exempted and user type `No`, the logic in `office-addin-debugging` will carried on the following steps (start the dev server and side load the add-in).**

**_Then the things get interested:_**
1. I first delete all loopback exempt
[...]

PR #847, opened 2024-03-28
[patch] Fix when add loopback exemption failed will break the debug process

ZYUN-MSFT:
# Change Description:

This PR fix the issue when add loopback exemption failed, the debug start process will totally break.\
The behavior will be the same as when user type `No` for add loopback exempt question.

Detail in this [issue 840](https://github.com/OfficeDev/Office-Addin-Scripts/issues/840)

## Change QA

No.
Maybe No.

I search the docs and found the following articles:
* [Manually sideload an add-in to Office on the web](https://learn.microsoft.com/en-us/office/dev/add-ins/testing/sideload-office-add-ins-for-testing#manually-sideload-an-add-in-to-office-on-the-web)
* [Start the Excel, PowerPoint, or Word add-in project](https://learn.microsoft.com/en-us/office/dev/add-ins/develop/debug-office-add-ins-in-visual-studio#start-the-excel-powerpoint-or-word-add-in-project)
* [Can't open add-in from localhost: Use a local loopback exemption](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-troubleshooting#cant-open-add-in-from-localhost-use-a-local-loopback-exemption)
* [When the add-in tries to open, get "ADD-IN ERROR We can't open this add-in from the localhost" error](https://learn.microsoft.com/en-us/office/dev/add-ins/concepts/browsers-used-by-office-web-add-ins#when-the-add-in-tries-to-open-get-add-in-error-we-cant-open-this-add-in-from-the-localhost-error)

And accroding to [issue 840](https://github.com/OfficeDev/Office-Addin-Scripts/issues/840), I assume that this PR won't change the docs.\
\+ @Rick-Kirkham for verify.

## Validation/testing performed:

1. Disable all loopback with fiddler tools
2. **_Code change_**, **_build_** and **_global link_** the package `office-addin-dev-settings`
3. Create a project with `npx --package yo --package generator-office -- yo office --skip-install`
4. In this project, run the `npm i` and `npm link office-addin-dev-settings`
5. Run `npm run start`
6. In loopback question, type `Y`
7. See following:
npm run start

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -306,7 +306,18 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
     // enable loopback for Edge
     if (isWindowsPlatform && parseInt(os.release(), 10) === 10) {
       const name = isDesktopAppType ? "EdgeWebView" : "EdgeWebBrowser";
-      await devSettings.ensureLoopbackIsEnabled(name);
+      try {
+        await devSettings.ensureLoopbackIsEnabled(name);
+      } catch (err: any) {
+        // if add loopback exemption failed, report the error then continue
+        console.error(err)
+        console.warn("Failed to add loopback exemption.\nTry to sideload the Office Add-in without loopback exemption.\nThe Office Add-in may not load correctly.\n")
+        usageDataObject.reportException("startDebugging()", err, {
+          app: app,
+          document: document,
+          appType: appType,
+        });
+      }
     }
 
     // enable debugging
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -311,7 +311,7 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
       } catch (err: any) {
         // if add loopback exemption failed, report the error then continue
         console.error(err)
-        console.warn("Failed to add loopback exemption.\nTry to sideload the Office Add-in without loopback exemption.\nThe Office Add-in may not load correctly.\n")
+        console.warn("Failed to add loopback exemption.\nWill try to sideload the Office Add-in without the loopback exemption, but it might not load correctly from localhost.\n")
         usageDataObject.reportException("startDebugging()", err, {
           app: app,
           document: document,


millerds (changes requested):
An interesting proposal and based on the documentation links it's something to consider, but it needs to be more comprehensive.

ZYUN-MSFT:
Hi @millerds ,

I push the new change in this PR, which may resolve your concern.

---

*The result of [`devSettings.ensureLoopbackIsEnabled`:](https://github.com/OfficeDev/Office-Addin-Scripts/blob/bea5f448089df377d0a421d873aa18356364978c/packages/office-addin-dev-settings/src/appcontainer.ts#L125)*
* **`true`**: 
    * could means check ok (exemption already exist)
    * could means add exemption ok
* **`false`**: user choice not to add exemption
* **throw error**: add exemption failed

_And for following logic:_
* **For `true` and `false` condition**: the start debug will go on the following logic.
* **For condition throw error**: the original logic will stop and break out, then report the telemetry.

---

In my new commit, I move the logic from `office-addin-dev-settings` to `office-addin-debugging`.\
For condition throw error, the logic will log the original error and warn that add-in load maybe not correctly.\
Of course the add exemption failed will be record telemetry.

new result

---

**_Please comments if you have any other concern._** 😃

packages/office-addin-debugging/src/start.ts
```diff
@@ -306,7 +306,18 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
     // enable loopback for Edge
     if (isWindowsPlatform && parseInt(os.release(), 10) === 10) {
       const name = isDesktopAppType ? "EdgeWebView" : "EdgeWebBrowser";
-      await devSettings.ensureLoopbackIsEnabled(name);
+      try {
+        await devSettings.ensureLoopbackIsEnabled(name);
+      } catch (err: any) {
+        // if add loopback exemption failed, report the error then continue
+        console.error(err)
+        console.warn("Failed to add loopback exemption.\nTry to sideload the Office Add-in without loopback exemption.\nThe Office Add-in may not load correctly.\n")
```
millerds, line 314: Just a slight change to the message . . . "Failed to add loopback exemption.\nWill try to sideload the Office Add-in without the loopback exemption, but it might not load correctly from localhost.\n"

millerds (changes requested):
Just need a slight change.

I still think it would be good to find the telemetry relating to the add-in being able to load and tracking if there is an increase in the can't load from localhost error message.

Rick-Kirkham:
I don't think it has any implications for documentation.

ZYUN-MSFT:
More thoughts on telemetry:

---

According to https://learn.microsoft.com/en-us/office/dev/add-ins/testing/sideload-office-add-ins-for-testing#manually-sideload-an-add-in-to-office-on-the-web the loopback exemption is necessary only for Legacy Edge (beacuse Legacey Edge is a totally UWP).

image

https://github.com/OfficeDev/Office-Addin-Scripts/issues/840 also support the above conclusion.

---

For Webview2, it doesn't need loopback exemption. But original debug logic not only break Webview2 User's experience but also report an unnesaccery telemetry.\
The exception telemetry of `Error: Access Denied, run the command as an administrator` bring by webview2 user actually means nothing to us.

There will be fewer and fewer legacy edge users in the future. So the information bring by the count of `Error: Access Denied, run the command as an administrator` would not carry out a meaningful analysis conclusion.

millerds:
> More thoughts on telemetry:
> 
> According to https://learn.microsoft.com/en-us/office/dev/add-ins/testing/sideload-office-add-ins-for-testing#manually-sideload-an-add-in-to-office-on-the-web the loopback exemption is necessary only for Legacy Edge (beacuse Legacey Edge is a totally UWP).
> 
> image
> 
> #840 also support the above conclusion.
> 
> For Webview2, it doesn't need loopback exemption. But original debug logic not only break Webview2 User's experience but also report an unnesaccery telemetry. The exception telemetry of `Error: Access Denied, run the command as an administrator` bring by webview2 user actually means nothing to us.
> 
> There will be fewer and fewer legacy edge users in the future. So the information bring by the count of `Error: Access Denied, run the command as an administrator` would not carry out a meaningful analysis conclusion.

When we are running this command we don't know what webview the host will be using so that doesn't play a factor.  Error coming from our tooling are always important because they tell us what users are facing.  Documentation tells us what is expected . . . but it doesn't tell us what customers are experiencing (does it match what our documentation says).

But my main point on telemetry is not whether or not the loop bypass hit an error . . . but whether or not the add-in had trouble loading (which is outside the tooling . . . so I'm not asking for logging, but rather knowledge of some other logs).  If we can correlate add-in loading issues (or lack there of) with the bypass errors then we can get a better picture of what kind and how much of an impact this change has and it would be for useful in determining whether we can remove this this automated check.  I want to understand the big picture . . . not this one use case.

millerds (approved):

PR #848, opened 2024-04-04
[minor] office-addin-lint: update package versions

akrantz:
Update the package versions for office-addin-lint.

For context, I was working on a project derived from Office-Addin-TaskPane-React template which involves custom functions, and was running into strange errors from eslint in TypeScript code in VS Code. I noticed that eslint was very out of date. After updating these package versions, building Office-Addin-Scripts and linking to office-addin-lint from the project, it fixed the errors.

NOTE: I removed engines-strict in .npmrc so I could use Node LTS. 

I always wanted to mention that I think there should be a process to regularly update packages to newer versions across the repository. Relying only on dependabot or making updates only for addressing vulnerabilities isn't really enough. It's better to stay current with TypeScript, eslint, react, etc.

No

No

I validated this in the project I was working on. Running the Dev Server in VS Code did not show any errors. 
I didn't get any lint errors in the project after linking to the new version of office-addin-lint.
`npm run build` in the project had no errors.
I ran `npm install` and `npm run build` in the root of the repository and there were no errors.
I ran `npm run test` in office-addin-lint project.

I noticed that `npm run prettier` in the template project wasn't working, so I fixed the path and it works now.

### DIFF
--- a/packages/office-addin-lint/src/lint.ts
+++ b/packages/office-addin-lint/src/lint.ts
@@ -11,7 +11,7 @@ const eslintPath = require.resolve("eslint");
 const prettierPath = require.resolve("prettier");
 const eslintDir = path.parse(eslintPath).dir;
 const eslintFilePath = path.resolve(eslintDir, "../bin/eslint.js");
-const prettierFilePath = path.resolve(prettierPath, "../bin-prettier.js");
+const prettierFilePath = path.resolve(prettierPath, "../bin/prettier.cjs");
 const eslintConfigPath = path.resolve(__dirname, "../config/.eslintrc.json");
 const eslintTestConfigPath = path.resolve(__dirname, "../config/.eslintrc.test.json");
 
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
-engine-strict=true
 registry=https://registry.npmjs.org/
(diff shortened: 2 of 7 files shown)


packages/office-addin-lint/src/lint.ts
```diff
@@ -11,7 +11,7 @@ const eslintPath = require.resolve("eslint");
 const prettierPath = require.resolve("prettier");
 const eslintDir = path.parse(eslintPath).dir;
 const eslintFilePath = path.resolve(eslintDir, "../bin/eslint.js");
-const prettierFilePath = path.resolve(prettierPath, "../bin-prettier.js");
+const prettierFilePath = path.resolve(prettierPath, "../bin/prettier.cjs");
```
millerds, line 14: I'm not very familiar with the prettier setup . . . what is effect of this (this is not a file in this package itself)?

akrantz: It's in the prettier code. I suppose you could ask whether the command should use npx instead, but the change here is just to match changes within prettier itself for its code.

millerds: Makes sense . . . I just haven't seen that file used anywhere so I wasn't sure how it was used.

millerds (approved):

PR #851, opened 2024-04-10
[Minor] Custom function input parameter supports CellValue

xai-msft:
This is to support using [cell values](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/excel-data-types-concepts) as custom function parameters.
We generate the corresponding cell value type in functions.json, and we throw errors for unsupported types.
Cell value types in return type will get converted to "any".

    No.

    We are not announcing cell value support here. No for now.

Added tests and ran tests.


packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/functions.js
```diff
@@ -0,0 +1,148 @@
...
+ * @param {Excel.BooleanCellValue[][]} x
+ * @customfunction
+ */
+function testBooleanCellValueMatrixParam(x) {
+    // empty
+}
+
+/**
+ * Test Excel.BooleanCellValue repeating matrix parameter
+ * @param {Excel.BooleanCellValue[][][]} x
```
donlvMSFT, line 115: Just curious when could CF have this kind of array? As data in the worksheet is 2d, we considering user just give input manually?

xai-msft: I didn't know this before either, but apparently [][][] can be used for repeating range parameters. See https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-parameter-options?tabs=javascript .

packages/custom-functions-metadata/test/cases/error-unsupported-input-parameter/functions.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+/**
+ * Test unsupported parameter
```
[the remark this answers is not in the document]
xai-msft: I wanted to give a different error message for unsupported cell values to reduce confusion / to convey the message that "we support cell values, just not this one". However, I can also just not distinguish between these two cases. What do you think?

akrantz:
@xai-msft Has the spec been updated? I would like to better understand the set of changes being made here before reviewing them.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -619,6 +640,14 @@ function getResults(
     type: resultType,
   };
 
+  // We only support cell value types as input and not return type for now.
```
akrantz, line 643: Not sure why you need to give an error if a cell value type is put for a return value. It will still just generate "any" for it since the return type info in the metadata. 

Later it would be good to tighten this, e.g. the linked entity provider function should return `Promise<CellValue>` where CellValue is any cell value type (including error types).

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -138,6 +138,27 @@ const TYPE_CUSTOM_FUNCTION_CANCELABLE = {
...
+// These does not work if the developer is using namespace/type alias. To support that, we'd need to
+// use typeChecker.getFullyQualifiedName, which in turns requires creating a program out of the source file
+// (which we don't do now) and passing the type checker all the way down to getParamType. This would be a
+// larger refactor (e.g., rewriting parseTree into a class) and is tracked as a work item.
+const CELLVALUETYPE_MAPPINGS = {
+  "Excel.CellValue": "cellvalue",
+  "Excel.BooleanCellValue": "booleancellvalue",
+  "Excel.DoubleCellValue": "doublecellvalue",
+  "Excel.EntityCellValue": "entitycellvalue",
+  "Excel.ErrorCellValue": "errorcellvalue",
```
akrantz, line 150: Aren't there other error cell value types? They should all be supported.

It would be nice to have type info which could tell that the type is derived from Excel.CellValue, but currently there is no type info for these types with custom-functions-metadate. If the types are pulled out into an excel.cellvalue.d.ts for the type info, then that could be used by the code here and perhaps have better knowledge about the type hierarchy, or be written where new types could be added and the checks in here can just tell whether the class derives from Excel.CellValue. But this approach based on the list of type names is OK for now.

packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/expected.json
```diff
@@ -0,0 +1,205 @@
+{
+    "allowCustomDataForDataTypeAny": true,
+    "functions": [
+        {
+            "description": "Test Excel.CellValue parameter",
```
akrantz, line 5: There is currently only one successful case which is for Excel.CellValue. Really, there should be a case for each type.

packages/custom-functions-metadata/test/cases/error-unsupported-return-type-cellvalue/functions.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+/**
+ * Cell Value types are not supported as return types
+ * @customfunction
+ */
+function testUnsupportedCellValue(x): Excel.BooleanCellValue {
```
akrantz, line 8: As per my other comment, I think this can/should be relaxed. The function can return a cell value or error cell value type. For the linked entity provider function, it MUST return Promise<Excel.CellValue> since it needs to return the linked entity.

akrantz (changes requested):
This is a good start. Please see my comments. Please also update the spec so it supports the changes.

xai-msft:
> @xai-msft Has the spec been updated? I would like to better understand the set of changes being made here before reviewing them.

Thank you for your comments!
I'd like to sum up a few things first before addressing the comments individually:

1. We do not have return value CellValue support on Excel parsing side, which means if we include any new types in return parameter, the whole add-in registration will fail. We really cannot add them in metadata at this stage.
To support return value means to implement both registration and validation logic, which is a bigger work item on its own. We will do input / return in iterations as we agreed with in the spec discussion.

2. One thing this PR deviates from the spec is that while in the spec I decided to convert unsupported CellValue types (Reference etc.) to "any", I decided in this PR that silently failing is not a good idea. The rest is the same as in the spec. I'll update it to reflect this design change. Thanks for the reminder!

4. Error for unsupported return type: it wasn't specified in the spec since the spec was about input type. For the same reason as above, I think silently failing by converting CellValue types into "any" is not ideal, since it might give dev the wrong impression that it persists and we do validation on Excel like with the rest of the supported types, which we don't. I have to filter it out in the generator since both input and return go through this same utility function. I can update the spec to include the information that metadata generator throws error on CellValue as return type too.

I hope this answers your questions and please let me know if there's further concern!

packages/custom-functions-metadata/test/cases/error-unsupported-return-type-cellvalue/functions.ts
xai-msft: See my other comment. Dev should continue to use Promise<any> at this stage until we actually support return type.

packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/expected.json
xai-msft: Sorry, could you clarify your comment? I have all the supported subtypes of CellValue here. What other type do you mean?

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: We don't support error subtypes now.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Please see my other comment, thanks.

xai-msft:
@akrantz Hi, I've updated the spec. Feel free to let me know what you think of my responses to your comments when you have time. Thank you!

packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/expected.json
akrantz: Each of the *CellValue types.

akrantz:
@xai-msft I ran out of time looking at this right now. Might be good to meet up and discuss rather than going back and forth in comments. Will look more later. Let me know when is a good time to talk.

akrantz:
@xai-msft I don't see the specification.md file added to the PR. Can you check this please?

xai-msft:
> @xai-msft I don't see the specification.md file added to the PR. Can you check this please?

Sorry I misunderstood. I thought you were talking about the spec doc on internal SharePoint that you previously reviewed.

We are not announcing this feature yet since Excel code is still rolling out. I'd like to update the spec here (together with official documentation) after the rollout completes. And I'll create a work item to track the spec update here. Would you be ok with that?

packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/expected.json
xai-msft: I hope I'm not missing something but I have all the *CellValue types here? If one of them fails the whole test will fail.

packages/custom-functions-metadata/test/cases/error-unsupported-return-type-cellvalue/functions.ts
xai-msft: Metadata can contain boolean/string/number as return types. Excel used to do type check on these but it no longer does for some reason.

That said, we had some discussions, and we agree that converting CellValues as return types to "any" is better. I'll update the PR.

xai-msft:
gently pinging @akrantz

To sum up:
1. CellValue type as return type has been changed to converting to "any" as requested
2. We want to postpone specification.md change till officially announcing the feature, since Excel code is still rolling out. 

Please let me know if there's other concerns. Thank you!

MiaofeiWang (approved):

packages/custom-functions-metadata-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "custom-functions-metadata-plugin",
-  "version": "1.6.2",
```
akrantz, line 3: @millerds This version update should be done by Lerna, right? I don't think the version should be manually updated, but cannot quite remember.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
-          return type;
-        }
-        if (array.typeName.getText() !== "Array") {
-          extra.errors.push(logError("Invalid type: " + array.typeName.getText(), typePosition));
-          return type;
-        }
-      }
-      arrayType = getArrayDimensionalityAndType(t);
-      kind = arrayType.type;
+  // Only get type for typescript files. js files will return "any" for all types
```
akrantz, line 1062: When using JavaScript, the user can annotate the type in the @param field, so it should be able to have a type when annotated for JavaScript, and we would expect this to be done for JavaScript in order to know the type.

Perhaps you want to add an error unit test case when a param in JavaScript doesn't specify a type, or perhaps this is the same any actually specifying `any` or `unknown` as the type in TypeScript. Might be useful to have test cases to clarify this.

In terms of the code here, it seems that not having a node implies that the type to return is any, so the code looks fine.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
     }
     // @ts-ignore
-    type = TYPE_MAPPINGS[kind];
-    if (!type) {
-      extra.errors.push(logError("Type doesn't match mappings", typePosition));
+    if (CELLVALUETYPE_MAPPINGS[nodeTypeName]) {
+      // @ts-ignore
+      let cellValue = CELLVALUETYPE_MAPPINGS[nodeTypeName]
+      if (cellValue === "unsupported") {
+        extra.errors.push(logError("Custom function does not support cell value type: " + typeReferenceNode.typeName.getText(), typePosition));
```
akrantz, line 1092: Suggestion, use back-tick string instead of concatenating strings with +.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
+        return type;
+      }
+      return cellValue;
     }
+
+    extra.errors.push(logError("Invalid type: " + typeReferenceNode.typeName.getText(), typePosition));
```
akrantz, line 1098: same suggestion here to use back-tick string.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
+    return type;
+  }
+
+  // @ts-ignore
+  type = TYPE_MAPPINGS[node.kind];
+  if (!type) {
+    extra.errors.push(logError("Type doesn't match mappings", typePosition));
```
akrantz, line 1105: This error message is describing the implementation (using the type mappings array). Please compose an error message meaningful to the user. For example, in ScriptLab, if this error was encountered, what is really the problem to tell the user. Is it that they used a type that isn't supported?

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
+function typeNodeIsArray(node: ts.TypeNode): boolean {
+  // [] case
+  if (ts.isArrayTypeNode(node)) {
+    return true;
+  }
+  // Array<sometype> case
+  if (ts.isTypeReferenceNode(node)) {
+    return (node as ts.TypeReferenceNode).typeName.getText() === "Array";
+  }
+  return false
```
akrantz, line 1125: semicolon is missing at the end of the line.

akrantz: Make sure to run prettier if you're not having prettier applied in VS Code.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
+
 /**
  * Wrapper function which will return the dimensionality and type of the array
  * @param node TypeNode
  */
 function getArrayDimensionalityAndType(node: ts.TypeNode): IArrayType {
   let array: IArrayType = {
     dimensionality: 0,
-    type: ts.SyntaxKind.AnyKeyword,
+    node: node,
```
akrantz, line 1135: You can use put; `node,` since the names are the same.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1029,50 +1055,84 @@ function hasInvocationParameter(
...
+  if (!node) {
+    return type;
+  }
+  const typePosition = getPosition(node);
+  
+  // Get the inner type node if it is an array
+  if (typeNodeIsArray(node)) {
+    let arrayType: IArrayType = {
+      dimensionality: 0,
+      node: node,
```
akrantz, line 1072: You can just put `node,` since the names are the same.

packages/custom-functions-metadata/src/parseTree.ts
akrantz: There are other places like this which can be updated in the same way.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -1083,49 +1143,47 @@ function getArrayDimensionalityAndType(node: ts.TypeNode): IArrayType {
...
     dimensionality: 1,
-    type: ts.SyntaxKind.AnyKeyword,
+    node: node,
   };
 
   let nodeCheck = (node as ts.ArrayTypeNode).elementType;
-  array.type = nodeCheck.kind;
+  array.node = nodeCheck
   while (ts.isArrayTypeNode(nodeCheck)) {
     array.dimensionality++;
```
akrantz, line 1158: nit: I think it's preferrable to put ++ at the beginning.

packages/custom-functions-metadata/test/cases/error-unsupported-input-parameter/expected.ts.errors.txt
```diff
@@ -0,0 +1 @@
+Invalid type: Excel.Worksheet (8,40)
```
akrantz, line 1: Rather than the message being `Invalid type`, it might be better to say that the type is not allowed for a custom function parameter.

akrantz (approved):
I've made some suggestions which I think would be good to make. I will mark as Approved so I don't block completion.

packages/custom-functions-metadata-plugin/package.json
millerds: Correct . . . do not update the version, that will be done as part of the publish process.

packages/custom-functions-metadata/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "custom-functions-metadata",
-  "version": "1.6.2",
```
millerds, line 3: do not update the version, that will be done as part of the publish process.

packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/functions.js
akrantz: https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-parameter-options?tabs=javascript#declaring-repeating-parameters

Does custom-functions-metadata support repeating scalar or matrix parameters?  The help mentions hand-writing the functions.json metadata to have `repeating: true`. It would be good to support it so it can be determined from the dimensionality of the array type, e.g. `number[][][]`. This would not be for this PR but something else to add to the backlog.

akrantz: It is supported.  

https://github.com/OfficeDev/Office-Addin-Scripts/blob/fb1b94d0e3654680aa67088d098f4719e94484cd/packages/custom-functions-metadata/test/cases/repeating-param-3d/functions.ts#L5

packages/custom-functions-metadata-plugin/package.json
xai-msft: Reverted.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Resolving as no issue present.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Changed.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Hi, I moved the code around a bit for readability, but this error message was created by someone else. I'm not sure if I understand what this is for, or if the code ever hits, and I prefer not changing it for safety.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Thanks. I ran prettier this iteration.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Thanks. I didn't know this syntax!

packages/custom-functions-metadata/package.json
xai-msft: reverted

akrantz (approved):
Thanks for making the updates!

xai-msft:
> Thanks for making the updates!

Thanks! Could you help me run `/azurepipelines run` ?

PR #863, opened 2024-05-09
User/adrians/add expiration date to sso

Adrian-MSFT:

---

**Change Description**:

    Updated the office-addin-sso configure command to include an optional parameter [secret-ttl] that takes the time to live for the secret used in the Azure app registration, in days. Default is 60 days.

   Added parameter to office-addin-sso configure command.
   [secret-ttl]: takes the time to live for the secret used in the Azure app registration, in days. Default is 60 days.

   [secret-ttl]: takes the time to live for the secret used in the Azure app registration, in days. Default is 60 days.

    Tested with no input for new parameter, bad input (not a number), and a valid input

### DIFF
--- a/packages/office-addin-sso/src/commands.ts
+++ b/packages/office-addin-sso/src/commands.ts
@@ -13,7 +13,7 @@ import inquirer = require("inquirer");
 
 /* global process, console */
 
-export async function configureSSO(manifestPath: string) {
+export async function configureSSO(manifestPath: string, secretTTL?: string) {
   // Check platform and return if not Windows or Mac
   if (process.platform !== "win32" && process.platform !== "darwin") {
     console.log(chalk.yellow(`${process.platform} is not supported. Only Windows and Mac are supported`));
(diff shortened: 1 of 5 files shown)


packages/office-addin-sso/src/commands.ts
```diff
@@ -31,6 +31,16 @@ export async function configureSSO(manifestPath: string) {
     }
   }
 
+  // If no expiration date was provided, or if we fail to parse the number, we default to 60 days
```
millerds, line 34: Looks like if we fail to parse a number we bail instead of defaulting to 60.  I think that makes sense, but the comment doesn't match.

millerds (changes requested):

millerds:
Also make sure you sync master and merge.

packages/office-addin-sso/src/commands.ts
Adrian-MSFT: Good catch, thank you. I will update this comment

millerds (approved):

PR #843, opened 2024-02-28
[minor] Add handling for new @linkedEntityDataProvider and @excludeFromAutoComplete jsdoc tags

jeaniehuynh:
**Change Description**:

    This PR adds handling for the new @linkedEntityDataProvider and @excludeFromAutoComplete metadata tags needed for 3P Linked Entities in Excel.

2. **Do these changes impact *documentation*?** (e.g., a tutorial on https://learn.microsoft.com/office/dev/add-ins/overview/office-add-ins)
    Yes, we need to add documentation for the new tags

If you answered yes to any of these please do the following:
    > Include 'Rick-Kirkham' in the review
    > Make sure the README file is correct

**Validation/testing performed**:

    Wrote new test cases and confirmed they all pass. Will link NPM package to office-js-api to test everything before checking in.

PR #867, opened 2024-06-07
[patch] move cell value type to a new attribute "cellValueType" in custom function metadata

xai-msft:
Cell value types have been introduced to custom function input parameter type. We decided to move it to a new attribute "cellValueType" for backward compatibility, so that older client won't break on unknown types.
"cellValueType" works as follows:
When "type" is one of "boolean"/"number"/"string", "cellValueType" won't be parsed by Excel. When "type" is "any" and "cellValueType" is a valid [cell value type](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/excel-data-types-concepts), the cell value type will be used as custom function parameter type. If "cellValueType" is invalid, Excel will use the fallback "type" "any".

Sample function metadata before change:

{
    "description": "Test Excel.CellValue parameter",
    "id": "TESTCELLVALUEPARAM",
    "name": "TESTCELLVALUEPARAM",
    "parameters": [
        {
            "name": "x",
            "type": "cellvalue"
        }
    ],
    "result": {}
}

after change:

{
    "description": "Test Excel.CellValue parameter",
    "id": "TESTCELLVALUEPARAM",
    "name": "TESTCELLVALUEPARAM",
    "parameters": [
        {
            "name": "x",
            "type": "any"
            "cellValueType": "cellvalue"
        }
    ],
    "result": {}
}

  No

    Cell value types are not public yet.

Validation/testing performed:
Updated and ran tests.


MiaofeiWang (approved):

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -34,6 +34,7 @@ export interface IFunctionParameter {
   name: string;
   description?: string;
   type: string;
+  subtype?: string;
```
akrantz, line 37: Consider something other than "sub" for the name here. It's not really "below". I think a name which is less generic than "type", such as "valueType".

Please provide more background info about the existing types and why there is an issue for backward compatibility. It would help me to understand more and perhaps offer different suggestions here with a better understanding.

akrantz (commented):
The change is simple but I'd like to discuss more about the naming here. I feel like there should be more examples / test cases. This is not really about the change here is what metadata is generated, but really is more about the native code which is parsing and handling the metadata. I wish I had known about this discussion earlier.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Hi Adam,
 
Backward compatibility issue: 
When an add-in adopts the new cell value types and use them in the functions.json, it will break the whole custom function registration pipeline on an older client without parser code change, because it's an unknown Enum. This means that no custom function can be used at all. Whereas putting it in a new attribute will make sure that the same metadata can work on both older and newer clients, and a developer can, for example, run different code in the same function after checking the requirement set. Older client will simply ignore the "subtype" attribute.

"Subtype" is used exactly in this sense: it will only be parsed when the "type" is "any", because it's meant to be a refinement of "any" by specifying which specific subtype it is, and it doesn't make sense to add a "subtype" when the "type" is anything other than "any". In that case it will simply be ignored. When "subtype" is invalid, it will also fallback to "any". Please see PR 3058051 for parser code change and relevant tests.
I think if we use "valueType", it wouldn't be clear what the relation between "type" and "valueType" is. It might be confusing to have two seemingly overlapping types there.

akrantz: Really this is where forward-compatibility should have been better considered for unknown types. The parser should have allowed some form of unknown types to be treated as any. Now you're adding a second "type" field to make it for the limitations of the first. 

It seems to me this is specifically targetting support for Cell Value Types, so how about making it just that: "cellValueType". I suppose that when type is something like "number" that cellValueType should be undefined.

What's going to happen when other types are adding in the future. How do older clients handle the fact that there may be new strings they don't recognize.

I don't like "subType" because this type is in now way "below" or "sub" the "type" field.

xai-msft: Back/forward compatibility: "any" was introduced alongside rich data in 2019, I believe, and since then there had been no major change to cf schema. We have to work with what we have now.

My major concern for "cellValueType" is that it's not clear from the naming what the relation with "type" is, or which one will take precedence over another when they don't agree. "Subtype" solves these issues since "type" will take precedence, as expected.

Cell Value Types encapsulate all possible content types in a cell, by definition. It's unlikely we will need to extend parameter type again since custom function inputs are values in cells.

Please let me know what you think!

akrantz: Well I've already let you know my perspective. Perhaps @AlexJerabek could weigh in about the naming.

AlexJerabek: I'm confused why we're reworking this for backward compatibility. Can't we say the expanded type support is only supported in Excel version X+ and tie it to a requirement set? That wouldn't be different to how we handle many other aspects of the product. It would also mitigate Adam's concern about the problem in the future.

I'm also confused as to how this actually helps developers. Won't they manually have to change their metadata as you did in the test cases? Isn't that then a breaking change? 

Separate from that, I'm with Adam that "sub" is confusing. As a developer, my instinct would be to provide both type and subtype, but in reality I only need subtype for legacy purposes. Maybe `fallbackType` to make it clear this is only needed if type is invalid?

Tagging @alison-mk and @davidchesnut to weigh in too.

AlexJerabek (changes requested):
Corollary to this: aren't there doc changes that'll need to happen? Have those been requested from the content team?

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: Resolving this per discussions via email.

akrantz (approved):
Looks good!

millerds:
@xai-msft . . . Alex's question still needs to be addressed.

AlexJerabek:
@millerds @xai-msft @akrantz
The PR description says these changes aren't public yet. Do we have an idea when they will be? Is there a preview phase? As long as this PR is separate from the feature request, we can loop in @alison-mk separately to document the new `cellValueType` and behavioral changes to `type`.

millerds:
> @millerds @xai-msft @akrantz The PR description says these changes aren't public yet. Do we have an idea when they will be? Is there a preview phase? As long as this PR is separate from the feature request, we can loop in @alison-mk separately to document the new `cellValueType` and behavioral changes to `type`.

This repo does not have separate internal or preview releases branches (we haven't needed one yet).  Once this gets checked in it will be part of the next publish of the packages (which is not on a set time schedule) . . . making it at least preview even if the full support of it is not official yet.

AlexJerabek:
Thanks @millerds! Then I'd like to wait to check this in until we have a minimum set of documentation ready to go in concurrently.  We have a legal requirement to document any public API surface area, so we'll at least need some mention of the new attribute.

@xai-msft and @alison-mk can get that staged in the https://github.com/OfficeDev/office-js-docs-pr repo.

alison-mk:
Thanks, all! I'll reach out to @xai-msft internally to coordinate the docs updates. I started a draft PR here: https://github.com/OfficeDev/office-js-docs-pr/pull/4632. Agreed that we should align the docs update with the feature release if possible.

xai-msft:
Thanks @millerds for the reminder and thank you all for the discussions. I'll work with Alison on documentations and update here when it's ready.

MiaofeiWang:
Hi @AlexJerabek , thanks for the reminder. This feature needs to be included in a new set of requirement "CustomFunctionsRuntime". We are working on the proposal and process of a new requirement set. 

However, this feature was highly asked by project Capricorn and we hope to bring it to LSEG first before GA. So could we check this in first, just like other Capricorn specific features like https://github.com/OfficeDev/Office-Addin-Scripts/pull/843?

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -159,6 +160,19 @@ const CELLVALUETYPE_MAPPINGS = {
   "Excel.ValueTypeNotAvailableCellValue": "unsupported",
 };
 
+const CELLVALUETYPE_TO_BASICTYPE_MAPPINGS = {
+  "cellvalue": "any",
+  "booleancellvalue": "boolean",
+  "doublecellvalue": "number",
+  "entitycellvalue": "any",
+  "errorcellvalue": "any",
+  "formattednumbercellvalue": "any",
```
akrantz, line 169: number?

akrantz (approved):
Made a suggestion about formattednumbercellvalue mapping but looks good to me otherwise. I've actually happen to see this updated to cellValueType and appreciate being receptive to previous feedback that I have given.

packages/custom-functions-metadata/src/parseTree.ts
xai-msft: I think that makes sense.

akrantz:
@AlexJerabek We should go ahead with this change before documentation has been updated. It won't have any impact on what a user would do. This is just to make improvements so Excel know more about the cell value types so it can coerce types to the desired cell value type.

akrantz (approved):

AlexJerabek (approved):

MiaofeiWang:
It seems merging this PR still needs repo admin's help. Would you please help run the status check, @akrantz @millerds ? Thank you.

PR #873, opened 2024-06-24
run audit fix

zhngx1:
- run audit fix
- update ws from 7.5.9 to 7.5.10


millerds:
This looks like it's updating a lot more than just the 'ws' vulnerability.  Only two of the packages have the vulnerability and only they should be updated.

millerds (approved):

PR #875, opened 2024-07-21
[minor] don't enable appcontainer loopback on Windows 11

akrantz:
I have been running into a prompt to enable loopback every time I start debugging on Windows 11. It seems like the appcontainer names for Edge changed, but there is also no need with Windows 11 to enable loopback, as it works when I continue.

This changes the debugging start command to skip enabling loopback when running on Windows 11. It also makes a change that it should enable loopback automatically instead of prompting.

I believe either of these should resolve the issue. I'm glad to discuss what change is actually desired. 

NO

Is loopback documented? It might require a documentation update but not sure.

    Describe manual testing done:
    
    I validated the version check under the debugger and that it skipped the check.

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -304,19 +304,35 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
     }
 
     // enable loopback for Edge
-    if (isWindowsPlatform && parseInt(os.release(), 10) === 10) {
-      const name = isDesktopAppType ? "EdgeWebView" : "EdgeWebBrowser";
-      try {
-        await devSettings.ensureLoopbackIsEnabled(name);
-      } catch (err: any) {
-        // if add loopback exemption failed, report the error then continue
-        console.error(err)
-        console.warn("Failed to add loopback exemption.\nWill try to sideload the Office Add-in without the loopback exemption, but it might not load correctly from localhost.\n")
-        usageDataObject.reportException("startDebugging()", err, {
-          app: app,
-          document: document,
-          appType: appType,
-        });
+    if (isWindowsPlatform) {
+      let enableLoopback = false;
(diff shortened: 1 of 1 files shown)


millerds:
There was a previous discussion about this: https://github.com/OfficeDev/Office-Addin-Scripts/issues/840.

While Win11 changed the name of the containers, I believe it also depends on which webview (IE, v1, or v2) is being used at the time of sideloading that makes is required.  On top of that, enabling loopback is an administrator operation (we call a shell command that requires admin privileges).  The existing pattern was to ask because it is an admin operation (and potential security risk) and once it was enabled, we didn't have to ask anymore.

By not asking, we are essentially making security decision without the user's consent.  On top of that, if they are not running as an administrator when we attempt to call the command, it will still fail.  If they then later on sideload using a webview that requires it in order to work (I don't recall which ones do, but I'm confident webview2 doesn't), then the add-in won't load and there will be no indication as to why.

Updating the container names checked and enabled would be another way to help the situation, but I think that might still be incomplete.  I think it would be nice to validate which webviews (and possibly OS combination) require the loopback enabling, adjust the check to look for that specifically, point out that it requires admin privileges, and verify the behavior with different versions of the host applications (including those that we support that still only use IE webview).  But so far no one has committed to take the time to do that.

akrantz:
@millerds You told me about a workaround for this which I performed and then it was no longer an issue. Was it to run as admin?

Wouldn't this change still help? Would anyone be using legacy Edge or Internet Explorer on Windows 11?  If it is just for a host such as Outlook, perhaps there is way we can skip this requirement for others hosts. Or perhaps we should have some way to mention the workaround which avoids the issue.

hermanwenhe:
@akrantz  @millerds  Just got an issue from customer about lookback issue. https://github.com/OfficeDev/OfficeAddinDevKit/issues/16

RudyCo:
Please re-open #840
This issue is really painful...

PR #895, opened 2024-10-09
[patch] Re-submit #867: Custom function - new field "cellValueType" for input cell value type information 

MiaofeiWang:
This is a duplicate of [PR#867](https://github.com/OfficeDev/Office-Addin-Scripts/pull/867) as the original branch is not able to get update.

---

**Change Description**:

Cell value types have been introduced to custom function input parameter type. We decided to move it to a new attribute "cellValueType" for backward compatibility, so that older client won't break on unknown types.
"cellValueType" works as follows:
When "type" is one of "boolean"/"number"/"string", "cellValueType" won't be parsed by Excel. When "type" is "any" and "cellValueType" is a valid [cell value type](https://learn.microsoft.com/en-us/office/dev/add-ins/excel/excel-data-types-concepts), the cell value type will be used as custom function parameter type. If "cellValueType" is invalid, Excel will use the fallback "type" "any".

Sample function metadata before change:
```
{
    "description": "Test Excel.CellValue parameter",
    "id": "TESTCELLVALUEPARAM",
    "name": "TESTCELLVALUEPARAM",
    "parameters": [
        {
        "name": "x",
        "type": "cellvalue"
        }
    ],
    "result": {}
}
```
after change:
```
{
    "description": "Test Excel.CellValue parameter",
    "id": "TESTCELLVALUEPARAM",
    "name": "TESTCELLVALUEPARAM",
    "parameters": [
        {
        "name": "x",
        "type": "any"
        "cellValueType": "cellvalue"
        }
    ],
    "result": {}
}
```
1. Do these changes impact command syntax of any of the packages? (e.g., add/remove command, add/remove a command parameter, or update required parameters)
No

1. Do these changes impact documentation? (e.g., a tutorial on https://learn.microsoft.com/office/dev/add-ins/overview/office-add-ins)
Cell value types are not public yet.

1. Validation/testing performed:
Updated and ran tests.


donlvMSFT (approved):

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -708,6 +722,13 @@ function getParameters(parameterItem: IGetParametersArguments): IFunctionParamet
         type: ptype,
       };
 
+      // for backward compatibility, we put cell value type in cellValueType instead of type.
+      if (Object.values(CELLVALUETYPE_MAPPINGS).includes(ptype)) {
+        // @ts-ignore
```
akrantz, line 727: Is it possible to not have `@ts-ignore` here. Usually the code could be modified. Can you please look into this?

akrantz (approved):
Looks OK. Suggest seeing if `@ts-ignore` can be removed, but that doesn't block check-in.

PR #896, opened 2024-10-14
[Major] Update project to use pipeline template

millerds:
**Change Description**:
We are required to use a pipeline template as a base for security purposes.  This updates the pipeline files to do so, but this also added the requirement to be able to run the linter at the monorepo root instead of separately for each project which also requires a shared tsconfig file at the root and some package updates.  All this added up to some standardization between project, build changes, and many linter fixes.  Nothing should be functionally different despite the many changes . . . but verification will be needed.

Should do a major version bump to cover the updated tsconfig building which updated some properties for some packages.

No.

No.

Ran the linter, ran the build, ran the automated tests, and used the branch to verify the pipeline changes passed.


Adrian-MSFT:
I think this is ok, however, would it be possible to break up these huge PRs in the future for ease of reviewing? Though, I realize, it may not always be possible

Adrian-MSFT (approved):

PR #898, opened 2024-10-23
[minor] remove path dependency

werreour:

---

## Description

Hi all.

I've removed the "path" dependency because it is provided by node instead of external package. The [original path package](https://www.npmjs.com/package/path) has not been updated in the last 9 years and was probably added by mistake.

While it is both harmless and useless (`require('path')` will return nodejs internal module), it causes unexpected side effects for bundlers, specifically `webpack`, which includes it in the bundle instead of the value provided in [fallback](https://webpack.js.org/configuration/resolve/#resolvefallback), if office addon scripts were added as dependencies to the project.

----

No

No

I have run automated tests via `TEMP=/var/tmp npm run test`.


werreour:

millerds:
Thanks for submitting this PR!

The change includes a lot of package-lock.json file changes unrelated to removing the "path" package from a couple of the packages.  It should be targeted to the specific thing you are changing (this includes the package-lock.json file changes for the packages you are updating).

werreour:
Hi @millerds. Thanks for the prompt reply!

The package lock updates were caused by the `npm install` command and I'm not sure why that happened considering the fact that I didn't change anything in package.json except "path" dep. I probably forgot to do something?

I can try to remove the package from the lockfile in the text editor, but I'm not sure that's the best way to do it.

```bash
node -v
v18.18.2

npm -v
9.8.1
```

millerds:
@werreour Correct . . . you should not try to manually edit package-lock.json file.  Is should be updated by the npm install command.  However, you should run `npm install` inside the folder where the package.json is being updated (packages/office-addin-manifest and packages/office-addin-project in this case).  This should get the correct package-lock.json file generated (I'm not sure why lerna does it differently when run at the root).  Only include the 4 files (the two from each of those packages) in the PR,

werreour:
@millerds
Thank you for clarifications!
I've updated the pull request.

PR #907, opened 2024-11-14
[minor] New tag "@CustomEnum" to generate custom enums for custom functions

MiaofeiWang:
**Change Description**:

This PR supports to generate metadata for custom enums tagged with "@CustomEnum" (case-insensitive). More details see /packages/custom-functions-metadata/Specification.md.

    No.

    Yes. This PR introduces a new tag "@CustomEnum" and we will add an introduction page for the custom enums in https://learn.microsoft.com/en-us/office/dev/add-ins/excel/custom-functions-overview.

All tests under custom-function-metadata pass.


packages/custom-functions-metadata/Specification.md
```diff
@@ -39,6 +40,22 @@ If the last function parameter is of type `CustomFunctions.CancelableInvocation`
 Indicates that the custom function captures the calling object i.e. the [Excel.CellValue](https://learn.microsoft.com/es-es/javascript/api/excel/excel.cellvalue)
 object that invoked the custom function, in the first parameter.
 
+---
+### @customEnum
```
donlvMSFT, line 44: Do we need to specify it's introducing a new usage, same layer tag with @customfunction, and other tags couldn't decorate this tag?

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -177,30 +198,129 @@ export function parseTree(sourceCode: string, sourceFileName: string): IParseTre
...
+    const extra: IFunctionExtras = {
+      errors: [],
+      javascriptFunctionName: "",
+    };
+
+    const id = enumDeclaration.name.text;
+    validateId(id, getPosition(enumDeclaration), extra);
+    extras.push(extra);
+
+    if (checkForDuplicate(metadataEnumIds, id)) {
```
donlvMSFT, line 256: Just to confirm, an Enum id and a Function name could be same, is that correct?

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -177,30 +198,129 @@ export function parseTree(sourceCode: string, sourceFileName: string): IParseTre
...
+
+      jsDocType = typeMatch? typeMatch[1].trim() : null;
+    }
+
+    // Get the first member of the enum to determine if it is a number or string enum
+    const firstMember = enumDeclaration.members[0];
+    if (firstMember.initializer) {
+      const initializerText = firstMember.initializer.getText();
+      isNumberEnum = !isNaN(Number(initializerText));
+      if (jsDocType) {
```
donlvMSFT, line 283: Do we want this to have an `else`?
Should handle if `jsDocType == null`, and `isNumberEnum == true`, in this case `jsDocType` should be set to `number` ?

MiaofeiWang: That is unnecessary because jsDocType won't be used anymore after this `if`. 
`isNumberEnum` is the flag in following logic and that is just why the type in JsDoc tag is optional.

packages/custom-functions-metadata/Specification.md
MiaofeiWang: Good point. I should and will separate the `@customEnum` with the tags for functions.

donlvMSFT (dismissed):

MiaofeiWang:
Hi @millerds , could you please help review this PR? Thank you.

millerds (approved):

akrantz (changes requested):
I'd like to take a look. I'll update the status soon.

packages/custom-functions-metadata/Specification.md
```diff
@@ -11,6 +11,7 @@ The function parameter types may be provided using the [@param](#param) tag in J
 ## Tags
 * [@cancelable](#cancelable)
 * [@capturesCallingObject](#capturescallingobject)
+* [@customEnum](#customenum) _{type}_
```
akrantz, line 14: nit: Other lines don't use italic or curly brace around the parameter. It would be nice for them all to be consistent. I think italic without curly brace may be enough.

packages/custom-functions-metadata/Specification.md
```diff
@@ -195,3 +196,23 @@ A function can return a Promise, which will provide the value when the promise i
...
 Any other type will be treated as an error.
+
+---
+## Custom enums
+
+The `@customEnum` JSDoc tag can be used in TypeScript to mark an enum as a custom enum. When a custom function uses this custom enum as the type for its arguments, the enum values will be displayed in a dropdown list for the end user when they enter the function.
+
+### Tag @customEnum
+
+Syntaxt: @customEnum _{type}_
```
akrantz, line 207: If the type is optional, this can be represented with square brackets, which is a common way to do it.

packages/custom-functions-metadata/Specification.md
```diff
@@ -195,3 +196,23 @@ A function can return a Promise, which will provide the value when the promise i
...
+Specify this tag to mark the TypeScript enum as an Excel custom enum. 
+
+This tag is required to generate metadata for the custom enum.
+
+* `{type}` should specify the type information within curly braces. Accepted types are `string` and `number`. Optional: if not specified, the type of the first enum value will be used.
+* Only TypeScript is supported, as JavaScript does not support enums.
+* If values of enum members are omitted, the first member will be assigned a numeric value of `0`, with subsequent values auto-incremented.
+* The types of enum members must be consistent. Mixed types within an enum will be treated as an error.
+* The enum name must be a unique identifier. Allowed characters are: A-Z, a-z, 0-9, underscore (_), and period (.).
+* The maximum length of the enum name is 128 characters.
```
akrantz, line 218: Thanks for being clear about the requirements here. Looks good.

Is there any special requirement for the first character to not be period or underscore, or perhaps has to be a letter? Just curious.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -97,6 +117,7 @@ interface IJsDocParamType {
 }
 
 const CUSTOM_FUNCTION = "customfunction"; // case insensitive @CustomFunction tag to identify custom functions in JSDoc
+const CUSTOM_ENUM = "customenum"; // case insensitive @CustomEnum tag to identify custom enums in JSDoc
```
akrantz, line 120: nit: put these constants in alphabetical order.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -431,11 +553,11 @@ function validateId(
     if (!id) {
       id = "Function name is invalid";
     }
-    const errorString = `The custom function id contains invalid characters. Allowed characters are ('A-Z','a-z','0-9','.','_'):${id}`;
+    const errorString = `The custom function or enum id contains invalid characters. Allowed characters are ('A-Z','a-z','0-9','.','_'):${id}`;
```
akrantz, line 556: nit: if you can tell whether it is a custom function or an enum, it would be nice that the error string was formatted to just say the one that it is.

packages/custom-functions-metadata/src/parseTree.ts
```diff
@@ -431,11 +553,11 @@ function validateId(
     if (!id) {
       id = "Function name is invalid";
```
akrantz, line 554: Can id be falsy here for an enum as well? This only says "Function name" is invalid.

akrantz (approved):
Looks good. I made some very minor suggestions.

packages/custom-functions-metadata/Specification.md
akrantz: Also, as commented below, if you want to denote optional parameters in square brackets and required parameters in curly braces, that is fine with me.

MiaofeiWang: Curly braces is not used to denote optional but as part of the type, for example `@customEnum {number}`. Wrapping type inside of curly braces seems to be a common practice, like custom function's `returns {number}` or TypeScript native `@enum {number}`: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#enum. 
BTW, I may need to remove this line because those tags are for functions.

packages/custom-functions-metadata/Specification.md
MiaofeiWang: This spec file seems not having a fixed practice to denote optional, but _italic_ is used in most cases, for example `@customfunction _id_ _name_`. 
Do you think we should update them all to use square brackets?

packages/custom-functions-metadata/src/parseTree.ts
MiaofeiWang: I changed to use `validateName` instead of `validateId` for a full validation.

packages/custom-functions-metadata/src/parseTree.ts
MiaofeiWang: Updated the tag strings alphabetically

packages/custom-functions-metadata/Specification.md
MiaofeiWang: Good point! Add more requirement about first character must be alphabetic character.

MiaofeiWang:
Hi @millerds , would you please help review and merge this PR? It was approved by multiple reviewers but a new minor iteration reset those approvals. Thank you.

PR #909, opened 2024-11-20
[patch] Add custom function metadata auto-gen support for @linkedEntityLoadService

jeaniehuynh:
**Change Description**:

    Updating naming of @linkedEntityDataProvider to @linkedEntityLoadService

    No

   No

### DIFF
--- a/packages/custom-functions-metadata/Specification.md
+++ b/packages/custom-functions-metadata/Specification.md
@@ -14,7 +14,7 @@ The function parameter types may be provided using the [@param](#param) tag in J
 * [@customfunction](#customfunction) id name
 * [@excludeFromAutoComplete](#excludeFromAutoComplete)
 * [@helpurl](#helpurl) url
-* [@linkedEntityDataProvider](#linkedEntityDataProvider)
+* [@linkedEntityLoadService](#linkedEntityLoadService)
 * [@param](#param) _{type}_ name description
 * [@requiresAddress](#requiresAddress)
 * [@requiresParameterAddresses](#requiresParameterAddresses)
@@ -74,7 +74,7 @@ Indicates that the function will be excluded from the autocomplete drop-down lis
 
 If the function is manually spelled correctly in the grid, the function will still execute.
 
-A function cannot have both `@excludeFromAutoComplete` and `@linkedEntityDataProvider` tags.
+A function cannot have both `@excludeFromAutoComplete` and `@linkedEntityLoadService` tags.
 
 ---
 ### @helpurl
@@ -84,7 +84,7 @@ Syntax: @helpurl _url_
 The provided _url_ is displayed in Excel.
 
 ---
-### @linkedEntityDataProvider
+### @linkedEntityLoadService
 
 Indicates that the function is a "special" custom function that is meant to act as the "loadFunction" for user defined `LinkedEntityDataDomain`s. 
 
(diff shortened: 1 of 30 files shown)


jeaniehuynh:
@millerds, could I get a review on this PR? Thank you!

millerds:
Please remove unrelated changes to other packages (all the package-lock file change).

millerds:
Why are we renaming it?  Does that break existing usages of the old name?

jeaniehuynh:
> Why are we renaming it? Does that break existing usages of the old name?

We've re-arched this portion of our feature and the new name/signature aligns better with its new function. See this PR for more context: https://office.visualstudio.com/Office/_git/Office/pullrequest/3418078

This is part of a tented project, so only those customers are using this API currently and they are aware of this upcoming API change.

jeaniehuynh:
> Please remove unrelated changes to other packages (all the package-lock file change).

@millerds done!

akrantz:
The rename of the tag is fine. Personally, I preferred the data provider name from the service as this isn't really a service, but that's my own preference. But I know that this is being renamed due to the change in how the data provider function functions.

 There is an opportunity to have better type validation for the requires arguments to the function which uses the tag. Currently the tests just show `unknown` for the type. The previous version had the linkedProviderId as the first argument. The new form has a request. I think there should be a defined type for that argument type and this type should be required for the function. I would recommend this be done as part of a future change.

akrantz (approved):

jeaniehuynh:
> The rename of the tag is fine. Personally, I preferred the data provider name from the service as this isn't really a service, but that's my own preference. But I know that this is being renamed due to the change in how the data provider function functions.
> 
> There is an opportunity to have better type validation for the requires arguments to the function which uses the tag. Currently the tests just show `unknown` for the type. The previous version had the linkedProviderId as the first argument. The new form has a request. I think there should be a defined type for that argument type and this type should be required for the function. I would recommend this be done as part of a future change.
> 
> @jeaniehuynh @MiaofeiWang

Thanks @akrantz! Here's [the thread](https://dev.azure.com/office/Office/_git/Office/pullRequest/3418078#1729047962) on how this name was ultimately chosen for some context. "This custom function technically acts as a proxy "service" in the add-in architecture"

Totally agree that it would be nice to have type validation, but the metadata plugin doesn't support Office.js types yet, and the type of `request` is `Excel.Excel.LinkedEntityLoadServiceRequest`. As soon as we have support for this, we will add validation!

jeaniehuynh:
@millerds Seems like there's an unrelated test failure (saw that it failed Miaofei's PR too), is that something you can override to get this checked in?

MiaofeiWang:
> Totally agree that it would be nice to have type validation, but the metadata plugin doesn't support Office.js types yet, and the type of `request` is `Excel.Excel.LinkedEntityLoadServiceRequest`. As soon as we have support for this, we will add validation!

@jeaniehuynh I don't think there is any plan to support all Office.js types in this package. One solution we used before is to treat `Excel.LinkedEntityLoadServiceRequest` as a indeterministic reference type, and we have made such change for `Excel.CellValue`: https://github.com/OfficeDev/Office-Addin-Scripts/blob/ecde2fa315023a0054354f9541adaa07c0253d28/packages/custom-functions-metadata/src/parseTree.ts#L1141 . Here is a test sample: https://github.com/OfficeDev/Office-Addin-Scripts/blob/master/packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/functions.ts

millerds:
> @millerds Seems like there's an unrelated test failure (saw that it failed Miaofei's PR too), is that something you can override to get this checked in?

This is an incompatibility between the tool and forked repos.  I need to double check on a fix to correct this.  I have been bypassing it for the time being.  Once the comments have been addressed I will do this.

jeaniehuynh:
> > Totally agree that it would be nice to have type validation, but the metadata plugin doesn't support Office.js types yet, and the type of `request` is `Excel.Excel.LinkedEntityLoadServiceRequest`. As soon as we have support for this, we will add validation!
> 
> @jeaniehuynh I don't think there is any plan to support all Office.js types in this package. One solution we used before is to treat `Excel.LinkedEntityLoadServiceRequest` as a indeterministic reference type, and we have made such change for `Excel.CellValue`:
> 
> https://github.com/OfficeDev/Office-Addin-Scripts/blob/ecde2fa315023a0054354f9541adaa07c0253d28/packages/custom-functions-metadata/src/parseTree.ts#L1141
> 
> . Here is a test sample: https://github.com/OfficeDev/Office-Addin-Scripts/blob/master/packages/custom-functions-metadata/test/cases/input-parameter-types-cellvalue/functions.ts

Thanks for sharing, @MiaofeiWang! This is a great workaround. @millerds As Adam mentioned, we'll take care of this in a future change as its outside the scope of this change, so we can go ahead and check this one in. Thanks!

PR #911, opened 2024-12-04
Enable parse @requiresAddress and @requiresParameterAddresses together with @streaming for custom functions

sunyao-ms:
**Change Description**:

 
This PR containing the following change:   
1. Enable parse @requiresAddress and @requiresParameterAddresses together with @streaming for custom functions
2. Add unit test

    No

3. **Do these changes impact *documentation*?** (e.g., a tutorial on https://learn.microsoft.com/office/dev/add-ins/overview/office-add-ins)
    Yes, we will introduce new properties for custom functions, so the doc on https://learn.microsoft.com/en-us/javascript/api/custom-functions-runtime/customfunctions.streaminginvocation?view=excel-js-preview should be updated with how to use the new tags.

    Local unit test pass.


akrantz:
I think this can be simplified so that the JSDoc tags are the same for streaming and non-streaming functions, e.g. `@requiresParameterAddress`. When the metadata is written out, it can determine whether it is a streaming function or not and can update the metadata appropriately. This way the developer has the same experience adding `@requiresParameterAddress` whether it is streaming or not, and there is an implementation detail that the metadata info is different due to the need to support backward compatibility.

akrantz (changes requested):
See previous comment

packages/custom-functions-metadata/test/cases/requiresstreamaddress/expected.json
```diff
@@ -0,0 +1,24 @@
+{
```
akrantz, line 1: Please call this requiresaddress-streaming (which is converted from error-requiresaddress-streaming), or consider merging both non-streaming and streaming cases into single requiresaddress test case.

packages/custom-functions-metadata/test/cases/requiresstreamparameteraddresses/expected.json
```diff
@@ -0,0 +1,68 @@
+{
+    "allowCustomDataForDataTypeAny": true,
```
akrantz, line 2: Let's name this test case requiresparameteraddress-streaming. This is consistent with other test cases in having the tag name first and the variant after the hyphen.

akrantz (commented):
OK, this is good. I have some suggestions for the test cases to update the test case names for consistency.

Is there an error test case when you have these tags but are missing the invocation parameter? It would be good to have that since putting one of these tags should require the invocation parameter which then provides the address / parameter addresses.

sunyao-ms:
> OK, this is good. I have some suggestions for the test cases to update the test case names for consistency.
> 
> Is there an error test case when you have these tags but are missing the invocation parameter? It would be good to have that since putting one of these tags should require the invocation parameter which then provides the address / parameter addresses.

I have added the missing invocation test case.

akrantz (approved):
Thanks for making the updates!

PR #924, opened 2025-03-05
[minor] include copilot agent files in zip

akrantz:
Adds files related to Copilot Agents to the zip file.

No

No

Start debugging in VS code with a project which uses JSON manifest with copilot agents.


packages/office-addin-manifest/src/export.ts
```diff
@@ -31,7 +38,36 @@ async function createZip(manifestPath: string): Promise<AdmZip> {
     throw new Error(`The file '${manifestPath}' does not exist`);
   }
 
-  const manifest: devPreview.DevPreviewSchema = await ManifestUtil.loadFromPath(manifestPath);
+  const manifest: DevPreviewSchema = readJsonFileSync(manifestPath);
```
millerds, line 41: So does the `ManifestUtil.loadFromPath` not work with the locally created schema then?

packages/office-addin-manifest/src/export.ts
```diff
@@ -31,7 +38,36 @@ async function createZip(manifestPath: string): Promise<AdmZip> {
...
+      const file: string = agent?.file;
+      const filePath = path.join(manifestDir, file);
+
+      if (!fs.existsSync(filePath)) {
+        throw new Error(
+          `copilotAgents.declarativeAgents[${agentIndex}].file does not exist:  ${filePath}`
+        );
+      }
+
+      zip.addLocalFile(filePath, "", file);
```
millerds, line 55: What is the 3-argument version of this method?  I"ve only seen the 1 or 2 argument version and that is all that is documented (which isn't great).

packages/office-addin-manifest/src/export.ts
```diff
@@ -31,7 +38,36 @@ async function createZip(manifestPath: string): Promise<AdmZip> {
...
+      const agentJson: DeclarativeCopilotManifestSchema = readJsonFileSync(filePath);
+
+      agentJson?.actions?.forEach((action, actionIndex) => {
+        if (action?.file) {
+          const actionFilePath = path.join(manifestDir, action.file);
+          if (!fs.existsSync(actionFilePath)) {
+            throw new Error(`actions[${actionIndex}].file does not exist: ${actionFilePath}`);
+          }
+
+          zip.addLocalFile(actionFilePath, "", action.file);
```
millerds, line 65: We need to maintain the path structure of the dependent files inside the zip, otherwise cases where there are other folders in the path for the agent files will break because the file will have been relocated in the zip folder.  See the icon files as an example.

Same is true of the base agent path.

millerds (changes requested):

packages/office-addin-manifest/src/export.ts
akrantz: I believe the return type was different. There's no real value to using their function as I needed to use the readJsonFileAsync() function for both JSON files. Seems simplest to use the same function for both.

packages/office-addin-manifest/src/export.ts
akrantz: I was just following what I saw in code elsewhere. Not sure what the second parameter is but it seemed the first was the file and the third was the string name for the file.

millerds:
Fixed in another PR that uses updated json manifest support packages.

PR #936, opened 2025-05-08
[minor] Include more files in manifest zip

millerds:
**Change Description**:
Update the zip file generation for json manifests to include adding language files and declarative copilot agent files.  Also bumped versions of related dependent packages.

No.

If documentation include information on the content of the genareated file then yes . . . otherwise probably not.

Locally updated and ran tests and tried a test project.


Rick-Kirkham (commented):
This is great, but one question: When there's a Copilot agent, there are usually two extra files; (1) a declarative agent JSON whose relative URL is in the manifest, and (2) a plugins JSON whose relative path is in the declarative agent JSON. Both of these files need to added to the package. Will this PR do that?

millerds:
> declarative

Yes it will.  The export.ts file gets all the agents from the manifest and for each agent it adds the agent json to the zip and then looks for the actions collection (I think this is also referred to as plugins) in that json file and adds each of those new json references as well.

Rick-Kirkham (approved):

PR #952, opened 2025-10-16
Update Vulnerable Packages

ztcdsb:
1. test against node 24.10.0
2. update vulnerable packages
3. update NPM and Node requirement range

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,8 +16,8 @@
         "typescript-eslint": "^8.8.1"
       },
       "engines": {
-        "node": ">=16 <21",
-        "npm": ">=6 <=11"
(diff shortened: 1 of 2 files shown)


millerds:
Also . . . what version of node are you using?  We should be compatible with current versions (I think v24 might be the new LTS supported version).

akrantz:
Node v24 became LTS on Oct 8, 2025. @millerds Are there changes needed to make it supported?

millerds:
> Node v24 became LTS on Oct 8, 2025. @millerds Are there changes needed to make it supported?

Other than validation that the packages don't break . . . probably not.  In the case of the templates, it required package version update for the tests to work.

PR #965, opened 2025-11-11
[minor] User/millerds/update app insights version

millerds:
**Change Description**:
The call to application insights to send telemetry data was randomly failing to connect to the end point.  The appInsights package we're using it 6 years and 2 major versions old.  New packages use a connection string instead of an instrumentation key which I think will be more reliable.  Updated office-addin-usage-data to use the new package, and then updated all the internal references to use the connections string option.  

I also fixed a couple of other package mistakes made in other check-ins.

No.  just internal method calls.

I don't think so.

Ran validation script and checked that the test telemetry was still being logged into the appInsights table.


packages/office-addin-manifest/src/export.ts
```diff
@@ -35,9 +35,9 @@ async function createZip(manifestPath: string): Promise<AdmZip> {
     throw new Error(`The file '${manifestPath}' does not exist`);
   }
 
-  const manifest: TeamsManifestVDevPreview = (await AppManifestUtils.readTeamsManifest(
+  const manifest: DevPreviewSchema = (await AppManifestUtils.readTeamsManifest(
     manifestPath
-  )) as TeamsManifestVDevPreview;
+  )) as DevPreviewSchema;
```
akrantz: nit: if you have the type on the variable declaration then I don't think you need the `as DevPreviewSchema`.

Oleg-O (approved):

PR #968, opened 2025-12-23
fix: resolve Invalid id in Terminal add in still installed error

archiesaxenaa:
**Root Cause**
- DA projects return title IDs with `T_` prefix (vs. `U_` for traditional Office add-ins)
- Validation regex didn't recognize `T_` prefix
- ATK CLI's `atk uninstall --mode title-id` doesn't support `T_` IDs (30s timeout)

**Solution**
- Detect `T_` prefix title IDs
- Strip the `T_` prefix to extract the manifest GUID
- Use `--mode manifest-id` instead of `--mode title-id` for uninstall
- `U_` IDs continue using `--mode title-id` (existing behavior)

**Changes**
- Modified `uninstallWithTeams()` in `packages/office-addin-dev-settings/src/publish.ts`
- Added separate regex patterns for `T_` and `U_` prefixes
- Auto-convert `T_<guid>` → `--mode manifest-id --manifest-id <guid>`

Local Testing done via `npm link`.
Stop command now works for DA projects without errors or timeouts. Clean add-in un-installation when office app is closed and debug is stopped.

image (2)

### DIFF
--- a/packages/office-addin-dev-settings/src/publish.ts
+++ b/packages/office-addin-dev-settings/src/publish.ts
@@ -57,10 +57,15 @@ export async function uninstallWithTeams(id: string): Promise<boolean> {
   return new Promise((resolve, reject) => {
     const guidRegex = /[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}/;
     const manifestIdRegex = new RegExp(`^${guidRegex.source}$`);
-    const titleIdRegex = new RegExp(`^U_${guidRegex.source}$`);
+    const titleIdRegexU = new RegExp(`^U_${guidRegex.source}$`);
+    const titleIdRegexT = new RegExp(`^T_${guidRegex.source}$`);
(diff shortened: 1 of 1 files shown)


millerds (changes requested):
The titleId was used because originally that was the only way to remove an add-in.  That id had to be stored in the registry to use later when removing the add-in.  Using the manifestId was added later when that support got added added to the ttk cli but left the titleId usage in place, so both were supported.

It seems odd to me the declarative agents need a different prefix and that this new prefix is not supported by the cli.  Seems like the bug is with the cli needing support for the new prefix.  Granted, the scripts code would still need to be updated because it's looking for a specific prefix, but it becomes a simpler change (potentially just updating the one regex to look for both prefixes).  If the cli is not going to be updated, then this script should be updated based on the general plans for the cli.

Bottom line, we shouldn't be trying to special case each prefix.  We should be making changes that are future proof the scripts code to work with add-ins generally and avoid examining values to determine what to do.  Maybe that means becoming titleId type agnostic, or maybe that means always using manifestId (if that is the broadly accepted way of identifying an add-in).  Examining the value is too fragile.

PR #980, opened 2026-03-13
[patch]: fix web sideload with manifest preview

aminya:
**Change Description**:

To side load on the web, the sourceLocation is required, however, the dev manifest preview fails if you add it in the manifest file. This adds a default location, so that web side load works with the manifest preview

1. **Do these changes impact *command syntax* of any of the packages?**  No

2. **Do these changes impact *documentation*?** No

`bunx -b office-addin-debugging start --dev-tools manifest.json web --document <some_doc>`

### DIFF
--- a/packages/office-addin-dev-settings/src/sideload.ts
+++ b/packages/office-addin-dev-settings/src/sideload.ts
@@ -135,13 +135,9 @@ export async function generateSideloadUrl(
   if (!manifest.id) {
     throw new ExpectedError("The manifest does not contain the id for the add-in.");
   }
-
-  if (
-    manifest.defaultSettings === undefined ||
-    manifest.defaultSettings.sourceLocation === undefined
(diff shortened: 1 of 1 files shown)


millerds (commented):
What do you mean by "dev manifest preview fails if you add it in the manifest"?  Can you explain some more what you are seeing happen.

I also noticed you are using the json manifest and I don't think that has a default location property.  It existed in the xml schema but doesn't appear to have been translated over to the json schema (that I can see).  I think you have found a legitimate flaw in the sideloading code because that value will always be empty for json manifests (only gets set in manifestHandlerXml.ts), but I think the better fix would be to set that value in manifestHandlerJson.ts with a legitimate value from the json manifest (from the runtimes section?) or maybe see if a sideloading URL can be crafted that works without that value.

PR #991, opened 2026-04-18
Add support for disk manifests and source bundle override

kwame-ansong:
    Yes. This change adds support for configuring disk manifests and overriding the source bundle JS file.

    Yes. This change adds support for configuring disk manifests and overriding the source bundle JS file.

No update is needed for the README.md file because the changes are meant for internal usage by our add-ins toolset.

All tests passed via `npm run test`

### DIFF
--- a/packages/office-addin-debugging/src/start.ts
+++ b/packages/office-addin-debugging/src/start.ts
@@ -8,7 +8,7 @@ import fspath from "path";
 import * as devCerts from "office-addin-dev-certs";
 import * as devSettings from "office-addin-dev-settings";
 import os from "os";
-import { DebuggingMethod, sideloadAddIn, launchDesktopApp } from "office-addin-dev-settings";
+import { DebuggingMethod, sideloadAddIn } from "office-addin-dev-settings";
 import { OfficeApp, OfficeAddinManifest } from "office-addin-manifest";
 import * as nodeDebugger from "office-addin-node-debugger";
 import * as debugInfo from "./debugInfo";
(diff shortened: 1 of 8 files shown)


bercovMSFT (approved):

ndeleuze (dismissed):

packages/office-addin-debugging/src/start.ts
```diff
@@ -394,6 +427,8 @@ export async function startDebugging(manifestPath: string, options: StartDebuggi
       } catch (err) {
         throw new Error(`Unable to sideload the Office Add-in. \n${err}`);
       }
+    } else if (app && isDesktopAppType && isWindowsPlatform) {
+      await launchDesktopApp(app, manifestInfo, undefined /* document */);
```
millerds, line 431: So far, all the sideload setup and app launching has been contained in the dev-settings package, but this starts to split some of that by using launchDesktopApp (not previously exported) outside of dev-settings.  It seems to me like configuring the load from disk and starting the host belongs in dev-setting (like the sideload call) and it should be just called from here.

In the case of Outlook sideloading has always meant going to a service (exchange or new service), but for WXP pre json manifst it also meant loading from disk.  With json manifest sideloading does focus more on loading through the service (not sure that looks like for WXP) and a distinction might be needed.  The load from disk scenario could be incorporated into the "sideloadAddIn" method in dev-settings of if you felt it better to make a separate method for local loading that would be find too.  The main thing is to keep that part of the process in the same place.

millerds, line 431: This line was setup with the assumption of document always being null which is specific Outlook behavior.  This package and library is for all host and while some hosts don't support the methodology for loading, we should be using the existing "document" object like is used for other calls.

millerds (changes requested):

packages/office-addin-debugging/src/start.ts
kwame-ansong: The README for `office-addin-dev-settings` says it "Provides the ability to configure developer settings for Office Add-ins.". The README for `office-addin-debugging` talks about debugging, which requires running Outlook. So, it seems a bit confusing that you need to use the dev-settings package to run Outlook instead of the debugging package. There's no way to debug the add-in without running Outlook.

I've incorporated disk-loading into the `sideloadAddIn` method as suggested. However, I want to point out that disk-loading is separate from side-loading, so the incorporation introduces added complexity. We can rename the method appropriately in the future or do some future clean up work.

kwame-ansong: No longer applicable in the latest commit.

millerds: Thanks.  All that makes sense to me.  I have had similar thoughts and had ideas on refactoring packages along those lines (the design has been this way since before me), but have yet had a strong reason (or time) to do so.

millerds (approved):

PR #993, opened 2026-05-13
[minor] Remove usage data logging

millerds:
**Change Description**:
Remove application insights dependency.  Packages stop calling methods that would send the telemetry and the related office-addin-usage-data calls that that manage this now do a no-op instead.

No. (though it does make a bunch of them no-op)

Maybe.  Not sure if we have documentation that describes the telemetry sending we do.

Ran automated tests (which were updated based on the changes in office-addin-usage-data)


packages/office-addin-dev-settings/src/sideload.ts
```diff
@@ -334,60 +333,54 @@ export async function sideloadAddIn(
...
+    case AppType.Web: {
+      if (!document) {
+        throw new ExpectedError(`For sideload to web, you need to specify a document url.`);
+      }
+      await launchWebApp(app, manifestPath, manifest, document);
+      break;
     }
-    usageDataObject.reportSuccess("sideloadAddIn()");
-  } catch (err: any) {
-    usageDataObject.reportException("sideloadAddIn()", err);
```
Oleg-O, line 389: I wonder... would it be better if instead of removing all the call sites for telemetry, we could keep them and replace with a helper that would do a no-op now, but in future would potentially do some utel.js call?

Oleg-O (dismissed):

packages/office-addin-dev-settings/src/sideload.ts
millerds: That's an interesting thought . . . keep the setup in place to allow for a future replacement.  

An updated to utel.js may not use the same method signatures, but even if it didn't, we would have all the hooks previously used as a starting point to implement updates.

Rick-Kirkham (commented):
Approving. (We don't mention the telemetry in the docs.) **But what is the reason we're making this change?**

millerds:
> Approving. (We don't mention the telemetry in the docs.) **But what is the reason we're making this change?**

The telemetry isn't being used, and we've had some privacy questions raised on some data that started getting automatically collected with the applicationsInsights package, and some customers have questioned our doing this.  We're taking this implementation out now, but will consider adding a more compliant telemetry collection model if the data is requested in the future.

Oleg-O (approved):

